PR #1, opened 2019-05-06
Initial client library submission

merlimat:
This is the current work in progress for native Go client library. 

Notes: 
 * This PR contains: 
    - Working version of producer
    - Almost all producer features implemented (batching, partitioning, compression, ...). 
    - Missing Authentication and TLS connections
    - Perf client for testing
 * Integration tests will be added once the consumer code is in place


pulsar/error.go
```diff
@@ -0,0 +1,104 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
+package pulsar
+
+import "C"
```
wolfstudy: `import C`? we should remove it?

I am thinking about another question, do we need to define Result or error here, because this looks more like the style of the c language, but not the style of the go language. Of course, the error in the go language itself does not support stack return. Do we consider using [pkg-errors](https://github.com/pkg/errors) or [juju-errors] (https://github.com/juju/errors)

perf/pulsar-perf-go.go
```diff
@@ -0,0 +1,49 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+	log "github.com/sirupsen/logrus"
```
wolfstudy: we should add package management tool. `go mod`, `dep`, `glide` and so on. I recommend using `go mod`

pulsar/impl/backoff.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"time"
+)
+
+type Backoff struct {
+	backoff time.Duration
+}
+
+const minBackoff = 100 * time.Millisecond
+const maxBackoff = 60 * time.Second
```
wolfstudy:
```suggestion
const (
    minBackoff = 100 * time.Millisecond
    maxBackoff = 60 * time.Second
)
```

pulsar/impl/batch_builder.go
```diff
@@ -0,0 +1,166 @@
...
+	"pulsar-client-go/pulsar/impl/compression"
+	pb "pulsar-client-go/pulsar/pulsar_proto"
+	"time"
+)
+
+const MaxMessageSize = 5 * 1024 * 1024
+
+const MaxBatchSize = 128 * 1024
+
+const DefaultMaxMessagesPerBatch = 1000
```
wolfstudy:
```suggestion
const (
    MaxMessageSize = 5 * 1024 * 1024
    MaxBatchSize = 128 * 1024
    DefaultMaxMessagesPerBatch = 1000
)
```

pulsar/util_test.go
```diff
@@ -0,0 +1,31 @@
...
+// under the License.
+//
+
+package pulsar
+
+import (
+	"fmt"
+	"time"
+)
+
```
wolfstudy: `_test` has a special meaning in go, which is a test file for the `util` file. When we execute the `go test`, we will scan this file. Can we consider naming this file as `test_helper` and so on

pulsar/error.go
merlimat: Ups, "import C" was a leftover (the API file were imported from the current library). 

Regarding exposing Result, the idea is more to give an option to downcast `error` to a `pulsar.Error` and retrieve the exact code, maybe to give the chance to application to handle different errors in different ways. It's not strictly necessary per-se. The important part is, at least initially, that we can print a meaningful error message.

perf/pulsar-perf-go.go
merlimat: Surely, I left it over for now since it's really a WIP and I'm not superfamiliar with these tools. Perfectly fine to go with `go mod`.

pulsar/util_test.go
merlimat: Sure, make sense to rename

pulsar/impl/backoff.go
merlimat: 👍

pulsar/impl/batch_builder.go
merlimat: 👍

perf/pulsar-perf-go.go
wolfstudy: ok, when this pull request is merged, I will add it.

wolfstudy: Another question is, do we need to add the `vendor` directory, after executing `go mod`, all the dependencies will be stored in the `vendor` directory. The advantage of adding the vendor directory is that when the code repository we reference is deleted, we can also maintain our own local version. Of course, this is not a standard. Some projects have uploaded vendor. For example: [etcd](https://github.com/etcd-io/etcd) Some items are not uploaded, for example: [TiDB](https://github.com/pingcap/tidb). so, in our project, do we upload the vendor?

pulsar/error.go
wolfstudy: ok, I agree with your point of view, then we are considering encapsulating the information of these error codes into the third-party Error library, which guarantees that we have only one error in the global use.

pulsar/producer.go
```diff
@@ -0,0 +1,167 @@
...
+	// Send a message
+	// This call will be blocking until is successfully acknowledged by the Pulsar broker.
+	// Example:
+	// producer.Send(ctx, pulsar.ProducerMessage{ Payload: myPayload })
+	Send(context.Context, *ProducerMessage) error
+
+	// Send a message in asynchronous mode
+	// The callback will report back the message being published and
+	// the eventual error in publishing
+	SendAsync(context.Context, *ProducerMessage, func(MessageID, *ProducerMessage, error))
```
wolfstudy: Here we can see that the interface exposed to the user is consistent with the cgo version of pulsar-client-go. I can understand that the benefit of this is compatible with the cgo version of pulsar-client-go, so that when the user is replacing Just replace the lib you depend on. But the operation of async, we use channel to deal with it is more reasonable? Similar like this:

```
Type XXXAsync interface {
        Input() chan<- *ProducerMessage
        Successes() <-chan *ProducerMessage
}
```

wolfstudy: Regarding this question, I am thinking, we are strictly following the cgo version of the API interface, or try to keep the interface with it, and more to write code according to the style of the go community.

perf/pulsar-perf-go.go
merlimat: I'd prefer to only check-in code that is manually written for this library in this repo. Of course there are always exception (eg: the protobuf generated code, for ease of use), but if the downloading of specific tag of a dependency can be automated, that seems a better solution.

We could rethink that if we see that tags are being removed upstream.

wolfstudy: +1

pulsar/producer.go
sijie: IMO it is better to design an API that is more native to go developers.

merlimat: Sure, the current code just imported the existing API and plugged the new implementation on that. 

 Since it’s a new repo, we don’t need to keep company with the existing go library. And until this gets officially released, we can keep iterating over the API without worries.

PR #4, opened 2019-05-15
add go mod for pulsar-client-go

wolfstudy:

add `go mod` for project, you can run

```
go mod download
``` 

get the corresponding dependencies.


wolfstudy:
@merlimat PTAL

wolfstudy:
When we execute `go mod init`, the path used is: `github.com/apache/pulsar-client-go`, so you may need to update your local folder path as follows:

```
mkdir -p ~/github.com/apache/pulsar-client-go
```

perf/perf-producer.go
```diff
@@ -22,11 +22,12 @@ package main
 import (
 	"context"
 	"encoding/json"
+	`github.com/Anankke/bilibili-backup/library/net/http/blademaster/middleware/rate`
```
merlimat: Is this import intended?

merlimat:
@wolfstudy Can you also add instructions in the top-level README.md on how to compile with dependencies? Would that apply only to people trying to do dev work on the library, right?

merlimat:
Actually, let me merge this, so it will simplify adding the unit tests scripts.

perf/perf-producer.go
wolfstudy: oh no, actually, we shouldn't import this, next pull request, i will remove it

merlimat: No worries. I already removed it in my code.

PR #12, opened 2019-05-16
package logrus for project

wolfstudy:

package `logrus` for project.

When we output log information, it is not recommended to capitalize the first letter. It is not recommended to add any punctuation at the end.

If we can clearly know the type that needs to be printed, it is recommended to use the format of the specified type. If there is no way to know, you can use `%v`, `%#v` or `%+v`.

```
Printf("%v", people)   {zhangsan}，
Printf("%+v", people)  {Name:zhangsan}
Printf("%#v", people)   main.Human{Name:"zhangsan"}
```


pkg/log/log.go
```diff
@@ -0,0 +1,201 @@
...
+            strings.Contains(name, "github.com/apache/pulsar-client-go/pkg/log")
+}
+
+// formatter is for compatibility with ngaut/log
+type formatter struct {
+    DisableTimestamp, EnableEntryOrder bool
+}
+
+// Format implements logrus.Formatter
+func (f *formatter) Format(entry *log.Entry) ([]byte, error) {
```
merlimat: I don't see why do we need to re-implement the logger API. Logrus seems to be well established and the API has been in use for a while. Why not just use that instead of reinventing it?

pulsar/impl_partition_producer.go
```diff
@@ -134,7 +131,7 @@ func (p *partitionProducer) grabCnx() error {
 	})
 
 	if err != nil {
-		p.log.WithError(err).Error("Failed to create producer")
+		log.Errorf("create producer [%s] error: %+v", *p.producerName, err)
```
merlimat: If you remove the context, all these logs will have no sense. The logs are supposed to be relative to some event which will have some context, eg. `topic=my-topic, producer-name=my-producer`. These are vital fro the logs. 

Repeating the the context information each time has 2 problems: 
 * It's hard to be consistent throughout the code base in terms of naming and parameters
 * We'd be losing the semi-structure of the logs

Saving the logger as part of the state ensure that we always have the context, added in a consistent way and will let the application process the logs easily.

pulsar/impl_partition_producer.go
```diff
@@ -152,10 +149,10 @@ func (p *partitionProducer) grabCnx() error {
 	}
 	p.cnx = res.Cnx
 	p.cnx.RegisterListener(p.producerId, p)
-	p.log.WithField("cnx", res.Cnx).Debug("Connected producer")
+	log.Debugf("connected producer [%s] cnx is: %+v", *p.producerName, res.Cnx)
```
merlimat: Why lose the clean way of attaching the context?

pulsar/impl_partition_producer.go
```diff
@@ -172,7 +169,7 @@ func (p *partitionProducer) ConnectionClosed() {
 }
 
 func (p *partitionProducer) reconnectToBroker() {
-	p.log.Info("Reconnecting to broker")
+	log.Info("reconnecting to broker")
```
merlimat: Same as above, without the producer context, this logs is not useful.

pulsar/internal/connection.go
```diff
@@ -141,7 +138,7 @@ func (c *connection) start() {
 }
 
 func (c *connection) connect() (ok bool) {
-	c.log.Info("Connecting to broker")
+	log.Info("connecting to broker")
```
merlimat: Again, context is needed here

pulsar/internal/lookup_service.go
```diff
@@ -81,7 +81,7 @@ func (ls *lookupService) Lookup(topic string) (*LookupResult, error) {
 			return nil, err
 		}
 
-		log.WithField("topic", topic).Debugf("Got topic lookup response: %s", res)
+		log.Debugf("got topic: %s lookup response: %s", topic, res)
```
merlimat: Using the `WithField()` provides a consistent way to present the context. In this file we're already changing 3 different ways to present the topic: 
```go
log.Debugf("got topic: %s lookup response: %s", topic, res)
log.Debugf("follow redirect to broker. %v / %v - Use proxy: %v - Topic is: %s",
log.Debugf("successfully looked up topic [%s] on broker. %s / %s - Use proxy: %t",
```

`WithField()` will always attach `topic=my-topic` to the log line. If I want to grep for that topic, that makes the life much easier even in text format.

pkg/log/log.go
```diff
@@ -0,0 +1,201 @@
...
+func SetOutput(filename string) {
+    output := &lumberjack.Logger{
+        Filename:  filename,
+        LocalTime: true,
+    }
+    log.SetOutput(output)
+}
+
+func init() {
+    log.SetLevel(defaultLogLevel)
```
merlimat: This is a library so we shouldn't set any specific logger formatting property.

We should just leave the application free to configure through Logrus and point then to that documentation.

wolfstudy:
1. Use logrus lib print:
```
time="2019-05-17T09:58:37+08:00" level=info msg="hello logrus"
```

2. Printed log information after rewriting the `Formatter`:

```
2019/05/17 10:00:25.530 test.go:9: [info] hello log
```
Users may prefer to see the form of this kind of log output below, which is why I am overriding the `Formatter` method.

3. Use logrus's `WithFiled` to print:
```    
    name := "pulsar"
    logrus.WithField("name", name).Info("hello pulsar name")
```

output:

```
time="2019-05-17T10:06:09+08:00" level=info msg="hello pulsar name" name=pulsar
```

4. Print using a packaged library:

```
func main() {
    name := "pulsar"
    logutil.Infof("hello pulsar name: %s", name)
}
```

output:

```
2019/05/17 10:09:17.905 test.go:9: [info] hello pulsar name: pulsar
```

pulsar/internal/lookup_service.go
wolfstudy: If we really have this kind of scenario requirement, we can repackage the `WithField` method as follows:

```
func WithField(key string, value interface{}) *log.Entry {
    return log.WithField(key, value)
}
```

perf/pulsar-perf-go.go
```diff
@@ -31,12 +30,6 @@ type ClientArgs struct {
 var clientArgs ClientArgs
 
 func main() {
-	log.SetFormatter(&log.TextFormatter{
-		FullTimestamp:   true,
-		TimestampFormat: "15:04:05.000",
-	})
-	log.SetLevel(log.InfoLevel)
-
```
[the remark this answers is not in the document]
wolfstudy: Here I repackaged the `SetLevel` method, the level defaults to the `info` form, so when the log level has not changed, we don't have to reset it.

If necessary, we can reset it by the following method:

```
func SetLevel(level string) {
    log.SetLevel(stringToLogLevel(level))
}
```

pkg/log/log.go
wolfstudy: The default form of `logrus` does not provide the line number information of the file. When the log is output, the user may prefer to locate the current log output by the line number information.

pulsar/impl_partition_producer.go
wolfstudy: Yes, the context information is the meaning of the log output. Users can quickly locate the problem based on this information. Here, I did not remove the context information, but changed the way to output.

wolfstudy:
Our main point of disagreement is, in what form the log is output. I prefer the latter, that is, re-implement the formatter method, customize our own log output format, and package the log into a lib for use by other modules. If, in the current way, we need to add the `log` field to most of the structs defined in the future, this will make things difficult.

pulsar/internal/lookup_service.go
merlimat: I think it makes it much easier to be consistent (adding `topic=xyz` in same format for each log message) and also allows for structured logs, if desired.

pulsar/impl_partition_producer.go
merlimat: In the above example, the topic name was already set in the `p.log` context and got missing in the refactoring.

merlimat:
> Our main point of disagreement is, in what form the log is output. I prefer the latter, that is, re-implement the formatter method, customize our own log output format, and package the log into a lib for use by other modules.

I think the log format is not up to us to decide 😄  . Since we're building a library, we need to leave it the users of our library, the applications, to decide which format they want. 

Logrus seems it does a very good job at a few key things: 
 * Application can decide the formatting of logs. All the library it depends on, that are using Logrus, will immediately get configured.
 * Structured logging. This makes the log message much more organized. This is especially critical if an application wants to write logs in JSON format (which is trivial with Logrus), eg. for getting them into ElasticSearch or similar.

> If, in the current way, we need to add the log field to most of the structs defined in the future, this will make things difficult.

I think it makes the logs very consistent and organized. Every log message printed by the producer controller should contain the context of, for example, the topic name. As I mentioned above, in the refactoring you were already using 3 different ways to pass the topic name within few lines of code, while many other log message were missing the topic name context.
This is normal, and we've been fighting around this in Java for long time, since we're constrained to Slf4J APIs. 

Setting the context in the logger itself and storing it in the producer/consumer/connection/..., makes it easier to not forget the context and always have the same format.

> Use logrus lib print:
> time="2019-05-17T09:58:37+08:00" level=info msg="hello logrus"

This gets printed when it's not on a terminal.

> Printed log information after rewriting the Formatter:
> 2019/05/17 10:00:25.530 test.go:9: [info] hello log

If we use Logrus as it is, the users will be able to decide which format text vs json, and which fields in which order to present. This looks more powerful than any default setting we can put as default.

wolfstudy:
> I think the log format is not up to us to decide 😄 . Since we're building a library, we need to leave it the users of our library, the applications, to decide which format they want.

haha, yes I was wrong, I confused this relationship.

will close this pull request, thanks @merlimat

merlimat:
@wolfstudy regarding the `Upper case` vs `lower case` for log messages, I'm open to that if that's the preferred form in Go.

PR #14, opened 2019-06-04
[issues 13] Fix go func(){} usage question

### LINKED ISSUE #13: [Bug Fix] About go func(){} usage
In [impl_producer.go#L79-L84](https://github.com/apache/pulsar-client-go/blob/master/pulsar/impl_producer.go#L79-L84), i think the correct way of handling should be:

```
	for partitionIdx, partition := range partitions {
		go func(index int) {
			prod, err := newPartitionProducer(client, partition, options, index)
			c <- ProducerError{partitionIdx, prod, err}
		}(partitionIdx)
	}
```

test code:

```
package main

import (
	"fmt"
	"time"
)

func main() {
	s := []string{"a", "b", "c"}
	for _, v := range s {
		go func() {
			fmt.Println(v)
		}()
	}
	time.Sleep(time.Second * 1000)
}
```

output:

```
c
c
c
```

As you can see, you can only get the value of the last loop before passing the copy value of the variable `v` into the anonymous function.

wolfstudy:

Fixes #13 

## Motivation

[issues 13] Fix go func(){} usage question

### DIFF
--- a/pulsar/impl_producer.go
+++ b/pulsar/impl_producer.go
@@ -77,10 +77,10 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
 	c := make(chan ProducerError, numPartitions)
 
 	for partitionIdx, partition := range partitions {
-		go func() {
-			prod, err := newPartitionProducer(client, partition, options, partitionIdx)
+		go func(index int) {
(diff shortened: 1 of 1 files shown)


wolfstudy:
@merlimat PTAL, thanks.

PR #37, opened 2019-08-01
[issue:34] Fix zlib compress Decompress logic

### LINKED ISSUE #34: In ZLIB compress, we shouldn't ignore error info
#### Expected behavior

pkg/compress/zlib.go

```
func (zlibProvider) Compress(data []byte) []byte {
	var b bytes.Buffer
	w := zlib.NewWriter(&b)
	_, err := w.Write(data)
	if err != nil {
		return nil
	}
	err = w.Close()
	if err != nil {
		return nil
	}

	return b.Bytes()
}

func (zlibProvider) Decompress(compressedData []byte, originalSize int) ([]byte, error) {
	r, err := zlib.NewReader(bytes.NewBuffer(compressedData))
	if err != nil {
		return nil, err
	}

	uncompressed := make([]byte, originalSize)
	for {
		_, err = r.Read(uncompressed)
		if err == io.EOF {
			break
		}
		if err != nil && err != io.EOF {
			return nil, err
		}
	}
	err = r.Close()
	if err != nil {
		return nil, err
	}
	return uncompressed, nil
}

}
```

In here, we should handle error.

#### Actual behavior

pkg/compress/zlib.go

```
func (zlibProvider) Compress(data []byte) []byte {
	var b bytes.Buffer
	w := zlib.NewWriter(&b)
	w.Write(data)
	w.Close()

	return b.Bytes()
}
```

```
func (zlibProvider) Decompress(compressedData []byte, originalSize int) ([]byte, error) {
	r, err := zlib.NewReader(bytes.NewBuffer(compressedData))
	if err != nil {
		return nil, err
	}

	uncompressed := make([]byte, originalSize)
	r.Read(uncompressed)
	r.Close()

	return uncompressed, nil
}
```

In here, we can't ignore error information, otherwise, even if compression errors occur, we will not be able to catch errors information.

#### Steps to reproduce

#### System configuration
**Pulsar version**: x.y

wolfstudy:

Fixes #34 

### Motivation

In `reader.Read()`, only a portion of the data will be read at a time. Here we should read all the data until it encounters `io.EOF` and then returns.

This problem has not been reproduced in the previous Test case, because we ignored the processing of error, so whether it succeeds or not, he returns correctly.

### DIFF
--- a/pkg/compression/zlib.go
+++ b/pkg/compression/zlib.go
@@ -37,12 +37,11 @@ func (zlibProvider) CanCompress() bool {
 func (zlibProvider) Compress(data []byte) []byte {
 	var b bytes.Buffer
 	w := zlib.NewWriter(&b)
-	_, err := w.Write(data)
-	if err != nil {
+
+	if _, err := w.Write(data); err != nil {
 		return nil
 	}
-	err = w.Close()
-	if err != nil {
+	if err := w.Close(); err != nil {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


wolfstudy:
@merlimat @jiazhai @sijie PTAL

wolfstudy:
Please merge https://github.com/apache/pulsar-client-go/pull/38 before you merge the pull request

PR #38, opened 2019-08-01
Fix the openjdk-8-jre-headless can't download

wolfstudy:

### Motivation

Seems Integration tests failed for reason:

```
E: Package `openjdk-8-jre-headless` has no installation candidate
```

We can't download  `openjdk-8-jre-headless` lib for integration tests, so we choose to replace a way to install java8 environment.


wolfstudy:
@merlimat @sijie @jiazhai PTAL

PR #56, opened 2019-08-15
[Issue:30] Support seek logic and add test case

### LINKED ISSUE #30: Talk about `grabcnx` logic

In `grabcnx()`, we encapsulate the logic of lookupTopic with the specific BaseCmd request. This will cause both the producer and the consumer to be implemented. 

Can we use it as a function so that the producer and consumer can share the same function?

wolfstudy:

Fixes #30 

### Motivation

- Support seek logic and add test case.

- Fix chan block issue: In go, the unbuffered chan is a blocked channel. For blocked chan, it is required that someone must wait for reception before sending data to chan. In `partitionConsumer` and `partitionProducer`, the `eventsChan` is blocked when send data to it. Turn to use un-blocked chan.


wolfstudy:
@sijie @jiazhai @merlimat PTAL thanks.

wolfstudy:
@sijie @jiazhai @merlimat PTAL thanks.

PR #81, opened 2019-11-02
Add reconnect timeout after unload topic

tuteng:
Currently, for a producer that is sending data, if the topic is unload, the following exception will be thrown, resulting in reconnection failure
```
ERRO[0094] Error: ServiceNotReady, Error Message: Topic is temporarily unavailable  local_addr="127.0.0.1:55800" remote_addr="pulsar://localhost:6650"
```

Reproduce:
```
1. pulsar-perf produce -s 10 test-topic
2. ./bin/pulsar-admin topics unload test-topic
```

### DIFF
--- a/pulsar/impl_partition_producer.go
+++ b/pulsar/impl_partition_producer.go
@@ -176,6 +176,8 @@ func (p *partitionProducer) ConnectionClosed() {
 func (p *partitionProducer) reconnectToBroker() {
 	p.log.Info("Reconnecting to broker")
 	backoff := internal.Backoff{}
+	// Delay one secnd to reconnect
+	time.Sleep(1 * time.Second)
 	for {
 		if p.state != producerReady {
 			// Producer is already closing
--- a/pulsar/impl_partition_producer.go
+++ b/pulsar/impl_partition_producer.go
@@ -176,7 +176,7 @@ func (p *partitionProducer) ConnectionClosed() {
 func (p *partitionProducer) reconnectToBroker() {
 	p.log.Info("Reconnecting to broker")
 	backoff := internal.Backoff{}
-	// Delay one secnd to reconnect
+	// Delay one second to reconnect
 	time.Sleep(1 * time.Second)
 	for {
 		if p.state != producerReady {


pulsar/impl_partition_producer.go
```diff
@@ -176,6 +176,8 @@ func (p *partitionProducer) ConnectionClosed() {
 func (p *partitionProducer) reconnectToBroker() {
 	p.log.Info("Reconnecting to broker")
 	backoff := internal.Backoff{}
+	// Delay one second to reconnect
+	time.Sleep(1 * time.Second)
```
sijie: @tuteng I am not sure this is a right fix. It seems to me that we didn't handle the exception `ServiceNotReady` well. I think we need to add logic to handle `ServiceNotReady` and reconnect.

merlimat:
Is it just printing the error message, right? The reconnection logic should go through after that. 

I think the initial sleep is actually correct, in the sense that we don’t want to immediately aggressively reconnect. The chances of the topic being ready immediately after a failover are thin. 

Though the 1sec sleep is too big. It will impose that latency spike on the produced messages. In the Java and C++ clients we’re actually using a 100ms initial back off delay, and then double it after that. 

One other option is to convert the loop into first sleeping and then trying, rather than the reverse

tuteng:
> Is it just printing the error message, right? The reconnection logic should go through after that.
> 
> I think the initial sleep is actually correct, in the sense that we don’t want to immediately aggressively reconnect. The chances of the topic being ready immediately after a failover are thin.
> 
> Though the 1sec sleep is too big. It will impose that latency spike on the produced messages. In the Java and C++ clients we’re actually using a 100ms initial back off delay, and then double it after that.
> 
> One other option is to convert the loop into first sleeping and then trying, rather than the reverse

What I have observed now is that after the topic unload, it tries to reconnect. As a result, the broker-side Topic is in a fence state, causing the reconnection to fail. Then produce hang lives and do not send data. I did not have this problem using the java client

pulsar/impl_partition_producer.go
tuteng: I'm not sure if there is logic for multiple retries after a failure on the go client-side. in my test, I found that after the topic was unload, the topic on the broker side was in the fence state, but the client did not seem to be performing multiple retries, the producer side was blocked.
You may be right, correctly handling this error `ServiceNotReady` and then try again many times should be able to solve this problem.

sijie: Can we step back and figure out if the client handles `ServiceNotReady` correctly before fixing this by adding time before reconnecting?

merlimat:
> As a result, the broker-side Topic is in a fence state, causing the reconnection to fail 

That's expected behavior. 

I think what's failing is the client side handling of the graceful topic close. 

Just double-checked that if there's a broker crash, the reconnect logic works as expected while it gets stuck in the graceful close (eg: unload).

merlimat:
eg: 

```
INFO[08:01:42.638] Stats - Publish rate:  100.0 msg/s -    0.8 Mbps - Latency ms: 50%   4.0 - 95%   4.6 - 99%   4.6 - 99.9%   4.6 - max    4.6
INFO[08:01:52.639] Stats - Publish rate:  100.0 msg/s -    0.8 Mbps - Latency ms: 50%   4.5 - 95%   4.7 - 99%   4.8 - 99.9%   4.8 - max    4.8
INFO[08:01:56.178] Broker notification of Closed consumer: 1     local_addr="127.0.0.1:61650" remote_addr="pulsar://localhost:6650"
INFO[08:01:56.178] Reconnecting to broker                        name=standalone-75-9 topic="persistent://public/default/my-topic"
ERRO[08:01:56.193] Error: ServiceNotReady, Error Message: Topic is temporarily unavailable  local_addr="127.0.0.1:61650" remote_addr="pulsar://localhost:6650"
INFO[08:01:56.193] Connection closed                             local_addr="127.0.0.1:61650" remote_addr="pulsar://localhost:6650"
INFO[08:01:56.193] Error reading from connection                 error="Short read when reading frame size" local_addr="127.0.0.1:61650" remote_addr="pulsar://localhost:6650"
INFO[08:02:02.641] Stats - Publish rate:   30.0 msg/s -    0.2 Mbps - Latency ms: 50%   3.7 - 95%   4.4 - 99%   4.4 - 99.9%   4.4 - max    4.4
INFO[08:02:12.641] Stats - Publish rate:    0.0 msg/s -    0.0 Mbps - Latency ms: 50%   0.0 - 95%   0.0 - 99%   0.0 - 99.9%   0.0 - max    0.0
INFO[08:02:22.638] Stats - Publish rate:    0.0 msg/s -    0.0 Mbps - Latency ms: 50%   0.0 - 95%   0.0 - 99%   0.0 - 99.9%   0.0 - max    0.0
```

tuteng:
> > As a result, the broker-side Topic is in a fence state, causing the reconnection to fail
> 
> That's expected behavior.
> 
> I think what's failing is the client side handling of the graceful topic close.
> 
> Just double-checked that if there's a broker crash, the reconnect logic works as expected while it gets stuck in the graceful close (eg: unload).



> > As a result, the broker-side Topic is in a fence state, causing the reconnection to fail
> 
> That's expected behavior.
> 
> I think what's failing is the client side handling of the graceful topic close.
> 
> Just double-checked that if there's a broker crash, the reconnect logic works as expected while it gets stuck in the graceful close (eg: unload).

Yes, I also agree with your point that the broker-side topic enters the fence state when the topic is unloaded, but what I want to say is that after the topic is unloaded and the client reconnects, produce should normally send data. 
As you mentioned above, I think the delay can be set to be smaller.  I will refer to the configuration of java and c++ clients

merlimat:
A fix for this issue is in #83

wolfstudy:
@tuteng Hello, the issue has been fixed in #83

PR #83, opened 2019-11-03
Handle RPC errors in reconnections

merlimat:
### Motivation

We're not currently handling the `ServerError` responses in the `RpcClient`. That leaves the caller hanging when any request fails on broker side.

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"net/url"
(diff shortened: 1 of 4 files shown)


pulsar/internal/connection.go
```diff
@@ -424,7 +422,24 @@ func (c *connection) handleResponse(requestID uint64, response *pb.BaseCommand)
...
+}
+
+func (c *connection) handleResponseError(serverError *pb.CommandError) {
+	requestID := serverError.GetRequestId()
+	c.mapMutex.RLock()
+	request, ok := c.pendingReqs[requestID]
+	if !ok {
+		c.log.Warnf("Received unexpected error response for request %d of type %s",
+			requestID, serverError.GetError())
+		return
```
cckellogg: Do we need to unlock the mapMutex before returning here?

pulsar/internal/connection.go
```diff
@@ -424,7 +422,24 @@ func (c *connection) handleResponse(requestID uint64, response *pb.BaseCommand)
 
 	delete(c.pendingReqs, requestID)
 	c.mapMutex.RUnlock()
-	request.callback(response)
+	request.callback(response, nil)
+}
+
+func (c *connection) handleResponseError(serverError *pb.CommandError) {
+	requestID := serverError.GetRequestId()
+	c.mapMutex.RLock()
```
cckellogg: Should we get a Lock() here since we might be deleting and entry from the map?

merlimat: Uhm, this locking was introduced in a7c1a045. 

I think that was not a correct fix. All the internal methods should be accessed only from the connection goroutine. If that was not the case, we should rather fix that aspect.

PR #84, opened 2019-11-04
Fixed locking between connection_reader and connection

merlimat:
### Motivation

Fixed locking around the `mapMutex` in consumer. The current behavior is mixing read-write and read-only locks even when modifying the map. 

The `pendingReqs` was really meant to be accessed by a single go-routine. Changed the logic to pass the received commands on a channel, to be processed in the connection's own go-routine.

cc/ @cckellogg

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -66,7 +66,7 @@ type Connection interface {
 }
 
 type ConsumerHandler interface {
-	MessageReceived(response *pb.CommandMessage, headersAndPayload []byte) error
+	MessageReceived(response *pb.CommandMessage, headersAndPayload Buffer) error
 
 	// ConnectionClosed close the TCP connection.
 	ConnectionClosed()
@@ -107,6 +107,11 @@ type request struct {
 	callback func(command *pb.BaseCommand, err error)
 }
 
+type incomingCmd struct {
+	cmd               *pb.BaseCommand
+	headersAndPayload Buffer
+}
+
 type connection struct {
 	sync.Mutex
 	cond  *sync.Cond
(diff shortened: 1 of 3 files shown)


wolfstudy:
The package of this Buffer structure is more similar to the way Java buffer is used, but may not be very friendly to developers of the go community.

pulsar/internal/connection.go
```diff
@@ -129,9 +134,9 @@ type connection struct {
 	requestIDGenerator uint64
 
 	incomingRequestsCh chan *request
+	incomingCmdCh     chan *incomingCmd
 	writeRequestsCh    chan []byte
 
-	mapMutex    sync.RWMutex
 	pendingReqs map[uint64]*request
```
cckellogg: Maybe add a comment on how this should only be accessed in certain places?

merlimat:
> In here, i am a bit confused. Go officially provided bytes.Buffer can not meet our needs? Why do we have to repackage a Buffer?

We need more functionalities like efficiently reading int32 and int16 while encapsulating the details.

> The package of this Buffer structure is more similar to the way Java buffer is used, but may not be very friendly to developers of the go community.

The ideas of Buffer are taken from Netty ByteBuf (rather than Java ByteBuffer) because the API is an excellent example of good abstractions for working with buffers and parsing/writing network protocols. The abstractions are not language specific.

pulsar/internal/connection.go
wolfstudy: Agree with @cckellogg 

> The pendingReqs was really meant to be accessed by a single go-routine.

`pendingReqs` not concurrently safe, so it is necessary for the user a comment.

merlimat:
run integration tests

PR #86, opened 2019-11-11
Simplify and refactor parts of  the single topic consumer.

cckellogg:
### Motivation

This simplifies the consumer api and refactors the internals of how a single consumer works. I think we should start with a simple api and get all of that functionality working and well tested before adding more complexity and features.

The goal of the refactor was to simplify the code and make more it more readable. The main idea is that each partition consumer will run two go routines. On go routine will be an event loop to process events from the broker and connection events. The second go routine will be a message dispatcher that will manage its internal queue. This dispatcher go routine will output messages on a shared message channel and while its internal queue is drained it will ask the broker for more messages.


pulsar/consumer_impl.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+	if errMsg != "" {
+		return fmt.Errorf(errMsg)
+	}
+	return nil
+}
+
+func (c *consumer) Receive(ctx context.Context) (message Message, err error) {
+	//fmt.Println("recv chan", c.messageCh)
```
merlimat: Remove commented line (or convert into log.debug)

pulsar/consumer.go
```diff
@@ -139,46 +141,15 @@ type Consumer interface {
...
-	AckCumulativeID(MessageID) error
-
 	// Close the consumer and stop the broker to push more messages
 	Close() error
-
-	// Seek reset the subscription associated with this consumer to a specific message id.
-	// The message id can either be a specific message or represent the first or last messages in the topic.
-	// Note: this operation can only be done on non-partitioned topics. For these, one can rather perform the
-	//       seek() on the individual partitions.
-	Seek(msgID MessageID) error
```
merlimat: I'd leave the `Seek()` operation on, since this is commonly used by apps.

pulsar/consumer.go
```diff
@@ -139,46 +141,15 @@ type Consumer interface {
 	// This calls blocks until a message is available.
 	Receive(context.Context) (Message, error)
 
-	// ReceiveAsync appends the message to the msgs channel asynchronously.
-	ReceiveAsync(ctx context.Context, msgs chan<- ConsumerMessage) error
-
-	// ReceiveAsyncWithCallback returns a callback containing the message and error objects
-	ReceiveAsyncWithCallback(ctx context.Context, callback func(msg Message, err error))
+	// Messages returns a channel to consumer messages from
+	Messages() <-chan ConsumerMessage
```
merlimat: Wouldn't be more intuitive to call this `Chan()` ?

pulsar/consumer.go
```diff
@@ -91,7 +93,7 @@ type ConsumerOptions struct {
 
 	// InitialPosition at which the cursor will be set when subscribe
 	// Default is `Latest`
-	SubscriptionInitPos InitialPosition
+	SubscriptionInitPos SubscriptionInitialPosition
```
merlimat: We could leave the variable name out here since it's the same as the enum type.

pulsar/consumer_partition.go
```diff
@@ -0,0 +1,508 @@
...
+		topic:          options.topic,
+		consumerID:     client.rpcClient.NewConsumerID(),
+		partitionIdx:   options.partitionIdx,
+		eventsCh:       make(chan interface{}, 3),
+		queueSize:      int32(options.receiverQueueSize),
+		queueCh:        make(chan []*message, options.receiverQueueSize),
+		connectedCh:    make(chan struct{}),
+		messageCh:      messageCh,
+		closeCh:        make(chan struct{}),
+		log:            log.WithField("topic", options.topic),
```
merlimat: We could also set more fields on the logger, like the subscription name and consumer name

pulsar/consumer_partition.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+func (pc *partitionConsumer) internalClose(req *closeRequest) {
+	defer close(req.doneCh)
+	if pc.state != consumerReady {
+		return
+	}
+
+	pc.state = consumerClosing
+	pc.log.Infof("Closing consumer=%d", pc.consumerID)
```
merlimat: Since we're repeating consumer Id, we should put the field directly in the logger, to ensure consistency. 

In this case though, the consumer id does not make any sense to the application, rather it's an internal id and I'd prefer to actually just hide it (for other purposes than client lib debug).

pulsar/consumer_partition.go
```diff
@@ -0,0 +1,508 @@
...
+
+	requestID := pc.client.rpcClient.NewConsumerID()
+	cmdClose := &pb.CommandCloseConsumer{
+		ConsumerId: proto.Uint64(pc.consumerID),
+		RequestId:  proto.Uint64(requestID),
+	}
+	_, err := pc.client.rpcClient.RequestOnCnxNoWait(pc.conn, requestID, pb.BaseCommand_CLOSE_CONSUMER, cmdClose)
```
merlimat: For closing consumer, we need to wait until the operation is done on the broker side.

pulsar/internal/connection.go
```diff
@@ -286,7 +285,10 @@ func (c *connection) run() {
 			if req == nil {
 				return
 			}
-			c.pendingReqs[req.id] = req
+			// does this request expect a response?
+			if req.id != RequestIDNoResponse {
```
merlimat: Rather than having a special value (for users), I'd prefer for the `..NoWait()` variant to not take a request id since it will never be used.

pulsar/consumer.go
cckellogg: Will add it back

pulsar/consumer.go
cckellogg: What about MessageChan() or Chan() works too?

pulsar/consumer.go
cckellogg: I was trying to be more descriptive for example:
```
pulsar.InitialPosition
vs
pulsar.SubscriptionInitialPosition
```
The second one seemed clearer to me. Thoughts?

pulsar/internal/connection.go
cckellogg: That makes sense. Should we make that change in the current pr or do a new one?

merlimat: New PR is fine

PR #89, opened 2019-11-11
Use internal.Buffer() for message metadata parsing

merlimat:
### Motivation

Unify the parsing of buffer through same interface.


merlimat:
retest this please

PR #91, opened 2019-11-11
Make starting the pprof server optional.

cckellogg:
### Motivation
Make starting the pprof server option when running produce and consume commands.


cckellogg:
run integration tests

PR #93, opened 2019-11-12
Move util files to internal so they are not exported.

cckellogg:
### Motivation

Move blocking queue and semaphore to the internal package so they are not exported with the client. These are only for internal purposes.


cckellogg:
run integration tests

PR #98, opened 2019-11-14
Implemented decompression on consumer side

merlimat:
### Motivation

Implemented the batch decompression on the consumer side

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"fmt"
+	"github.com/apache/pulsar-client-go/pkg/compression"
 	"math"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pulsar/consumer_partition.go
```diff
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"fmt"
+	"github.com/apache/pulsar-client-go/pkg/compression"
```
cckellogg: nitpick group the pulsar import together

pulsar/consumer_partition.go
```diff
@@ -220,10 +230,19 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 	reader := internal.NewMessageReader(headersAndPayload)
 	msgMeta, err := reader.ReadMessageMetadata()
 	if err != nil {
-		// TODO send discardCorruptedMessage
+		pc.discardCorruptedMessage(pbMsgID, pb.CommandAck_ChecksumMismatch)
 		return err
 	}
 
+	uncompressedHeadersAndPayload, err := pc.Decompress(msgMeta, headersAndPayload)
```
cckellogg: is there a way to implement this logic in the message reader?

merlimat: Main reason to have this outside is to handle the proper "decompress error" without the need to check the error types.

PR #101, opened 2019-11-16
Add client function to list topics in a namespace.

cckellogg:
### Motivation

This patch is needed to implement the regex consumer. Doing this in a separate patch to make reviews easier.

### DIFF
--- a/pulsar/client_impl_test.go
+++ b/pulsar/client_impl_test.go
@@ -19,10 +19,9 @@ package pulsar
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"io/ioutil"
 	"testing"
-
-	"github.com/stretchr/testify/assert"
 )
 
 func TestClient(t *testing.T) {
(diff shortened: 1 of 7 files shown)


pulsar/client_impl_test.go
```diff
@@ -222,3 +221,84 @@ func TestTopicPartitions(t *testing.T) {
...
+	// fetch from namespace that does not exist
+	name := generateRandomName()
+	topics, err := ci.namespaceTopics(fmt.Sprintf("%s/%s", name, name))
+	assert.Equal(t, 0, len(topics))
+}
+
+func TestNamespaceTopics(t *testing.T) {
+	name := generateRandomName()
+	namespace := fmt.Sprintf("public/%s", name)
+	namespaceUrl := fmt.Sprintf("http://localhost:8080/admin/v2/namespaces/%s", namespace)
```
merlimat: In `test_helper.go` we have: 

```
	serviceURL    = "pulsar://localhost:6650"
	serviceURLTLS = "pulsar+ssl://localhost:6651
```

we could add the `http://localhost:8080` there as well

Issue #107, opened 2019-11-27
Producer deadlocks when pulsar downtime


hailin0:
#### Expected behavior

Producer works normally after pulsar restart

#### Actual behavior

producer#flush deadlock after pulsar restart

#### Steps to reproduce

	for  {
		asyncMsg := pulsar.ProducerMessage{
			Payload: []byte(fmt.Sprintf("async-message")),
		}

		producer.SendAsync(ctx, &asyncMsg, func(mid pulsar.MessageID, msg *pulsar.ProducerMessage, err error) {
			if err != nil { log.Fatal(err) }
		})

		// deadlock
		producer.Flush() 
	}

=======================
Running the above program and restarting pulsar will cause producer#flush deadlock

#### System configuration
**Pulsar version**: latest

jiazhai:
Thanks @reugn for helping on this issue

hailin0:
@reugn 
I tested [pr 108](https://github.com/apache/pulsar-client-go/pull/108) but it didn't recover

wolfstudy:
@reugn Maybe the issue is similar to #124. Can you try it again?

PR #108, opened 2019-11-28
[Issue #107] SetWriteDeadline on connection

reugn:
Issue #107 

### Modifications
SetWriteDeadline on connection.

### Verifying this change

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -100,6 +100,7 @@ func (s connectionState) String() string {
 }
 
 const keepAliveInterval = 30 * time.Second
+const writeDeadline = 15 * time.Second
 
 type request struct {
 	id       *uint64
@@ -309,7 +310,7 @@ func (c *connection) WriteData(data []byte) {
 
 func (c *connection) internalWriteData(data []byte) {
 	c.log.Debug("Write data: ", len(data))
-	err := c.cnx.SetWriteDeadline(time.Now().Add(time.Second * 15))
+	err := c.cnx.SetWriteDeadline(time.Now().Add(writeDeadline))
 
 	if err != nil {
 		c.log.WithError(err).Warn("Failed to SetWriteDeadline")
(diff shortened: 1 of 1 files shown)


pulsar/internal/connection.go
```diff
@@ -309,7 +309,13 @@ func (c *connection) WriteData(data []byte) {
 
 func (c *connection) internalWriteData(data []byte) {
 	c.log.Debug("Write data: ", len(data))
-	if _, err := c.cnx.Write(data); err != nil {
+	err := c.cnx.SetWriteDeadline(time.Now().Add(time.Second * 15))
```
cckellogg: Should this be set once per connection or once per write? Also, this should be a variable and not hardcoded so we can make it configurable in the future if needed.

reugn: Makes sense. Done.

pulsar/internal/connection.go
```diff
@@ -309,7 +310,13 @@ func (c *connection) WriteData(data []byte) {
 
 func (c *connection) internalWriteData(data []byte) {
 	c.log.Debug("Write data: ", len(data))
-	if _, err := c.cnx.Write(data); err != nil {
+	err := c.cnx.SetWriteDeadline(time.Now().Add(writeDeadline))
```
merlimat: Shouldn't this be set just once per each connection? doing it here would set it each time we write on the connection. 

Also, since this will affect the publish timeouts, it would be better to keep it configurable (as a client option). 

Finally, what is the exact meaning of the write deadline? Will it fail the write op if the connection is not writable for the specified amount of time? I think we are already covering this scenario with the keep-alive mechanism that detects either a clogged or stale TCP connection.

reugn: Should be set before each Write.

Connections produced via the connection pool and clients are not aware of the underlying configuration. Exposing such low-level config is redundant, we can increase the timeout though.

If a connection got stuck on Write, it will not reach the ping ticker message. Also, it will not release the Flush WaitGroup (impl_partition_producer.go: 438).

@merlimat  WDYT?

jiazhai:
@merlimat  Would you please help take a review again?

merlimat:
The issue was fixed in #131

PR #109, opened 2019-11-29
Fix subscriber bug and use Consumer interface for multi topic consumer.

cckellogg:
### Motivation

Fixes a bug in the subscriber function and converts the multi topic consumer to use the subscriber function.

### DIFF
--- a/pulsar/consumer_multitopic.go
+++ b/pulsar/consumer_multitopic.go
@@ -51,24 +51,19 @@ func newMultiTopicConsumer(client *client, options ConsumerOptions, topics []str
 	}
 
 	var errs error
-	consumers := make(map[string]Consumer, len(topics))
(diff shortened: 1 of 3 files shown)


pulsar/consumer_multitopic.go
```diff
@@ -45,54 +45,32 @@ func newMultiTopicConsumer(client *client, options ConsumerOptions, topics []str
...
-	}
-
-	go func() {
-		wg.Wait()
-		close(ch)
-	}()
-
 	var errs error
-	for ce := range ch {
+	consumers := make(map[string]Consumer, len(topics))
```
reugn: Why do we need temporary consumers value here?

cckellogg: That's a good question. The idea was to only set the consumers when everything succeeded but I don't think it matters since we return nil on failure. I'll remove it.

PR #113, opened 2019-12-08
Fix #111: set consumer options

### LINKED ISSUE #111: consumer.Subscription() returns "" since options are never set in internalTopicSubscribe()
#### Expected behavior
Subscriber.Subscription() should return the subscription name

#### Actual behavior
returns ""

#### Steps to reproduce

create subscriber with pulsar.ConsumerOptions{
SubscriptionName: "anything-at-all"
}

Then try to call sub.Subscription()

It appears you pass the options correctly when creating the newPartitionConsumer, but you forgot to set them in the consumer you return at: https://github.com/apache/pulsar-client-go/blob/862285fee18a213289062d855156f44823e5a724/pulsar/consumer_impl.go#L198

reugn:
Fixes #111


jiazhai:
Thanks for your contribute @reugn

Issue #112, opened 2019-12-08
Can't install go module.


Jim-Lambert-Bose:
#### Expected behavior
go get github.com/apache/pulsar-client-go@latest

should download and install the latest commit on the master branch

#### Steps to reproduce

$ go get github.com/apache/pulsar-client-go@latest
go: finding github.com/apache/pulsar-client-go latest
go build github.com/apache/pulsar-client-go: no non-test Go files in /Users/jimlambrt/go/pkg/mod/github.com/apache/pulsar-client-go@v0.0.0-20191205202832-862285fee18a

I think this is happening because you only have one .go file in the parent directory https://github.com/apache/pulsar-client-go  and it's a test file.

#### System configuration
$ go version
go version go1.13.4 darwin/amd64

Jim-Lambert-Bose:
Looks like the module is download and "installed" but the go.mod file is not updated to reflect the new version.

jiazhai:
Thanks for @Jim-Lambert-Bose your reporting. Seems you get the root cause. Would you like to provide a PR to fix it?

reugn:
@Jim-Lambert-Bose why do you need to build it?
And what do you mean by 
> go.mod file is not updated to reflect the new version.

I think this dummy main file is redundant.
Opened a PR #116 to tidy this out a bit.

Jim-Lambert-Bose:
I agree, your solution is better and should fix the issue.  `go get` was able to download @latest version of the module, but since the "build" failed (with only test files in the root dir) the go tooling would not update the go.mod file in my project that used the pulsar client... effectively, the module was "installed" but the tooling didn't auto-magically update the go.mod file... this meant I could test locally, but there's no way I'd be able to use the @latest version in my deployments.

Jim-Lambert-Bose:
Feel free to close my PR since it's superseded by your PR

jiazhai:
cool. Thanks a lot for your contribute

jiazhai:
Since The PR is merged. will close this issue.

grtfou:
Install module fail.  
```
$ go get -u github.com/apache/pulsar/pulsar-client-go/pulsar

go: finding github.com/apache/pulsar/pulsar-client-go/pulsar latest
go: finding github.com/apache/pulsar/pulsar-client-go latest
go: finding golang.org/x/crypto latest
go: finding golang.org/x/sys latest
go: finding golang.org/x/tools latest
go: finding gopkg.in/check.v1 latest
[...]

PR #115, opened 2019-12-09
[Issue #112] Fix installation of go module

Jim-Lambert-Bose:
### Motivation

Fix the installation of the client's go module: github.com/apache/pulsar-client-go
With this change, no one can install the client using the standard `go get` tool chain

https://github.com/apache/pulsar-client-go/issues/112

### Modifications

Added a no_code.go source file to the root directory.  This is required the module currently won't install because the only source in the root directory is licence_test.go (which walks the source tree looking for license txt in each file).

### Verifying this change

  - The public API:  no
  - The schema: no 
  - The wire protocol: no

### Documentation

No need for documentation besides the comments in the no_code.go source file.  If the project changes how licensing is checked, then this change can be reverted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@
 *.out
 
 perf/perf
+
+
+# dummy executable in root dir
+pulsar-client-go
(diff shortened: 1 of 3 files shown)


Jim-Lambert-Bose:
fixed by: https://github.com/apache/pulsar-client-go/pull/116

PR #116, opened 2019-12-09
[Issue #112] Move license_test from project root; change package name

reugn:
Master Issue: #112 

### Motivation
Dummy main file is not a good solution

### Modifications
1. Move license_test file from project root
2. Change file package
3. go mod tidy

### DIFF
--- a/integration-tests/license_test.go
+++ b/integration-tests/license_test.go
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package main
+package license_test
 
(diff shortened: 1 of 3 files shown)


integration-tests/license_test.go
```diff
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package main
+package license_test
```
cckellogg: Will this fix the go get build issue? Is there a way to test it?

PR #119, opened 2019-12-11
[Issue 118][client_impl.go] fix race condition when accessing client.handlers

Jim-Lambert-Bose:
### Motivation

fix race condition caused by writing to map in multiple goroutines

### Modifications

Created new internal type: ClientHandlers which encapsulates a sync.Mutex to make read/writes to clients.handlers concurrent safe.

### Verifying this change

This change added tests and can be verified as follows:  just run: `go test -timeout 30s github.com/apache/pulsar-client-go/pulsar/internal -run "^(TestClientHandlers)$"`

### Documentation

### DIFF
--- a/pulsar/internal/client_handlers.go
+++ b/pulsar/internal/client_handlers.go
@@ -31,10 +31,10 @@ func NewClientHandlers() ClientHandlers {
 		l:        &sync.RWMutex{},
 	}
 }
-func (h *ClientHandlers) Set(c Closable, b bool) {
+func (h *ClientHandlers) Add(c Closable) {
 	h.l.Lock()
 	defer h.l.Unlock()
-	h.handlers[c] = b
+	h.handlers[c] = true
 }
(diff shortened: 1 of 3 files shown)


pulsar/internal/client_handlers.go
```diff
@@ -0,0 +1,52 @@
...
+	l        *sync.RWMutex
+}
+
+func NewClientHandlers() ClientHandlers {
+	return ClientHandlers{
+		handlers: map[Closable]bool{},
+		l:        &sync.RWMutex{},
+	}
+}
+func (h *ClientHandlers) Set(c Closable, b bool) {
```
merlimat: Since we only need a `Set` in this case (as opposed to a `Map`) we could rename the `Set()` into `Add()` and remove the `bool` parameter.

Jim-Lambert-Bose: Agreed and add a commit to this PR for the change.

Issue #120, opened 2019-12-11
Data race: type connection_pool.go GetConnection() is not concurrent safe


Jim-Lambert-Bose:
## Expected behavior

Internal `func (p *connectionPool) GetConnection(logicalAddr *url.URL, physicalAddr *url.URL) (Connection, error)`  should be concurrent safe.   It reaches into the connection type and reads attrribs protected by a mutex without locking/unlocking

## Actual behavior

Current implementation allows a race condition since there is no call to lock/unlock the sync.Mutex in the connection object while reading its private attribs

https://github.com/apache/pulsar-client-go/blob/dfc17abd753ac2e488021aede74433e681b5df58/pulsar/internal/connection_pool.go#L56
 
## Steps to reproduce
run some tests using the race detector.

I'll submit a PR for this in a few minutes.

wolfstudy:
Thanks @Jim-Lambert-Bose work for this.

I can't reproduce the issue in my local, I will close this issue if you encounter this issue again, please reopen it or please let me know, thanks.

PR #121, opened 2019-12-11
[Issue 120] [connection_pool.go] fixing data race when GetConnection() reads connection attribs

Jim-Lambert-Bose:
# Motivation
https://github.com/apache/pulsar-client-go/issues/120

fixing race condition caused by reading attribs protected by sync.Mutex but mutex was not locked/unlocked

# Modifications
Simply locked/unlocked connection since there's a mutex embedded in it, when reading attribs from the connection

### Verifying this change

### Documentation

### DIFF
--- a/pulsar/internal/connection_pool.go
+++ b/pulsar/internal/connection_pool.go
@@ -53,7 +53,13 @@ func (p *connectionPool) GetConnection(logicalAddr *url.URL, physicalAddr *url.U
 	cachedCnx, found := p.pool.Load(logicalAddr.Host)
 	if found {
 		cnx := cachedCnx.(*connection)
-		log.Debug("Found connection in cache:", cnx.logicalAddr, cnx.physicalAddr)
+		cnx.Lock()
+		// stash these vars, so we can release the lock ASAP
+		// (without using a defer, which holds it until the end of the func)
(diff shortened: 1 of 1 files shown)


pulsar/internal/connection_pool.go
```diff
@@ -53,15 +53,21 @@ func (p *connectionPool) GetConnection(logicalAddr *url.URL, physicalAddr *url.U
 	cachedCnx, found := p.pool.Load(logicalAddr.Host)
 	if found {
 		cnx := cachedCnx.(*connection)
-		log.Debug("Found connection in cache:", cnx.logicalAddr, cnx.physicalAddr)
+		cnx.Lock()
+		// stash these vars, so we can release the lock ASAP
+		// (without using a defer, which holds it until the end of the func)
+		logicalAddr = cnx.logicalAddr
+		physicalAddr = cnx.physicalAddr
+		cnx.Unlock()
```
reugn: @Jim-Lambert-Bose you have no reason to lock and stash here.

Jim-Lambert-Bose: Hum... you're correct.  I can't reproduce the race condition I saw earlier and it could have been related to the other PR I submitted today.  I'll close this.  Ty for the review.

PR #124, opened 2019-12-13
[Issue #123] Ensure message is sent if no room in current batch.

### LINKED ISSUE #76: Transfer data is blocked by flush
To pump data buy filebeat to pulsar cluster , by using go client .

Found that ,after sending message to pulsar successfully, there will be blocked by using flush function.

the code is below ：
```
	p.internalFlushCurrentBatch()

	pi, ok := p.pendingQueue.PeekLast().(*pendingItem)
	if !ok {
		p.log.Error("internalFlush block ,Flush error")
		fr.waitGroup.Done()
		return
	}
	p.log.Info("InternalFlush corrctly: ")  
	pi.sendRequests = append(pi.sendRequests, &sendRequest{
		msg: nil,
		callback: func(id MessageID, message *ProducerMessage, e error) {
			fr.err = e
			fr.waitGroup.Done()
		},
	})
}

func (p *partitionProducer) Send(ctx context.Context, msg *ProducerMessage) error {
	wg := sync.WaitGroup{}
	wg.Add(1)

	var err error

	p.internalSendAsync(ctx, msg, func(ID MessageID, message *ProducerMessage, e error) {
		err = e
		wg.Done()
	}, true)

	// When sending synchronously we flush immediately to avoid
	// the increased latency and reduced throughput of batching
	p.log.Info("-------------------beginning flsuh ----------------")
	if err = p.Flush(); err != nil {
		return err
	}
	p.log.Info("-------------------end flsuh ----------------")
	wg.Wait()
	return err
}
```

log as below
```
time="2019-10-25T10:39:53+08:00" level=info msg="InternalFlush corrctly: " name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="-------------------end flsuh ----------------" name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="-------------------beginning flsuh ----------------" name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="InternalFlush corrctly: " name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="-------------------end flsuh ----------------" name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="-------------------beginning flsuh ----------------" name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="InternalFlush corrctly: " name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="-------------------end flsuh ----------------" name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="-------------------beginning flsuh ----------------" name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
time="2019-10-25T10:39:53+08:00" level=info msg="InternalFlush corrctly: " name=dc_jh1 topic="persistent://crossjhwx/jh2wx/20191024-partition-0"
```
the problem is occured on below section

```
p.log.Info("InternalFlush corrctly: ")  
	pi.sendRequests = append(pi.sendRequests, &sendRequest{
		msg: nil,
		callback: func(id MessageID, message *ProducerMessage, e error) {
			fr.err = e
			fr.waitGroup.Done()
		},
	})
```
no idea the logic ,but make sure the error is occur on it.

### LINKED ISSUE #114: The pulsar go client missing data when transfer
Now I found one problem relate to pulsar go client 

The whole process of mine is   csv file  filebeatpulsar flume hive

At past time I verified the data count ,found that the data by this path missing about 0.46% compare with the path of logstashkafka .

No idea the reason why ,so I tried below steps 

1 Reduce the amount of data from 100% to 50%, excluding data loss caused by too much pressure in the pulsar cluster, the data loss is still 0.46%
2 Reduce the amount of data from 50% to 10%, make the pulsar cluster stress-free, and the data loss is still 0.46%
3 Use bin / pulsar-perf to quantitatively send test test data to the pulsar cluster, which is 100% acceptable, eliminating the problems of the pulsar cluster
4 Quantitatively send test data to the pulsar cluster with filebeat. 99.6% can receive it (send 36077, receive 35091)
 

I put the data file in the attachment. Please take a look at this problem. Now the loss rate is about five thousandths, which is a little too large.

Another thing to say, I use the asynchronous sending of the go client, which has been confirmed with @tuteng

Hopefully your response

### LINKED ISSUE #123: The client batch is full, resulting in data loss or client is blocked
This is the reason for these two problems https://github.com/apache/pulsar-client-go/issues/76 and https://github.com/apache/pulsar-client-go/issues/114.

#### Expected behavior

When the batch is full, the current data should be retried to ensure that the data can be successfully sent.

#### Steps to reproduce

1. SendAsyc is used to send data 
2. ensure that each piece of data is large enough 
3. trigger batch full after multiple pieces of data are sent.

I think there is a problem here https://github.com/apache/pulsar-client-go/blob/master/pulsar/producer_partition.go#L259. If the current data is checked and found to exceed the size of the batch, the `p.batchBuilder.Add` function will return false and then trigger Flush, but after triggering flush, the current data is not processed. Therefore, if the client uses waitgroup to synchronize, it will be blocked in the callback. If there is no synchronization, Then this message will not be sent correctly resulting in data loss, so I think there should be more retries to ensure that the current data is sent correctly when the batch is full, just like java client https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L396

#### System configuration
**Pulsar version**:  master  
go client: master

cckellogg:
Fixes #123
Fixes #76 
Fixes #114 

If the current message being published does not fit in the current batch that message is never sent. This patch will add the message to the batch after flushing the current batch.

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -265,12 +265,18 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
 			if ok := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters); !ok {
 				p.log.WithField("size", len(msg.Payload)).
 					WithField("sequenceID", sequenceID).
-					Error("unable to send message")
+					WithField("properties", msg.Properties).
+					Error("unable to add message to batch")
 			}
 		}
 	} else {
 		// Send individually
-		p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters)
+		if added := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters); !added {
+			p.log.WithField("size", len(msg.Payload)).
+				WithField("sequenceID", sequenceID).
+				WithField("properties", msg.Properties).
+				Error("unable to send single message")
+		}
 		p.internalFlushCurrentBatch()
 	}
 
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -256,10 +256,17 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
 	}
 
 	if sendAsBatch {
-		ok := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters)
-		if ok == false {
+		added := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters)
+		if !added {
 			// The current batch is full.. flush it and retry
 			p.internalFlushCurrentBatch()
+
+			// after flushing try again to add the current payload
+			if ok := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters); !ok {
+				p.log.WithField("size", len(msg.Payload)).
+					WithField("sequenceID", sequenceID).
+					Error("unable to send message")
+			}
 		}
 	} else {
 		// Send individually
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -787,7 +787,8 @@ func TestConsumerMetadata(t *testing.T) {
 		t.Fatal(err)
 	}
 	subs := stats["subscriptions"].(map[string]interface{})
-	meta := subs["my-sub"].(map[string]interface{})["consumers"].([]interface{})[0].(map[string]interface{})["metadata"].(map[string]interface{})
+	cons := subs["my-sub"].(map[string]interface{})["consumers"].([]interface{})[0].(map[string]interface{})
+	meta := cons["metadata"].(map[string]interface{})
 	assert.Equal(t, len(props), len(meta))
 	for k, v := range props {
 		mv := meta[k].(string)
--- a/pulsar/negative_acks_tracker_test.go
+++ b/pulsar/negative_acks_tracker_test.go
@@ -29,9 +29,9 @@ import (
 const testNackDelay = 300 * time.Millisecond
 
 type nackMockedConsumer struct {
-	ch chan messageID
+	ch     chan messageID
 	closed bool
-	lock sync.Mutex
+	lock   sync.Mutex
 	msgIds []messageID
 }
 
@@ -42,7 +42,7 @@ func newNackMockedConsumer() *nackMockedConsumer {
 	go func() {
 		// since the client ticks at an interval of delay / 3
 		// wait another interval to ensure we get all messages
-		time.Sleep(testNackDelay + 101 * time.Millisecond)
+		time.Sleep(testNackDelay + 101*time.Millisecond)
 		t.lock.Lock()
 		defer t.lock.Unlock()
 		t.closed = true
@@ -69,7 +69,7 @@ func sortMessageIds(msgIds []messageID) []messageID {
 	return msgIds
 }
 
-func (nmc *nackMockedConsumer) Wait() <- chan messageID {
+func (nmc *nackMockedConsumer) Wait() <-chan messageID {
 	return nmc.ch
 }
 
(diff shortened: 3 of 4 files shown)


pulsar/producer_partition.go
```diff
@@ -256,10 +256,17 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
...
+
+			// after flushing try again to add the current payload
+			if ok := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters); !ok {
+				p.log.WithField("size", len(msg.Payload)).
+					WithField("sequenceID", sequenceID).
+					Error("unable to send message")
+			}
 		}
 	} else {
 		// Send individually
```
tuteng: I think there may be the same problem here. If false is returned after the https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/batch_builder.go#L107 function is executed.

```
// Add will add single message to batch.
func (bb *BatchBuilder) Add(metadata *pb.SingleMessageMetadata, sequenceID uint64, payload []byte,
	callback interface{}, replicateTo []string) bool {
	if replicateTo != nil && bb.numMessages != 0 {
		// If the current batch is not empty and we're trying to set the replication clusters,
		// then we need to force the current batch to flush and send the message individually
		return false
	} else if bb.msgMetadata.ReplicateTo != nil {
		// There's already a message with cluster replication list. need to flush before next
		// message can be sent
		return false
	} else if bb.hasSpace(payload) {
		// The current batch is full. Producer has to call Flush() to
		return false
	}

	if bb.numMessages == 0 {
		bb.msgMetadata.SequenceId = proto.Uint64(sequenceID)
		bb.msgMetadata.PublishTime = proto.Uint64(TimestampMillis(time.Now()))
		bb.msgMetadata.SequenceId = proto.Uint64(sequenceID)
		bb.msgMetadata.ProducerName = &bb.producerName
		bb.msgMetadata.ReplicateTo = replicateTo

		bb.cmdSend.Send.SequenceId = proto.Uint64(sequenceID)
	}
	addSingleMessageToBatch(bb.buffer, metadata, payload)

	bb.numMessages++
	bb.callbacks = append(bb.callbacks, callback)
	return true
}
```

 This data will not be retried, resulting in incorrect processing. Therefore, I think retry should be added after internalFlushCurrentBatch under this condition.

cckellogg: You think we need to retry in the send individually block as well?

tuteng: Yes, I think it is ok to try again, at least a warning should be issued to inform the user that a message cannot be correctly added to the buffer.

wolfstudy: Agree with @tuteng 

```
if add := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters); !add {
    //print warn log
}
```

wolfstudy: Agree with @tuteng , we need to handle the return value of `p.batchBuilder.Add()`:

```
if add := p.batchBuilder.Add(smm, sequenceID, msg.Payload, request, msg.ReplicationClusters); !add {
    // print warn log
}
```

cckellogg: Sounds good. I'll add a log there.

PR #125, opened 2019-12-17
Check license header of pulsar-client-go files

wolfstudy:

### Modifications

- check license header of `pulsar-client-go` files


cckellogg:
I think this will break the go get command like before. See https://github.com/apache/pulsar-client-go/pull/116

wolfstudy:
> I think this will break the go get command like before. See #116

Thanks @cckellogg the license check already exists in `integration-tests/license_test.go`, will close the pull request. thanks again.

PR #127, opened 2019-12-17
Add release process file for pulsar-client-go

wolfstudy:

### Modifications

- Add release process file for pulsar-client-go

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -1,0 +1,225 @@
+This guide illustrates how to perform a release for Apache pulsar-client-go.
+
+In general, you need to perform the following steps:
+
+1. Create a release branch.
+2. Update the version and tag of a package.
+3. Build and inspect an artifact.
+4. Sign and stage the artifacts.
+5. Write a release note.
+6. Run a vote.
+7. Promote the release.
+8. Update the release note.
+9. Announce the release.
+
+## Steps in detail
+
+1. Create a release branch.
+
+This example creates a branch `branch-0.X.0` based on `master` where tags will be generated and new fixes will be applied as parts of the maintenance for the release. `0.X.0` is the version of the release.
+
+The branch needs only to be created when creating a minor release rather than a patch release.
+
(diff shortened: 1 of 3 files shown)


wolfstudy:
@sijie @tuteng @jiazhai @merlimat PTAL

docs/release-process.md
```diff
@@ -0,0 +1,225 @@
...
+$ git tag -u $USER@apache.org v0.X.0-candidate-1 -m 'Release v0.X.0-candidate-1'
+
+# Push both the branch and the tag to Github repo
+git push origin branch-0.X.0
+git push origin v0.X.0-candidate-1
+```
+
+3. Build and inspect an artifact.
+
+Generate a release candidate package.
```
merlimat: What's the point of releasing a src package? All the users of this library are going to fetch the sources from the Github repo in any case. 

I think we should be fine by providing an "official" tag.

docs/release-process.md
```diff
@@ -0,0 +1,225 @@
...
+If you have not installed Go, install it according to the [installation instruction](http://golang.org/doc/install).
+
+> NOTE: Since the `go mod` package management tool is used in this project, **Go 1.11 or higher** version is required.
+
+After the build, inspect the artifacts:
+
+Run the following command to verify the license headers in the source files:
+
+```
+go test license_test.go
```
merlimat: All these tests are being automatically run for every PR. I don't see much value in repeating them in manual form.

docs/release-process.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+Run the standalone Pulsar service and check that the Go client can produce and consume messages correctly:
+
+```
+$ cd examples/consumer
```
merlimat: Same here, we're running a more extensive set of tests for each commit. I don't think these manual steps will be useful in catching potential problems.

docs/release-process.md
wolfstudy: Yes, agree with you. will remove the step.

wolfstudy:
@merlimat PTAL again, thanks.

PR #130, opened 2019-12-18
Allow to specify a function that supply a token

merlimat:
### Motivation

Similarly to the C++ based Go client and to Java, C++ and Py APIs, we need to allow application to specify a function provider for the token itself. This is required in all scenarios where tokens are refreshed and not written on a file.


merlimat:
retest this please

pkg/auth/token.go
```diff
@@ -40,7 +40,8 @@ func NewAuthenticationTokenWithParams(params map[string]string) (Provider, error
 	}
 }
 
-// NewAuthenticationToken return a interface of Provider with a string token.
+// Return a token auth provider that will use the specified token to
+// talk with Pulsar brokers
```
wolfstudy: Externally exposed functions (aka, capitalized) need to start with the function name when adding a comment.  e.g: 

```
NewAuthenticationToken return a token ....
```

Otherwise these comments will not be displayed in `godoc`

pkg/auth/token.go
```diff
@@ -52,6 +53,15 @@ func NewAuthenticationToken(token string) Provider {
 	}
 }
 
+// Return a token auth provider that get the token data from a user supplied
+// function. The function is invoked each time the client library needs to
+// use a token in talking with Pulsar brokers
```
wolfstudy: As same above.

```
NewAuthenticationTokenFromSupplier return a token auth provider .....
```

pulsar/client_impl_test.go
```diff
@@ -174,6 +174,30 @@ func TestTokenAuth(t *testing.T) {
...
+func TestTokenAuthWithSupplier(t *testing.T) {
+	client, err := NewClient(ClientOptions{
+		URL:            serviceURL,
+		Authentication: NewAuthenticationTokenFromSupplier(func() (s string, err error) {
+			token, err := ioutil.ReadFile(tokenFilePath)
+			if err != nil {
+				return "", err
+			} else {
+				return string(token), nil
+			}
```
wolfstudy:
```
if err != nil {
	return "", err
}

return string(token), nil
```

Maybe we can remove `else` branch, right?

pulsar/client.go
```diff
@@ -39,6 +39,13 @@ func NewAuthenticationToken(token string) Authentication {
 	return auth.NewAuthenticationToken(token)
 }
 
+// Return a token auth provider that get the token data from a user supplied
+// function. The function is invoked each time the client library needs to
+// use a token in talking with Pulsar brokers
```
wolfstudy: As same above.

```
NewAuthenticationTokenFromSupplier return a token auth provider .....
```

pkg/auth/token.go
merlimat: Got it, I didn't know it was expecting that. Will fix all of them.

pulsar/client_impl_test.go
merlimat: I don't see much difference between the 2 approaches

wolfstudy: @merlimat Yes, there is no difference between the 2 approaches. But in `golint`, the following writing is recommended. 

Currently, ci is ok, because we didn't add the check to `.golangci.yml`

merlimat: Got it. Changed

PR #131, opened 2019-12-19
Fixed detection and handling of stale connections

### LINKED ISSUE #107: Producer deadlocks when pulsar downtime

merlimat:
### Motivation

This is fixing an issue in which the Ping task is getting stalled when the connection is not writable (eg: network partition with the broker). The checking of connection staleness needs to be done in its own goroutine to ensure it's running even when the write/read routines are blocked.

This makes sure clients can successfully recover from partitions and failures with the brokers, also during the initial handshake. 

Verified that this fixes #107

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -62,6 +62,7 @@ type Connection interface {
 	UnregisterListener(id uint64)
 	AddConsumeHandler(id uint64, handler ConsumerHandler)
 	DeleteConsumeHandler(id uint64)
+	ID() string
 	Close()
 }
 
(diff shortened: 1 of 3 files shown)


wolfstudy:
@merlimat The #108 seems to be fixing the same issue, right? If yes, when the pull request is merged, please close it.

PR #132, opened 2019-12-19
Move pkg stuff under pulsar/internal

merlimat:
### Motivation

I know this was changed in #7, though my concern is that everything under `pkg` is currently publicly visible and thus becomes part of public APIs. 

Since all the code in there is meant to be strictly private, we should make sure this is easily enforced.


wolfstudy:
@merlimat the `../pulsar/internal/pb/PulsarApi.pb.go` license header check fail, maybe we should ignore it.

```
--- FAIL: TestLicense (0.17s)
    license_test.go:90: ../pulsar/internal/pb/PulsarApi.pb.go: license header not present
FAIL
```

In `integration-tests/license_test.go`:

```
var skip = map[string]bool{
	"../pkg/pb/PulsarApi.pb.go": true,
}
```

please fix the path.

PR #134, opened 2019-12-19
Exit goroutine when signal is caught.

cckellogg:
This change ensures that the go routine listening for the interrupt exists properly. It work before because we exited the program whenever we received a signal from this channel. This change just ensures proper clean up so we don't ever leak a go routine.


wolfstudy:
@cckellogg please provide some description for this change, thanks.

PR #138, opened 2019-12-20
Enable golangci lint for project

wolfstudy:

### Modifications

Enable golangci lint for project, checks as follows:

```
    - bodyclose
    - deadcode
    - gocritic
    - goimports
    - golint
    - gosimple
    - govet
    - ineffassign
    - interfacer
    - misspell
    - staticcheck
    - structcheck
    - stylecheck
    - typecheck
    - unconvert
    - unparam
    - unused
    - varcheck
    - lll
    - prealloc
    - maligned
```

### DIFF
--- a/run-ci.sh
+++ b/run-ci.sh
@@ -30,8 +30,7 @@ go build ./pulsar
 go build -o pulsar-perf ./perf
 
 # Project code style check
-wget -O - -q https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s v1.18.0
-./bin/golangci-lint run -c ./.golangci.yml ./...
+docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.21.0 golangci-lint run -c ./.golangci.yml ./...
 
 ./pulsar-test-service-start.sh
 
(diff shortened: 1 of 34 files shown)


run-ci.sh
```diff
@@ -29,6 +29,10 @@ go mod download
 go build ./pulsar
 go build -o pulsar-perf ./perf
 
+# Project code style check
+wget -O - -q https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s v1.18.0
```
merlimat: This should go in the Docker image so that we can cache it

wolfstudy:
- `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)

```
if err != nil {
		return nil, err
	} else {
		return internal.NewBufferWrapper(uncompressed), nil
	}
```

- comment on exported var `FlagProfile` should be of the form `FlagProfile ...` (golint)

```
// global flags
var FlagProfile bool
```

run-ci.sh
wolfstudy: Good idea @merlimat, has been fixed, please take a look again, thanks.

wolfstudy:
@merlimat PTAL thanks

Issue #145, opened 2019-12-26
producer cannot connect to broker through pulsar proxy


freeznet:
pulsar-client-go cannot connect to pulsar broker through pulsar proxy, the connection failed when send `PRODUCER` command to pulsar proxy, proxy will throw `UnsupportedOperationException`, and client got `Short read when reading frame size` error.  

After code review, I found out that once client get `LOOKUP` result with `ProxyThroughServiceUrl == true`, client should open a new connection to proxy with `CONNECT` command and `ProxyToBrokerUrl` be set. Then according to https://github.com/apache/pulsar/blob/master/pulsar-proxy/src/main/java/org/apache/pulsar/proxy/server/ProxyConnection.java#L207 , pulsar proxy will tunnel commands to broker. 

Currently, `pulsar-client-go` ignored `LOOKUP` result if `ProxyThroughServiceUrl` is true, then all commands are handled by pulsar proxy and not tunneled to broker.

PR #146, opened 2019-12-26
[Issue #145] fix producer cannot connect to broker through pulsar proxy

freeznet:
Fixes: #145 

if producer connect to broker through pulsar proxy, producer will get `Short read when reading frame size` error because pulsar proxy not pass commands to broker.


merlimat:
@freeznet @wolfstudy This change is not correct. The physicalAddress and logicalAddress are the way for a client to know where to connect.

Issue #153, opened 2019-12-31
Memory leaks when producer/consumer close


freeznet:
#### Expected behavior

When `producer.Close()` or `consumer.Close()`, all related memory should be freed. 

#### Actual behavior

When we rapidly start and close multiple producers (or consumers), memory cost of the process are becomes very large. `pprof` results as follows:

```
$ go tool pprof -base ./mem10.out mem99.out
Type: inuse_space
Time: Dec 30, 2019 at 5:32pm (CST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 4017.42kB, 100% of 4017.42kB total
Showing top 10 nodes out of 13
      flat  flat%   sum%        cum   cum%
 2600.21kB 64.72% 64.72%  2600.21kB 64.72%  github.com/apache/pulsar-client-go/pulsar/internal.NewBlockingQueue
  902.59kB 22.47% 87.19%   902.59kB 22.47%  compress/flate.NewWriter
  514.63kB 12.81%   100%  1417.21kB 35.28%  runtime/pprof.(*profileBuilder).locForPC
         0     0%   100%   902.59kB 22.47%  compress/gzip.(*Writer).Write
         0     0%   100%  2600.21kB 64.72%  github.com/apache/pulsar-client-go/pulsar.newPartitionProducer
         0     0%   100%  2600.21kB 64.72%  github.com/apache/pulsar-client-go/pulsar.newProducer.func2
         0     0%   100%  1417.21kB 35.28%  main.main
         0     0%   100%  1417.21kB 35.28%  runtime.main
         0     0%   100%   902.59kB 22.47%  runtime/pprof.(*profileBuilder).flush
         0     0%   100%  1417.21kB 35.28%  runtime/pprof.WriteHeapProfile

```

after 100 iterations of create and close producer with different topic, a possible memory leak was found (`github.com/apache/pulsar-client-go/pulsar/internal.NewBlockingQueue`). Which means `partitionProducer`s are not been garbage collected. After code reviews, we found that `ClientHandlers` only cleanup the cached `producer` and `consumer` after `client.Close()` been called. Which caused the closed `producer` still been referenced by `ClientHandlers` and not been garbage collected. 

#### Steps to reproduce

```

	client, err := pulsar.NewClient(pulsar.ClientOptions{
		URL: "pulsar://127.0.0.1:16650",
	})

	defer client.Close()

	if err != nil {
		log.Fatal(err)
		return
[...]

PR #154, opened 2019-12-31
[Issue #153] Fix handler memory leak

freeznet:
Fixes: #153 

Makes producer & consumer removed from client handler when producer / consumer closed.

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -41,6 +41,7 @@ type acker interface {
 }
 
 type consumer struct {
+	client    *client
 	options   ConsumerOptions
 	consumers []*partitionConsumer
 
@@ -114,6 +115,7 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 func internalTopicSubscribe(client *client, options ConsumerOptions, topic string,
 	messageCh chan ConsumerMessage) (*consumer, error) {
 	consumer := &consumer{
+		client:    client,
 		options:   options,
 		messageCh: messageCh,
 		closeCh:   make(chan struct{}),
@@ -317,6 +319,7 @@ func (c *consumer) Close() {
 		wg.Wait()
 		close(c.closeCh)
 	})
+	c.client.handlers.Del(c)
 }
 
 var r = &random{
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -318,8 +318,8 @@ func (c *consumer) Close() {
 		}
 		wg.Wait()
 		close(c.closeCh)
+		c.client.handlers.Del(c)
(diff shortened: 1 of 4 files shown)


cckellogg:
Can you add a test for this?

pulsar/internal/client_handlers.go
```diff
@@ -36,6 +36,13 @@ func (h *ClientHandlers) Add(c Closable) {
 	defer h.l.Unlock()
 	h.handlers[c] = true
 }
+
+func (h *ClientHandlers) Del(c Closable) {
+	h.l.Lock()
+	defer h.l.Unlock()
+	delete(h.handlers, c)
+}
+
```
cckellogg: The close method in this class will have to change otherwise there is a deadlock.

We should make a copy of the handlers then release the lock and then call close on the list of handlers. Otherwise the one handler (producer or consumer) will try to call Del() and be deadlocked.

Something like this should work

func (h *ClientHandlers) Close() {
        h.l.Lock()
	handlers := make([]Closable, 0, len(h.handlers))
	for handler := range h.handlers {
		handlers = append(handlers, handler)
	}
	h.l.Unlock()

	for _, handler := range handlers {
		handler.Close()
	}
}

jiazhai:
@freeznet Would you please help handle @cckellogg 's comments? Thanks

freeznet:
@cckellogg deadlock has been resolved as you mentioned, also add test for `Del`. PTAL again, thanks.

pulsar/consumer_impl.go
```diff
@@ -317,6 +319,7 @@ func (c *consumer) Close() {
 		wg.Wait()
 		close(c.closeCh)
 	})
+	c.client.handlers.Del(c)
```
cckellogg: Is there a reason not to call this in the closeOnce?

freeznet: the initial thought was `delete` operation will be no-op if consumer/producer not in handlers, so i didnt put `Del` call in the `closeOnce`. But your idea is better, which will prevent non necessary locks&unlocks.

freeznet:
> It seems like the tests did not run on the last commit?

@cckellogg the test runs ok on my laptop, can you provide any detail? thanks.

```
=== RUN   TestClientHandlers_Del
--- PASS: TestClientHandlers_Del (0.00s)
    client_handlers_test.go:61: closable1 is: closed  true
    client_handlers_test.go:62: closable2 is: closed  true
```

cckellogg:
What happens when you start pulsar in standalone and run
```
go test -race ./...
```
Is there a way we can trigger the unit/integration tests? It would be nice to have those run to ensure the issue has been fixed.

merlimat:
retest this please

Issue #152, opened 2019-12-30
Support receive timeout


xujianhai666:
**Is your feature request related to a problem? Please describe.**
be consistent with java consumer impl 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

xujianhai666:
working on this now.

wolfstudy:
> working on this now.

Cool, thanks for your contribution.

cckellogg:
Doesn't the consumer already support a receive timeout? 

https://golang.org/pkg/context/#WithTimeout

You can create a context with a timeout and if that expires the receive will return.

xujianhai666:
> Doesn't the consumer already support a receive timeout?
> 
> https://golang.org/pkg/context/#WithTimeout
> 
> You can create a context with a timeout and if that expires the receive will return.

@cckellogg yes, I impl with that.

cckellogg:
@xujianhai666 Why do you need to add another method if you can already do it with current method Receive function?

PR #156, opened 2020-01-01
[ISSUE #152][consumer] add consumer timeout

xujianhai666:
### Motivation
- *support receiveMessage with timeout*

### Modifications
- *add `ReceiveBlock(context.Context, time.Duration) (Message, error)` for consumer interface* 
- *impl ReceiveBlock for consumer、regexConsumer、multiTopicConsumer*

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -135,6 +135,10 @@ type Consumer interface {
 	// This calls blocks until a message is available.
 	Receive(context.Context) (Message, error)
 
+	// Receive a single message.
+	// This calls blocks until a message is available or timeout.
+	ReceiveBlock(context.Context, time.Duration) (Message, error)
+
 	// Chan returns a channel to consume messages from
 	Chan() <-chan ConsumerMessage
 
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -240,6 +240,12 @@ func (c *consumer) Receive(ctx context.Context) (message Message, err error) {
 	}
 }
 
+func (c *consumer) ReceiveBlock(ctx context.Context, timeout time.Duration) (message Message, err error) {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	return c.Receive(ctx)
+}
+
 // Messages
 func (c *consumer) Chan() <-chan ConsumerMessage {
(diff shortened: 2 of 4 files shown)


pulsar/consumer.go
```diff
@@ -135,6 +135,10 @@ type Consumer interface {
 	// This calls blocks until a message is available.
 	Receive(context.Context) (Message, error)
 
+	// Receive a single message.
+	// This calls blocks until a message is available or timeout.
+	ReceiveBlock(context.Context, time.Duration) (Message, error)
```
cckellogg: I don't think we should add this api. The user can wrap a context with a timeout in the Receive method. In my opinion this complicates the api and does not add anything functionality.

wolfstudy: > The user can wrap a context with a timeout in the Receive method.

@cckellogg Agree with you, In here, we can use `context.WithTimeout()`, as you provide:

```
ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
defer cancel()
```

But I think a more user-friendly way is as follows:

- add `Timeout` filed in `ConsumerOptions` struct
- provide a new receive method as @xujianhai666 did

Anyway, these forms are ok with me.

cckellogg: @wolfstudy I disagree that is more user friendly.

The purpose of having a Context with the current Receive method is to allow users the ability to add a timeout or cancel to a receive call if they want to. Adding a new receive method that does something that can already be done through the current api is not good for the user. It confusing and complicates the api. We should try to keep the api simple because once you publicly expose something you have to always support it. We should add documentation on how to add a timeout to a receive as opposed to adding a new api.

A user can add a few lines to their program and get a timeout with every receive. For example.

```
func receiveWithTimeout(receiveFn func(ctx context.Context) (Message, error), duration time.Duration) (Message, error) {
	ctx, cancel := context.WithTimeout(context.Background(), duration)
	defer cancel()
	return receiveFn(ctx)
}

m, err := receiveWithTimeout(c.Receive, 500*time.Millisecond)
```

We should give example like this instead of adding an additional api. Thoughts?

xujianhai666: @cckellogg I agree with u, And under wolfstudy  idea, I thought the ConsumerOptions is better, Do u agree with that?

PR #158, opened 2020-01-02
Topic reader implementation

merlimat:
### Motivation

Added Reader implemenation

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -177,6 +177,33 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
 	pc.conn.DeleteConsumeHandler(pc.consumerID)
 }
 
+func (pc *partitionConsumer) getLastMessageID() (*messageID, error) {
+	req := &getLastMsgIdRequest{doneCh: make(chan struct{})}
+	pc.eventsCh <- req
+
+	// wait for the request to complete
+	<-req.doneCh
+	return req.msgId, req.err
+}
+
+func (pc *partitionConsumer) internalGetLastMessageID(req *getLastMsgIdRequest) {
+	defer close(req.doneCh)
+
+	requestID := pc.client.rpcClient.NewRequestID()
+	cmdGetLastMessageID := &pb.CommandGetLastMessageId{
+		RequestId:  proto.Uint64(requestID),
+		ConsumerId: proto.Uint64(pc.consumerID),
+	}
+	res, err := pc.client.rpcClient.RequestOnCnx(pc.conn, requestID, pb.BaseCommand_GET_LAST_MESSAGE_ID, cmdGetLastMessageID)
+	if err != nil {
+		pc.log.WithError(err).Error("Failed to get last message id")
+		req.err = err
+	} else {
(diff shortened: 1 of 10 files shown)


pulsar/consumer_partition.go
```diff
@@ -516,15 +577,27 @@ func (pc *partitionConsumer) grabConn() error {
...
+		PriorityLevel:              nil,
+		Durable:                    proto.Bool(pc.options.subscriptionMode == durable),
+		Metadata:                   internal.ConvertFromStringMap(pc.options.metadata),
+		ReadCompacted:              proto.Bool(pc.options.readCompacted),
+		Schema:                     nil,
+		InitialPosition:            initialPosition.Enum(),
+		ReplicateSubscriptionState: proto.Bool(pc.options.replicateSubscriptionState),
 	}
+
+	pc.startMessageID = pc.clearReceiverQueue()
```
cckellogg: Should we ensure that the receiver queue is only ever cleared from a single go routine? This was the reason it was done in the dispatcher. There could be a small random race condition here because the dispatcher go routine and the eventsLoop go routine will both be reading from the queueCh so when the nil message is sent there is a possibility the dispatcher will read it and then grabConn with never return.

pulsar/consumer_partition.go
```diff
@@ -247,7 +294,7 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 	if msgMeta.NumMessagesInBatch != nil {
 		numMsgs = int(msgMeta.GetNumMessagesInBatch())
 	}
-	messages := make([]*message, numMsgs)
+	messages := make([]*message, 0)
```
cckellogg: we can do this `make([]*message, 0, numMsgs)` then append will work and we don't ever have to reallocate the slice.

merlimat: The problem here is that we are doing the filtering, so the actual number of messages might be lower than that

pulsar/consumer_partition.go
merlimat: The `clearReceiverQueue()` is only called on reconnection from the consumer routine, so I don't think there should be any race condition here.

merlimat: @cckellogg You're right, there's a race condition there. It should be fixed with the latest commit. PTAL again.

pulsar/consumer_partition.go
```diff
@@ -109,9 +129,11 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 		eventsCh:       make(chan interface{}, 3),
 		queueSize:      int32(options.receiverQueueSize),
 		queueCh:        make(chan []*message, options.receiverQueueSize),
+		startMessageID: options.startMessageID,
 		connectedCh:    make(chan struct{}),
 		messageCh:      messageCh,
 		closeCh:        make(chan struct{}),
+		clearQueueCh:   make(chan func(id *messageID)),
```
cckellogg: Are we ok with this channel being unbuffered?

merlimat: I think it's ok since the dispatcher routine is not blocked on writing on the application channel

PR #162, opened 2020-01-04
Avoid double cnx close notification on graceful server side close

merlimat:
### Motivation

We're processing the "connection closed" event twice when the broker sends a "producer closed" or "consumer closed" command. 

Basically after the producer reconnects, it will reconnect again on the same connection because it got the notification 2 times. One when the graceful close command and another when the connection is closed (eg: broker is shutting down).

This is not harmful, though it's logically wrong.

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -550,16 +550,19 @@ func (c *connection) handleCloseConsumer(closeConsumer *pb.CommandCloseConsumer)
 	consumerID := closeConsumer.GetConsumerId()
 	if consumer, ok := c.consumerHandler(consumerID); ok {
 		consumer.ConnectionClosed()
+		delete(c.listeners, consumerID)
 	} else {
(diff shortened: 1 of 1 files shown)


pulsar/internal/connection.go
```diff
@@ -550,16 +550,19 @@ func (c *connection) handleCloseConsumer(closeConsumer *pb.CommandCloseConsumer)
 	consumerID := closeConsumer.GetConsumerId()
 	if consumer, ok := c.consumerHandler(consumerID); ok {
 		consumer.ConnectionClosed()
+		delete(c.listeners, consumerID)
```
cckellogg: Should we use this function DeleteConsumeHandler to delete the handler? We get the handler using the consumerHandlersLock so to be consistent should we delete with it too?

cckellogg: Seems like this is a different map so it ok.

Issue #151, opened 2019-12-30
consumer.Subscribe()  blocked  after the  "Short read when reading frame size"  error happend


dsmlily:
##### 1、code

```
func main(){
	client, err := pulsar.NewClient(pulsar.ClientOptions{
		URL: "pulsar://10.32.32.20:6650",
		ConnectionTimeout:time.Second * time.Duration(2),
		OperationTimeout: time.Second * time.Duration(2),
	})

	if err != nil {
		log.Fatal(err)
	}

	GConsumerClient = client
	wg := sync.WaitGroup{}

	defer client.Close()
	for i := 0; i < 20; i++ {
		index := i
		time.Sleep(time.Second*2)
		roomId := strconv.Itoa(index)
		subscriptionName := fmt.Sprintf("sub-name-%s", roomId)
		c, e := initRoomPulsarConsumer(roomId,subscriptionName)
		if e != nil {
			fmt.Printf("create faild  -- %d\n", index)
			//wg.Done()
			//log.Fatal(e)
			continue
		}
		time.Sleep(time.Second*1)
		wg.Add(1)
		go func(con pulsar.Consumer,sb string) {
			fmt.Printf("to close room %s.....\n",sb)
			err := con.Unsubscribe()
			if err != nil {
				fmt.Printf("******* Unsubscribe error  %s.....\n",err.Error())
			}
			con.Close()
			fmt.Printf("to close room [%s close done***]\n",sb)
			wg.Done()
		}(c,subscriptionName)
		fmt.Printf("wait group -- %d done\n", index)	
	}
	wg.Wait()
}

func initRoomPulsarConsumer(roomId string, subscriptionName string) (consumer pulsar.Consumer, err error) {
	fmt.Printf("to creat room consumer SubscriptionName:%s .....\n",subscriptionName)
	cfg := pulsar.ConsumerOptions{
		Topic:               fmt.Sprintf("dev_pid2/signal_geo/%s", roomId),
		SubscriptionName:    subscriptionName,
		Type:                pulsar.Failover,
	}
   
	consumer, err = GConsumerClient.Subscribe(cfg)
	if err != nil {
		fmt.Printf("*****   could not establish subscription %v, err: %s", roomId, err.Error())
		return nil, err
	}
	fmt.Printf("creat room consumer done, SubscriptionName:%s  ***creat done\n",cfg.SubscriptionName)
	return
}
```
##### 2、Screenshot
[...]

Issue #144, opened 2019-12-26
For a pulsar.Client instance, Once Client.Subscribe() times out, all subsequent Client.Subscribe() fail


dsmlily:
#### Expected behavior

When the network is restored, the subscription is successfully created

#### Actual behavior

When the network is restored, creating a subscription fails

##### 1、code file  test.go

```
func main(){
	client, err := pulsar.NewClient(pulsar.ClientOptions{
		URL: "pulsar://10.32.32.20:6650",
		ConnectionTimeout:time.Second * time.Duration(2),
		OperationTimeout: time.Second * time.Duration(2),
	})
	if err != nil {
		log.Fatal(err)
	}

	GConsumerClient = client

	defer client.Close()
	for i := 0; i < 10; i++ {
		index := i
		time.Sleep(time.Second*2)
		roomId := strconv.Itoa(index)
		subscriptionName := fmt.Sprintf("sub-name-%s", roomId)
		c, e := initRoomPulsarConsumer(roomId,subscriptionName)
		if e != nil {
			fmt.Printf("create faild  -- %d\n", index)
			//wg.Done()
			//log.Fatal(e)
			continue
		}
		time.Sleep(time.Second*1)
		go func(con pulsar.Consumer,sb string) {
			fmt.Printf("to close room %s.....\n",sb)
			err := con.Unsubscribe()
			if err != nil {
				fmt.Printf("******* Unsubscribe error  %s.....\n",err.Error())
			}
			con.Close()
			fmt.Printf("to close room [%s close done***]\n",sb)
		}(c,subscriptionName)
		fmt.Printf("wait group -- %d done\n", index)	
	}
}
```

```
func initRoomPulsarConsumer(roomId string, subscriptionName string) (consumer pulsar.Consumer, err error) {
	fmt.Printf("to creat room consumer SubscriptionName:%s .....\n",subscriptionName)
	cfg := pulsar.ConsumerOptions{
		Topic:               fmt.Sprintf("dev_pid2/signal_geo/%s", roomId),
		SubscriptionName:    subscriptionName,
		Type:                pulsar.Failover,
	}
   
	consumer, err = GConsumerClient.Subscribe(cfg)
	if err != nil {
		fmt.Printf("*****   could not establish subscription %v, err: %s", roomId, err.Error())
		return nil, err
	}
	fmt.Printf("creat room consumer done, SubscriptionName:%s  ***creat done\n",cfg.SubscriptionName)
	return
}
```

##### 2、steps

[...]

PR #167, opened 2020-01-08
res channel causes connection.Run() to block When request timeout occuers

dsmlily:
Change the RPCResultChannel to a buffer channel to avoid blocking connection.Run() when request timeout occuers

https://github.com/apache/pulsar-client-go/issues/144
https://github.com/apache/pulsar-client-go/issues/151

Signed-off-by: dsmlily <dsmlily_2008@126.com>

### DIFF
--- a/pulsar/internal/rpc_client.go
+++ b/pulsar/internal/rpc_client.go
@@ -90,7 +90,7 @@ func (c *rpcClient) Request(logicalAddr *url.URL, physicalAddr *url.URL, request
 		*RPCResult
 		error
 	}
-	ch := make(chan Res, 1)
+	ch := make(chan Res, 10)
 
 	// TODO: in here, the error of callback always nil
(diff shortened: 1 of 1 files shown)


pulsar/internal/rpc_client.go
```diff
@@ -90,7 +90,7 @@ func (c *rpcClient) Request(logicalAddr *url.URL, physicalAddr *url.URL, request
 		*RPCResult
 		error
 	}
-	ch := make(chan Res)
+	ch := make(chan Res, 1)
```
wolfstudy:
```suggestion
	ch := make(chan Res, 10)
```

Normally we use 10 as the buffer size.

wolfstudy:
@cckellogg @merlimat Why the Jenkins is not triggered? The only committer can trigger?

wolfstudy:
retest this please

Issue #148, opened 2019-12-27
Add seek by msgID on consumer


wolfstudy:
**Describe the solution you'd like**

Add seek by msgID on the consumer, the interface as follows:

```
Seek (msgID MessageID) error
```

wolfstudy:
The feature has been implemented in #168

PR #168, opened 2020-01-08
[Issue-148][pulsar-client-go] add seek by messageID

stevenwangnarvar:
Master Issue: [#148](https://github.com/apache/pulsar-client-go/issues/148)

### Motivation

*Add seek by messageID*

### Modifications

*Add seek by messageID*
*Refactor some UT code*

### Verifying this change

*(example:)*
  - *Added integration tests for seek by messageID*

  - The public API: (*yes* / no)

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -303,6 +303,10 @@ func (c *consumer) Close() {
 }
 
 func (c *consumer) Seek(msgID MessageID) error {
+	if len(c.consumers) > 1 {
+		return errors.New("for partition topic, seek command should perform on the individual partitions")
+	}
+
 	mid, ok := c.messageID(msgID)
 	if !ok {
 		return nil
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -256,9 +256,8 @@ func (c *consumer) Ack(msg Message) {
 
 // Ack the consumption of a single message, identified by its MessageID
 func (c *consumer) AckID(msgID MessageID) {
-	mid, ok := msgID.(*messageID)
+	mid, ok := c.messageID(msgID)
 	if !ok {
-		c.log.Warnf("invalid message id type")
 		return
 	}
 
@@ -267,14 +266,7 @@ func (c *consumer) AckID(msgID MessageID) {
 		return
 	}
 
-	partition := mid.partitionIdx
-	// did we receive a valid partition index?
-	if partition < 0 || partition >= len(c.consumers) {
-		c.log.Warnf("invalid partition index %d expected a partition between [0-%d]",
-			partition, len(c.consumers))
-		return
-	}
-	c.consumers[partition].AckID(mid)
+	c.consumers[mid.partitionIdx].AckID(mid)
 }
 
 func (c *consumer) Nack(msg Message) {
@@ -282,9 +274,8 @@ func (c *consumer) Nack(msg Message) {
 }
 
 func (c *consumer) NackID(msgID MessageID) {
-	mid, ok := msgID.(*messageID)
+	mid, ok := c.messageID(msgID)
 	if !ok {
-		c.log.Warnf("invalid message id type")
 		return
 	}
 
@@ -293,15 +284,7 @@ func (c *consumer) NackID(msgID MessageID) {
 		return
 	}
 
-	partition := mid.partitionIdx
-	// did we receive a valid partition index?
-	if partition < 0 || partition >= len(c.consumers) {
-		c.log.Warnf("invalid partition index %d expected a partition between [0-%d]",
-			partition, len(c.consumers))
-		return
-	}
-
-	c.consumers[partition].NackID(mid)
+	c.consumers[mid.partitionIdx].NackID(mid)
 }
 
 func (c *consumer) Close() {
@@ -317,6 +300,15 @@ func (c *consumer) Close() {
 		wg.Wait()
 		close(c.closeCh)
 	})
+}
+
+func (c *consumer) Seek(msgID MessageID) error {
+	mid, ok := c.messageID(msgID)
+	if !ok {
+		return nil
+	}
+
+	return c.consumers[mid.partitionIdx].Seek(mid)
 }
 
 var r = &random{
@@ -363,4 +355,22 @@ func toProtoInitialPosition(p SubscriptionInitialPosition) pb.CommandSubscribe_I
 	}
 
 	return pb.CommandSubscribe_Latest
+}
+
+func (c *consumer) messageID(msgID MessageID) (*messageID, bool) {
+	mid, ok := msgID.(*messageID)
+	if !ok {
+		c.log.Warnf("invalid message id type")
+		return nil, false
+	}
+
+	partition := mid.partitionIdx
+	// did we receive a valid partition index?
+	if partition < 0 || partition >= len(c.consumers) {
+		c.log.Warnf("invalid partition index %d expected a partition between [0-%d]",
+			partition, len(c.consumers))
+		return nil, false
+	}
+
+	return mid, true
 }
--- a/pulsar/consumer_multitopic.go
+++ b/pulsar/consumer_multitopic.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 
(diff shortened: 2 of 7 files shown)


wolfstudy:
@stevenwangnarvar Is it possible to split this pull request?

- One of them is to implement seek by message ID
- One of them is to refactoring the `consumer_test.go`

pulsar/consumer_impl.go
```diff
@@ -319,6 +302,15 @@ func (c *consumer) Close() {
 	})
 }
 
+func (c *consumer) Seek(msgID MessageID) error {
+	mid, ok := c.messageID(msgID)
+	if !ok {
+		return nil
```
cckellogg: Shouldn't we return an error here?

pulsar/consumer_partition.go
```diff
@@ -256,6 +256,50 @@ func (pc *partitionConsumer) Close() {
...
+		RequestId:  proto.Uint64(requestID),
+		MessageId:  id,
+	}
+	_, err = pc.client.rpcClient.RequestOnCnx(pc.conn, requestID, pb.BaseCommand_SEEK, cmdSeek)
+	if err != nil {
+		pc.log.WithError(err).Error("Failed to reset to message id")
+		seek.err = err
+	}
+
+	pc.conn.DeleteConsumeHandler(pc.consumerID)
```
cckellogg: Why are we deleting the consumer handler here? That does not seem correct.

pulsar/test_helper.go
```diff
@@ -150,6 +153,12 @@ func deleteTopic(topic string) error {
 	return httpDelete("admin/v2/persistent/" + fmt.Sprintf("%s?force=true", topic))
 }
 
+func createPartitionTopic(t *testing.T, topicName string, partitions int) {
```
cckellogg: No need to pass the testing.T here. Let's just return an error and handle the assert in the test.

pulsar/consumer_test.go
```diff
@@ -663,128 +539,144 @@ func TestConsumerNack(t *testing.T) {
...
+		if i == 4 {
+			seekID = msg.ID()
+		}
+	})
+
+	// seek to fourth message, expected receive fourth message.
+	err := consumer.Seek(seekID)
+	assert.Nil(t, err)
+
+	time.Sleep(500 * time.Millisecond)
```
cckellogg: Do we need a sleep here? It's best to avoid sleeps in test if possible.

pulsar/consumer_impl.go
```diff
@@ -364,3 +356,21 @@ func toProtoInitialPosition(p SubscriptionInitialPosition) pb.CommandSubscribe_I
 
 	return pb.CommandSubscribe_Latest
 }
+
+func (c *consumer) messageID(msgID MessageID) (*messageID, bool) {
```
cckellogg: maybe call this validateMessageID?

pulsar/consumer_test.go
```diff
@@ -202,192 +157,160 @@ func TestConsumerWithInvalidConf(t *testing.T) {
...
-			receivedConsumer1++
-			consumer1.Ack(cm.Message)
-		case cm, ok := <-consumer2.Chan():
-			if !ok {
-				break
-			}
-			receivedConsumer2++
-			consumer2.Ack(cm.Message)
+	previousMsg, previousKey := "", ""
+	receive(t, consumer1, ctx, 3, func(t *testing.T, msg Message, i int) {
```
cckellogg: This seems way more complex than the previous test. The previous test to me is more easy to reason about.

pulsar/test_helper.go
```diff
@@ -164,3 +173,55 @@ func topicPath(topic string) string {
...
+
+func createProducer(t *testing.T, client Client, options ProducerOptions) Producer {
+	producer, err := client.CreateProducer(options)
+	assert.Nil(t, err)
+	assert.Equal(t, options.Topic, producer.Topic())
+
+	return producer
+}
+
+func send(t *testing.T, producer Producer, ctx context.Context, times int, message *ProducerMessage) {
```
cckellogg: Check the consumer_regex_test genMessages https://github.com/apache/pulsar-client-go/blob/master/pulsar/consumer_regex_test.go#L361 this might be a way to make it more flexible.

pulsar/consumer_test.go
```diff
@@ -663,128 +539,144 @@ func TestConsumerNack(t *testing.T) {
...
+
+	times := 10
+	msgPayload := "hello"
+
+	ctx := context.Background()
+	send(t, producer, ctx, times, &ProducerMessage{
+		Payload: []byte(msgPayload),
+	})
+
+	var seekID MessageID
```
cckellogg: The producer returns a messageID after send can we just keep track of that and then call seek Instead of receiving messages to the seek point?

pulsar/consumer_test.go
wangjia007bond: My purpose is to reduce duplicate code and unify test cases.

pulsar/consumer_impl.go
wangjia007bond: I checked other interfaces like Ack and NackID, they didn't handle error, because their interface didn't need return error, but I am not sure the correct behavior here. so I follow the current way.

pulsar/consumer_partition.go
wangjia007bond: sure, removed.

pulsar/test_helper.go
wangjia007bond: Yes, will do the ut refactor in the future, this pr will remove those.

pulsar/consumer_test.go
wangjia007bond: Removed.

pulsar/consumer_impl.go
wangjia007bond: I though this before, but this function will convert MessageID(uppercase M) to messageID(lowercase m), so any function name like XXXMessageID, will cause confusion...

pulsar/test_helper.go
wangjia007bond: OK.

pulsar/consumer_test.go
wangjia007bond: Got it. Done.

wolfstudy:
retest this please

wangjia007bond:
> The check of `golanglint-ci` fail, please fix them.
> 
> you can run `golangci-lint run -c ./golangci.yml ./...` in locally.
> 
> ```
> pulsar/test_helper.go:201:1: context.Context should be the first parameter of a function (golint)
> func send(t *testing.T, producer Producer, ctx context.Context, times int, message *ProducerMessage) {
> ^
> pulsar/test_helper.go:216:1: context.Context should be the first parameter of a function (golint)
> func receive(t *testing.T, consumer Consumer, ctx context.Context, times int, inspect func(*testing.T, Message, int)) {
> ^
> pulsar/test_helper.go:25: File is not `goimports`-ed (goimports)
> 	"github.com/stretchr/testify/assert"
> pulsar/consumer_test.go:56: File is not `goimports`-ed (goimports)
> 	msgProperties := map[string]string{ "key-1": "pulsar-1" }
> ```

Done.

pulsar/consumer_multitopic.go
```diff
@@ -160,3 +160,14 @@ func (c *multiTopicConsumer) Close() {
 		close(c.closeCh)
 	})
 }
+
+func (c *multiTopicConsumer) Seek(msgID MessageID) error {
```
cckellogg: We should return an error here and not allow Seek on multi topic consumer since it does not make sense. Same for the regex consumer.

pulsar/consumer_impl.go
```diff
@@ -319,6 +302,15 @@ func (c *consumer) Close() {
 	})
 }
 
+func (c *consumer) Seek(msgID MessageID) error {
```
cckellogg: If the consumer is partitioned we should return an error

pulsar/consumer_multitopic.go
wolfstudy: @wangjia007bond  agree with @cckellogg .

Note: this operation can only be done on non-partitioned topics. For these, one can rather perform the seek() on the individual partitions.

stevenwangnarvar: yes, make sense, this particularly msg id actually can't applied to all the topics for multtopic consumer or regex consumer.

stevenwangnarvar: Done.

pulsar/consumer_impl.go
wangjia007bond: How can I know this consumer is partitioned?

cckellogg: If there is more than 1 consumer in the list then it's a partitioned topic. If topic is partitioned and has only 1 partition or the topic is not partitioned then there will be 1 consumer which is kind of the same thing.

stevenwangnarvar: Done.

PR #175, opened 2020-01-16
Fix partition key not set in batch builder

wolfstudy:

### Motivation

Currently, when we specify the `Key` for the message, the broker can not receive the partition key. 

image

### Modifications

- set partition key in BatchBuilder

### DIFF
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -280,7 +280,7 @@ func TestConsumerKeyShared(t *testing.T) {
 	defer producer.Close()
 
 	ctx := context.Background()
-	for i := 0; i < 10; i++ {
+	for i := 0; i < 100; i++ {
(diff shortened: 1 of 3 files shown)


wolfstudy:
@sijie @merlimat @jiazhai PTAL thanks

pulsar/consumer_test.go
```diff
@@ -280,7 +280,7 @@ func TestConsumerKeyShared(t *testing.T) {
 	defer producer.Close()
 
 	ctx := context.Background()
-	for i := 0; i < 10; i++ {
+	for i := 0; i < 100; i++ {
```
cckellogg: What's the purpose of changing this from 10 -> 100?

wolfstudy: Just to ensure that every consumer receives a message

PR #176, opened 2020-01-17
PIP-55: Refresh auth credentials

merlimat:
### Motivation

Companion PR of https://github.com/apache/pulsar/pull/6074. 

It's currently difficult to test this feature in the Go integration tests until we have a released version of Pulsar with this feature. In the meantime, it doesn't cause any incompatibility.


pulsar/internal/connection.go
```diff
@@ -36,6 +36,14 @@ import (
 	"github.com/apache/pulsar-client-go/pulsar/internal/pb"
 )
 
+const (
+	// TODO: Find a better way to embed the version in the library code
```
wolfstudy: Maybe we can create a `version` file and read the version value from the file

merlimat: Problem is that we're shipping this as a library, so I guess the version file is not going to be bundled. Anyway, as a start we could go by with remembering to update the version string in source code, until we find a better solution.

PR #184, opened 2020-02-13
Dead letter queue implementation

merlimat:
### Motivation

Added support for dead letter queue message routing


pulsar/dlq_router.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			r.log.Debug("Closed DLQ router")
+			return
+		}
+	}
+}
+
+func (r *dlqRouter) close() {
+	if r.closeCh != nil {
+		r.closeCh <- nil
```
cckellogg: I think this could cause a deadlock for anything but a single topic because there will be multiple calls to close() and only the first call will be able to send on the closeCh because the go routine reading from the closeCh will exit.

We could do something like this:
```
select {
 case r.closeCh <- nil: // this case will only occur when we can send on the channel
default:
// the closeCh is either nil or it's blocked and we can't send on it
}
```

pulsar/consumer_impl.go
```diff
@@ -74,6 +75,11 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		messageCh = make(chan ConsumerMessage, 10)
 	}
 
+	dlq, err := newDlqRouter(client, options.DLQ)
```
cckellogg: what are the defaults here? will every topic by default have dlq enabled?

pulsar/consumer_partition.go
```diff
@@ -402,23 +404,27 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
...
-				msgID:       msgID,
-				payLoad:     payload,
+				publishTime:         timeFromUnixTimestampMillis(msgMeta.GetPublishTime()),
+				eventTime:           timeFromUnixTimestampMillis(smm.GetEventTime()),
+				key:                 smm.GetPartitionKey(),
+				properties:          internal.ConvertToStringMap(smm.GetProperties()),
+				topic:               pc.topic,
+				msgID:               msgID,
+				payLoad:             payload,
+				replicationClusters: msgMeta.GetReplicateTo(),
```
jerrypeng: why is this added?

pulsar/dlq_router.go
merlimat: Added select and test case with multi topics

pulsar/consumer_impl.go
merlimat: By default the policy is `nil` and the it will result in no-op

pulsar/consumer_partition.go
merlimat: We need to copy all properties of the original message and noticed that this was not being exposed in the internal struct.

PR #186, opened 2020-02-13
Expose EventTime consistently as a non-pointer

merlimat:
### Motivation

When producing a message, we are currently accepting event time as a pointer:

```go
type ProducerMessage struct {
 /// .... 
    // EventTime set the event time for a given message
    EventTime *time.Time
}
```

though on a receiving message, we're exposing as a value: 

```go
type Message interface {
	/// ...
	// PublishTime get the publish time of this message. The publish time is the timestamp that a client
	// publish the message.
	PublishTime() time.Time

	// EventTime get the event time associated with this message. It is typically set by the applications via
	// `ProducerMessage.EventTime`.
	// If there isn't any event time associated with this event, it will be nil.
	EventTime() time.Time
```

We should unify these 2 and also with the same approach used for `PublishTime`.

### DIFF
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -503,7 +503,7 @@ func TestConsumerEventTime(t *testing.T) {
 	et := timeFromUnixTimestampMillis(uint64(5))
 	_, err = producer.Send(ctx, &ProducerMessage{
 		Payload:   []byte("test"),
-		EventTime: &et,
+		EventTime: et,
 	})
 	assert.Nil(t, err)
 
--- a/pulsar/dlq_router.go
+++ b/pulsar/dlq_router.go
@@ -92,12 +92,11 @@ func (r *dlqRouter) run() {
 
 			msg := cm.Message.(*message)
 			msgID := msg.ID()
-			eventTime := msg.EventTime()
 			producer.SendAsync(context.Background(), &ProducerMessage{
 				Payload:             msg.Payload(),
 				Key:                 msg.Key(),
 				Properties:          msg.Properties(),
-				EventTime:           &eventTime,
+				EventTime:           msg.EventTime(),
 				ReplicationClusters: msg.replicationClusters,
 			}, func(MessageID, *ProducerMessage, error) {
 				r.log.WithField("msgID", msgID).Debug("Sent message to DLQ")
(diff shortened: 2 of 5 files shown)


merlimat:
> I thought pointer can be used for presenting the value is not set. We are changing the behavior to treat 0 as not set. 

In many places in Go, I've seen passed a `time.Time` or `time.Duration` by value with default value (0) representing unset. One example: https://godoc.org/cloud.google.com/go/pubsub#PublishSettings

> I think this is not consistent across different language implementations. I'd prefer a semantically consistent behavior between different language implementations rather than consistency within a language client.

We're using `long` everywhere else:  
 * Java https://pulsar.apache.org/api/client/2.5.0-SNAPSHOT/org/apache/pulsar/client/api/Message.html#getEventTime--
 * C++ https://pulsar.apache.org/api/cpp/2.5.0-SNAPSHOT/classpulsar_1_1_message.html#a0e87338a9b4ebd3b41d53826664f900c

Though, at this micro-api level, I'd rather go for language level consistency over cross-language consistency, unlike other concepts (Client, Producer, Consumer, etc..)

In Go, `time.Time` is how you define a timestamp and 0 is not a "valid" value.

In any case, we should, at the very least, be consistent across `ProducerMessage` and `Message` definitions since they represent the same optional value.

sijie:
I think my concern is how we are going to tell if an `event_time` is not set vs the `event_time` is set to 0. I have seen users were frustrated by some inconsistent behaviors between different language clients. Should we at least try to maintain a better consistent behavior across the languages maintained by the ASF. 

> at the very least, be consistent across ProducerMessage and Message definitions

I understand the reason to make ProducerMessage and Message consistent. So not make Message return EventTime as pointer and using Nil for presenting EventTime not set?

merlimat:
> I understand the reason to make ProducerMessage and Message consistent. So not make Message return EventTime as pointer and using Nil for presenting EventTime not set?

I was leaning toward using by-value because: 
 * 0 is not "valid" for `time.Time` since it's Jan 1, 1970 and it's seems to be accepted like that in Go
 * We're using eventTime=0 to mean event time is not set in Java and C++ already

Either way works and if there's more consensus on using pointer I'll switch to that.

sijie:
I am fine with using value == 0. Although I'd like to see it documented in godoc as how we document it in java client.

merlimat:
> I am fine with using value == 0. Although I'd like to see it documented in godoc as how we document it in java client.

Yes, I tried to expand and clarify a bit more. PTAL

sijie:
Thank you.

PR #197, opened 2020-03-17
Add AckTimeout logic

wolfstudy:

### Modifications

- Add AckTimeout logic


wolfstudy:
> The ack timeout was intentionally left out from the Go client.
> 
> Negative ack are the preferred way to solve that problem and ack timeout should be regarded as legacy.

Yes, agree with you, will close the change.

PR #198, opened 2020-03-18
Fix pprof typo

wolfstudy:

### Modifications

- Fix pprof typo

### DIFF
--- a/perf/pulsar-perf-go.go
+++ b/perf/pulsar-perf-go.go
@@ -118,8 +118,8 @@ func serveProfiling(addr string, stop <-chan struct{}) error {
 	}()
 
 	fmt.Printf("Starting pprof server at: %s\n", addr)
-	fmt.Printf("  use go tool pprof http://%s/debug/pprof/prof to get pprof file(cpu info)\n", addr)
-	fmt.Printf("  use go tool pprof http://%s/debug/pprof/heap to get inuse_space file\n", addr)
+	fmt.Printf("  use `go tool pprof http://%s/debug/pprof/profile` to get pprof file(cpu info)\n", addr)
+	fmt.Printf("  use `go tool pprof http://%s/debug/pprof/heap` to get inuse_space file\n", addr)
 	fmt.Println()
 
 	return s.ListenAndServe()


perf/pulsar-perf-go.go
```diff
@@ -69,6 +70,15 @@ func main() {
 		Use: "pulsar-perf-go",
 	}
 
+	go func() {
```
cckellogg: In my opinion this should not be started by default and should be controlled with a cli flag. Also, if you run two process only one server will start.

wolfstudy: By default, `pprof` does not collect project metrics. So it will not cause any loss in performance. 

But if we do this, we can provide the user with a `pprof` which is directly accessed by the browser. What's wrong?

In this way, people who use the pulsar-go-client in the production environment can also monitor their running status through the browser at all times, instead of collecting each time through the `go tool pprof`.

> Also, if you run two processes only one server will start. Why is this being added?

Are you sure? When testing in my local environment, we can access it in the browser through `localhost: 8889`, or we can use `go tool pprof -http=:8081 $HOME/pprof/pprof.samples.cpu.004.pb.gz` to access.

wolfstudy: This feature is also raised by users using pulse-client-go. I think this feature is reasonable, we should provide this way, what do you think?

cckellogg: I'm not against the feature i just think it should be hidden behind a cli flag and not started by default. There is already a FlagProfile and a profiling server is started when that is enabled. Can this profiling be added to that server? I'm confused on why two need to be started.

If you start `pulsar-perf produce` and `pulsar-perf consume` on your laptop won't the server try to bind to same port and one of them not start?

wolfstudy:
@cckellogg PTAL thanks.

perf/pulsar-perf-go.go
```diff
@@ -98,12 +99,33 @@ func stopCh() <-chan struct{} {
...
 	go func() {
-		if err := serveProfiling("0.0.0.0:6060", stop); err != nil && err != http.ErrServerClosed {
+		addr := getIntranetIP()
+		if err := serveProfiling(addr, stop); err != nil && err != http.ErrServerClosed {
 			log.WithError(err).Error("Unable to start debug profiling server")
 		}
 	}()
 }
 
+func getIntranetIP() string {
```
cckellogg: Why is this needed? We would still run into a port binding issues when running a producer and consumer right?

wolfstudy:
@cckellogg PTAL again thanks.

PR #212, opened 2020-04-11
[Issue 211] Fix usage of MessageChannel in Reader

### LINKED ISSUE #211: ReaderOptions.MessageChannel is ignored
#### Expected behavior

Reader uses the `ReaderOptions.MessageChannel` channel to send messages.

#### Actual behavior

Reader never sends to that channel if set.

#### Steps to reproduce

Search for code references to `ReaderOptions.MessageChannel`.

cornelk:
Fixes #211

### Motivation

- `ReaderOptions.MessageChannel` was never used
- `ReaderMessage.Reader` was never set

### Modifications

- removed `ReaderMessage` as it was not really used
- `ConsumerMessage.Consumer` type was changed to `interface{}` and can be set to `Consumer` or `Reader` now 

### Verifying this change

CI is failing due to an unrelated race condition.

- new unit tests can be added if the changes are accepted.

  - The public API: (**yes**)
  - The schema: (don't know)
  - The wire protocol: (no)

### Documentation

  - Does this pull request introduce a new feature? (no, actually implements a documented feature)

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -24,7 +24,7 @@ import (
 
 // Pair of a Consumer and Message
 type ConsumerMessage struct {
-	Consumer
+	Consumer interface{} // can be of type Consumer or Reader
 	Message
 }
 
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -82,7 +82,7 @@ type partitionConsumer struct {
 	client *client
 
 	// this is needed for sending ConsumerMessage on the messageCh
-	parentConsumer Consumer
+	parentConsumer interface{}
 	state          consumerState
 	options        *partitionConsumerOpts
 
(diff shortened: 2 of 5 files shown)


pulsar/consumer.go
```diff
@@ -24,7 +24,7 @@ import (
 
 // Pair of a Consumer and Message
 type ConsumerMessage struct {
-	Consumer
```
merlimat, line 27: this changes the user API. having a reference to the `Consumer` object is to give the chance to the user to call `consumer.Ack()` with the message. If we pass an `interface{}` it would have to downcast

cornelk: what other solution do you propose to fix this bug?

bithavoc:
@cornelk is there an alternative PR to fix this? why was it closed?

PR #214, opened 2020-04-12
Fix Consumer Seek not returning a potential error

cornelk:
### Motivation

Review of Consumer Seek() did reveal a potential error that is not returned as such.

### Modifications

- refactored `messageID()` to not use a logger but as a helper function that returns an error. It can later be used by a currently missing `Reader` `Seek()` implementation.
- return a potential error from `Consumer.Seek()`
- log the error in `AckID()` and `NackID()` - changing the interface to return an error for those function should be a better approach

### Verifying this change

This change is already covered by existing tests

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -265,8 +265,9 @@ func (c *consumer) Ack(msg Message) {
 
 // Ack the consumption of a single message, identified by its MessageID
 func (c *consumer) AckID(msgID MessageID) {
-	mid, ok := c.messageID(msgID)
-	if !ok {
+	mid, err := newMessageIDFromInterface(msgID, len(c.consumers))
+	if err != nil {
+		c.log.Warnf(err.Error())
 		return
 	}
 
@@ -283,8 +284,9 @@ func (c *consumer) Nack(msg Message) {
 }
 
 func (c *consumer) NackID(msgID MessageID) {
-	mid, ok := c.messageID(msgID)
-	if !ok {
+	mid, err := newMessageIDFromInterface(msgID, len(c.consumers))
+	if err != nil {
+		c.log.Warnf(err.Error())
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pulsar/impl_message.go
```diff
@@ -118,6 +120,24 @@ func newMessageID(ledgerID int64, entryID int64, batchIdx int, partitionIdx int)
 	}
 }
 
+// creates a new messageID object from the given MessageID, consumers is the amount of
+// consumers, 1+ for Consumer, 1 for Reader.
+func newMessageIDFromInterface(msgID MessageID, consumers int) (*messageID, error) {
```
cckellogg, line 125: This is not returning a new message id it's the same struct right?

pulsar/consumer_impl.go
```diff
@@ -283,8 +284,9 @@ func (c *consumer) Nack(msg Message) {
 }
 
 func (c *consumer) NackID(msgID MessageID) {
-	mid, ok := c.messageID(msgID)
-	if !ok {
+	mid, err := newMessageIDFromInterface(msgID, len(c.consumers))
```
cckellogg, line 287: I'm confused what is the missing error?

cornelk: if the `msgID` interface paramter is not of the internal type `messageID` `newMessageIDFromInterface()` will fail

pulsar/impl_message.go
cornelk: msgID is an interface, the returned value the internal struct

pulsar/consumer_impl.go
cckellogg: Why is that error needed? It looks like it's just being logged and not returned to the client?

pulsar/impl_message.go
cckellogg: The method name is confusing to me something with new should return a new object. This is just casting it to something.

cornelk: I renamed and moved the function so that it later can be used by `Reader` as well. This function existed before, it just moved. It converts the given interface implementation to the internal type. If this conversion fails, the callers like NackID can not succeed and should return an error to their caller. I propose this change instead of logging and not returning an error. This would be an API change tho.

PR #221, opened 2020-04-14
Auto update the client to handle changes in number of partitions

merlimat:
### Motivation

Client should automatically discover new partitions as they're added.

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -41,9 +41,11 @@ type acker interface {
 }
 
 type consumer struct {
-	client    *client
-	options   ConsumerOptions
-	consumers []*partitionConsumer
+	topic        string
+	client       *client
+	options      ConsumerOptions
+	consumers    []*partitionConsumer
+	consumerName string
 
 	// channel used to deliver message to clients
 	messageCh chan ConsumerMessage
@@ -52,6 +54,7 @@ type consumer struct {
 	closeOnce sync.Once
 	closeCh   chan struct{}
 	errorCh   chan error
+	ticker    *time.Ticker
 
 	log *log.Entry
 }
@@ -118,9 +121,12 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 	return nil, newError(ResultInvalidTopicName, "topic name is required for consumer")
 }
 
-func internalTopicSubscribe(client *client, options ConsumerOptions, topic string,
+
+func newInternalConsumer(client *client, options ConsumerOptions, topic string,
 	messageCh chan ConsumerMessage, dlq *dlqRouter) (*consumer, error) {
+
 	consumer := &consumer{
+		topic:     topic,
 		client:    client,
 		options:   options,
 		messageCh: messageCh,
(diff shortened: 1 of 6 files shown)


pulsar/consumer_impl.go
```diff
@@ -130,56 +135,106 @@ func internalTopicSubscribe(client *client, options ConsumerOptions, topic strin
...
-	numPartitions := len(partitions)
-	consumer.consumers = make([]*partitionConsumer, numPartitions)
+	// set up timer to monitor for new partitions being added
+	duration := options.AutoDiscoveryPeriod
+	if duration <= 0 {
+		duration = defaultAutoDiscoveryDuration
+	}
+	consumer.ticker = time.NewTicker(duration)
+
+	go func() {
```
cckellogg, line 156: Should we only start this if the topic is partitioned?

pulsar/consumer_impl.go
```diff
@@ -130,56 +135,106 @@ func internalTopicSubscribe(client *client, options ConsumerOptions, topic strin
...
+func (c *consumer) internalTopicSubscribeToPartitions() error {
+	partitions, err := c.client.TopicPartitions(c.topic)
+	if err != nil {
+		return err
+	}
+
+	oldNumPartitions := 0
+	newNumPartitions := len(partitions)
+
+	if c.consumers != nil {
```
cckellogg, line 175: do we need a lock for c.consumers? We use this for acking/nacking messages right?

pulsar/producer_impl.go
```diff
@@ -73,25 +83,67 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
...
+	err := p.internalCreatePartitionsProducers()
 	if err != nil {
 		return nil, err
 	}
 
-	numPartitions := len(partitions)
-	p.producers = make([]Producer, numPartitions)
+	p.ticker = time.NewTicker(partitionsAutoDiscoveryInterval)
+
+	go func() {
```
cckellogg, line 93: Will this run for non-partitioned topics as well? Is that ok?

pulsar/producer_impl.go
```diff
@@ -73,25 +83,67 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
...
+func (p *producer) internalCreatePartitionsProducers() error {
+	partitions, err := p.client.TopicPartitions(p.topic)
+	if err != nil {
+		return err
+	}
+
+	oldNumPartitions := 0
+	newNumPartitions := len(partitions)
+
+	if p.producers != nil {
```
cckellogg, line 112: Do we need a lock for the producers?

pulsar/consumer_impl.go
merlimat: Problem is that the `TopicPartitions()` will just return a list, so we don't necessarely know if it's 1 partition or no partitions. In any case, if the topic is not partitioned, it would not be possible to create a partitioned topic with same name.

pulsar/consumer_impl.go
merlimat: Good point, the changes to the `c.consumers` and `p.producers` should be protected by mutex to avoid races.

pulsar/producer_impl.go
merlimat: Added

pulsar/consumer_impl.go
```diff
@@ -130,56 +136,108 @@ func internalTopicSubscribe(client *client, options ConsumerOptions, topic strin
...
+
+		c.log.WithField("old_partitions", oldNumPartitions).
+			WithField("new_partitions", newNumPartitions).
+			Info("Changed number of partitions in topic")
+	}
+
+	consumers := make([]*partitionConsumer, newNumPartitions)
+
+	// Copy over the existing consumer instances
+	for i := 0; i < oldNumPartitions; i++ {
```
jerrypeng, line 195: Why isn't there a mutex here when we are setting new consumers to old consumers?  Should this whole block in a mutux?

https://github.com/apache/pulsar-client-go/pull/221/files#diff-8b65201e8d71d0f36fdcbdfc7b5d5f0aR176-R197

pulsar/consumer_impl.go
```diff
@@ -130,56 +136,108 @@ func internalTopicSubscribe(client *client, options ConsumerOptions, topic strin
...
+		for range consumer.ticker.C {
+			consumer.log.Debug("Auto discovering new partitions")
+			consumer.internalTopicSubscribeToPartitions()
+		}
+	}()
+
+	return consumer, nil
+}
+
+func (c *consumer) internalTopicSubscribeToPartitions() error {
```
jerrypeng, line 167: Shouldn't this whole method be under a mutex instead of only where c.consumers is be in set?  There could be race conditions if "Unsubscribe", "Close", etc are being called while this method is running.

jerrypeng: I think it make sense that that methods like "Unsubscribe", "Close" and "internalTopicSubscribeToPartitions" cannot run at the same time

merlimat: Sure, that makes sense.

pulsar/producer_impl.go
```diff
@@ -73,25 +87,69 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
...
+		for range p.ticker.C {
+			p.log.Debug("Auto discovering new partitions")
+			p.internalCreatePartitionsProducers()
+		}
+	}()
+
+	return p, nil
+}
+
+func (p *producer) internalCreatePartitionsProducers() error {
```
jerrypeng, line 107: Same comment as consumers

pulsar/consumer_impl.go
merlimat: Updated

jerrypeng:
General question about how locks are handled in Go, when multiple threads are waiting on a lock. In what order are the locks released? FIFO?  Or just everyone attempts to acquire and first one wins?  I am asking because I am wondering if there are going to be any thread starvation going on for the thread that periodically checks for new partitions especially when "sendAsync" can be called quite frequently.

merlimat:
Yes, I was trying to avoid getting the extra mutex on the write path, though I think it shouldn't be too bad. 

> when multiple threads are waiting on a lock. In what order are the locks released? FIFO? 

I haven't really seen it specified, though a mutex always should include some sort of queuing for fairness. I'm not worried about that aspect. Also, the thread that updates the list of producers is not "critical" in the sense that it only needs to get executed eventually, with no hard time bound.

PR #222, opened 2020-04-14
WIP: [Issue 218] Add Seek functions to Reader

### LINKED ISSUE #218: Seek function is missing in Reader
`Seek()` and `SeekByTime()` are available in `Consumer` but not `Reader`. The C++ Client has the function in both classes.

cornelk:
Fixes #218 

WIP due to `TestReaderSeek` hanging when executing `reader.Next()` after `reader.Seek()` - not sure why this happens and how to continue here, any help or hint is welcome.

### Motivation

`Seek()` was missing in `Reader` interface but is available in the C++ client lib.

### Modifications

* implemented Seek in the similar way as `Consumer` does

### Verifying this change

- `TestReaderSeek*`

  - The public API: (**yes**)
  - The schema: (don't know)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/reader_impl.go
+++ b/pulsar/reader_impl.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"context"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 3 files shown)


wolfstudy:
```
panic: test timed out after 20m0s
```

wolfstudy:
Move this change to 0.2.0

pulsar/reader_impl.go
```diff
@@ -150,3 +151,33 @@ func (r *reader) hasMoreMessages() bool {
...
+
+	partition := mid.partitionIdx
+	// did we receive a valid partition index?
+	if partition != 0 {
+		r.log.Warnf("invalid partition index %d expected 0", partition)
+		return nil, false
+	}
+
+	return mid, true
+}
```
nitishv, line 170: Given the method name, the purpose of this method is to convert MessageID to get *messageID
In doing so failure in type assertion is not really a failure. The given msgID can still be converted. Refer to a this fix I just made https://github.com/apache/pulsar-client-go/pull/305

I think that change can be re-factored into this unexported method for reader, as it being used in couple of places now with this change.
Also checking for non-zero partition index does not seem to be the purpose of this method, and should not be included here. That check is only relevant to Seek APIs, and should be moved there.

pulsar/reader_test.go
```diff
@@ -362,3 +365,108 @@ func TestReaderHasNext(t *testing.T) {
 
 	assert.Equal(t, 10, i)
 }
+
+func TestReaderSeek(t *testing.T) {
```
nitishv, line 369: There can be an added test case which does a Seek on a custom MessageID refer to https://github.com/apache/pulsar-client-go/pull/305

wolfstudy:
ping @cornelk Can you check the @nitishv comments and fix the logic of `seek` for reader.

cornelk:
> ping @cornelk Can you check the @nitishv comments and fix the logic of `seek` for reader.

pong. I am focusing on my own Pulsar Go Client now, someone else can take over this MR

PR #225, opened 2020-04-15
Added an function to serialize the messageID to string

skyrocknroll:
MessageID to String Serialization

### DIFF
--- a/pulsar/impl_message.go
+++ b/pulsar/impl_message.go
@@ -18,6 +18,7 @@
 package pulsar
 
 import (
+	"fmt"
 	"math/big"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


skyrocknroll:
@merlimat Can you please review it?

pulsar/impl_message.go
```diff
@@ -93,6 +94,10 @@ func (id *messageID) Serialize() []byte {
 	data, _ := proto.Marshal(msgID)
 	return data
 }
+func (id *messageID) SerializeToString() string {
+
+	return fmt.Sprintf("%d:%d:%d:%d", id.ledgerID, id.entryID, id.batchIdx, id.partitionIdx)
```
merlimat, line 99: My concern here is that we're exposing the String format in the API directly. The string format itself shouldn't be used for any other purpose than debug, since it's not guaranteed to be kept compatible.

For that, if the purpose of this is to print a message id on logs, wouldn't that already work even without the `SerializeToString()`? 
For example, check https://github.com/apache/pulsar-client-go/blob/master/examples/consumer/consumer.go#L52-L53 There, all the internal components of the message id are going to get printed on the log.

skyrocknroll: Make sense. Thanks @merlimat

Issue #171, opened 2020-01-11
Enable TLS hostname verification by default


EronWright:
#### Expected behavior

The client should use hostname verification by default, especially when using the host's CA bundle.
Hostname verification is a critical aspect of TLS when using public root certificates, to prevent an [MITM attack](https://en.wikipedia.org/wiki/Man-in-the-middle_attack).  See the famous paper: ["The Most Dangerous Code in the World"](http://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf) which emphasizes the need for hostname verification.

_I see a possible related issue, that the `ServerName` field in `tls.Config` should be set in all cases, to support SNI.  It is possible to separate certificate verification from hostname verification (see golang/go#21971)._

#### Actual behavior

The client doesn't use hostname verification by default.

#### Steps to reproduce

Create a pulsar client with TLS enabled and  with default options (which corresponds to `pulsar.ClientOptions { ValidateHostname: false }`), and observe that hostname verification does not occur.

#### System configuration
**Pulsar-client-go version**: `master`

PR #238, opened 2020-05-03
Implement TLS VerifyPeerCertificate callback to skip hostname verfication

zzzming:
### Contribution Checklist
  
This PR addresses a problem how to disable TLS ValidateHostname. The current implementation with empty `tlsConfig.ServerName` would not work, because the connected server host will be inferred in absence of tlsConfig.ServerName by Go Tls library. One of the use cases is when DNS name in the server certificate returned does not match the broker host name that the client connects to. The client can be deployed within the same Pulsar kubernetes cluster if the client connects to the internal proxy or broker host directly instead of the public fqdn. This problem may also rise for self-signed cert.

This specific problem and solution are described by this issue report,  https://github.com/golang/go/issues/21971

This PR implements a TLS VerifyPeerCertificate callback to skip host name validation if any client chooses to disable TLSValidateHostname in ClientOption.

I understand TLSValidateHostname is false by default because Go initializes bool as `false`. There is an existing issue #171 that is tracking the problem. I think it will open up a discussion how to support backward compatibility that might require consensus from the community. Therefore, altering the current default is beyond the scope of this PR.

  - The public API: (yes) Yes the current implementation of skip hostname check is broken, this is a fix.
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/client_impl_test.go
+++ b/pulsar/client_impl_test.go
@@ -67,6 +67,24 @@ func TestTLSInsecureConnection(t *testing.T) {
 	client.Close()
 }
 
+func TestTLSInsecureConnectionWithCerts(t *testing.T) {
+	client, err := NewClient(ClientOptions{
+		URL:                        serviceURLTLS,
+		TLSAllowInsecureConnection: true,
+		TLSTrustCertsFilePath:      caCertsPath,
+	})
+	assert.NoError(t, err)
+
+	producer, err := client.CreateProducer(ProducerOptions{
+		Topic: newTopicName(),
+	})
+
+	assert.NoError(t, err)
+	assert.NotNil(t, producer)
+
+	client.Close()
+}
+
 func TestTLSConnection(t *testing.T) {
 	client, err := NewClient(ClientOptions{
 		URL:                   serviceURLTLS,
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -716,7 +716,7 @@ func (c *connection) getTLSConfig() (*tls.Config, error) {
 	} else if !tlsConfig.InsecureSkipVerify {
 		// Solution is credited to https://github.com/golang/go/issues/21971
 		// Code is adapted from the original implementation of handshake_client.go at
-		// 		https://github.com/golang/go/blob/81555cb4f3521b53f9de4ce15f64b77cc9df61b9/src/crypto/tls/handshake_client.go#L327-L344,%20but%20adapted%20to%20skip%20the%20hostname%20verification
+		// https://github.com/golang/go/blob/master/src/crypto/tls/handshake_client.go#L804
 		// disable the default verification; use customized VerifyPeerCertificate
 		tlsConfig.InsecureSkipVerify = true
 		tlsConfig.VerifyPeerCertificate = func(rawCerts [][]byte, certChain [][]*x509.Certificate) error {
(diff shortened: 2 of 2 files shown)


wolfstudy (approved):
cool, LGTM +1

pulsar/internal/connection.go
```diff
@@ -713,6 +713,41 @@ func (c *connection) getTLSConfig() (*tls.Config, error) {
 
 	if c.tlsOptions.ValidateHostname {
 		tlsConfig.ServerName = c.physicalAddr.Hostname()
```
EronWright, line 715: To support SNI, shouldn't the `tlsConfig.ServerName` be set in all cases?

EronWright, line 715: To support SNI, shouldn't the `tlsConfig.ServerName` be set in all cases?

zzzming: @EronWright you are right that ServerName should be set as suggested by https://github.com/golang/go/blob/62a3f2e27c7732656bb3ae8f14047b74a9956e77/src/crypto/tls/common.go#L542
I think the problem is the default value of TLSValidateHostname is `false`. It should have been called `DisabledTLSValidateHostname`. Do you have any suggestion how to handle backward compatibility of TLSValidateHostname? 
I made an update to add ServerName but only skip verification if InsecureSkipVerify is true.

wolfstudy:
Move this change to 0.2.0

wolfstudy:
ping @EronWright PTAL

pulsar/client_impl_test.go
```diff
@@ -67,6 +67,24 @@ func TestTLSInsecureConnection(t *testing.T) {
 	client.Close()
 }
 
+func TestTLSInsecureConnectionWithCerts(t *testing.T) {
```
EronWright, line 70: This test doesn't seem to do anything substantive.  I'd expect it to exercise the resultant `tls.Config` in some way, maybe by making an actual TLS connection.

pulsar/internal/connection.go
```diff
@@ -711,8 +711,46 @@ func (c *connection) getTLSConfig() (*tls.Config, error) {
 		}
 	}
 
-	if c.tlsOptions.ValidateHostname {
-		tlsConfig.ServerName = c.physicalAddr.Hostname()
+	tlsConfig.ServerName = c.physicalAddr.Hostname()
+
+	if tlsConfig.InsecureSkipVerify {
```
EronWright, line 716: Why isn't `c.tlsOptions.ValidateHostname` used anywhere?  I was thinking:

```
tlsConfig := &tls.Config{
    // disable built-in verification if hostname verification is to be skipped
		InsecureSkipVerify: c.tlsOptions.AllowInsecureConnection || !c.tlsOptions.ValidateHostname
	}

if !c.tlsOptions.AllowInsecureConnection && tlsConfig.InsecureSkipVerify {
  // implement a limited verification routine that checks the certificate but skips hostname verification 
  tlsConfig.VerifyPeerCertificate = func() {...}
}
```

pulsar/internal/connection.go
```diff
@@ -711,8 +711,46 @@ func (c *connection) getTLSConfig() (*tls.Config, error) {
...
+				cert, err := x509.ParseCertificate(asn1Data)
+				if err != nil {
+					return fmt.Errorf("tls: failed to parse server certificate error: %s", err.Error())
+				}
+				certs[i] = cert
+			}
+
+			if tlsConfig.RootCAs == nil {
+				return nil
+			}
```
EronWright, line 735: How would this deal with the common scenario where the client is using the system's bundle?  I am unsure of whether `RootCAs` is `nil` in that scenario.

EronWright (changes requested):

snowcrumble:
Why do not add an option let user set an arbitrary `tls.Config`

wolfstudy:
ping @zzzming Can you check @EronWright comments? thanks.

nodece:
This PR is cool, any updates?

PR #247, opened 2020-05-13
[Issue 246][compression.Provider] Not sharing compression.Provider among producers and consumers

### LINKED ISSUE #246: compression.Provider should not be shared

Currently the producers and consumers are all sharing global compression providers.

https://github.com/apache/pulsar-client-go/blob/e7f1673350f208b5063823282d14906d70d66904/pulsar/internal/compression/compression.go#L36-L41

And some compression providers are not thread safe, for example the zstd provider will reusing its zstd encoder. In the case of utilizing multiple consumers or producers, the provider may generate wrong compression/decompression results. 

Let each producer and consumer to keep its own compression providers.

rueian:
Fixes #246

### Motivation

Not sharing compression.Provider globally to avoid race condition.

### Modifications

Aliasing the `compression.XXXXProvider` to `compression.NewXXXXProvider` functions.
Keep providers in each `partitionConsumer` and `BatchBuilder`


pulsar/consumer_partition.go
```diff
@@ -875,6 +868,20 @@ func (pc *partitionConsumer) discardCorruptedMessage(msgID *pb.MessageIdData,
 		})
 }
 
+func (pc *partitionConsumer) getCompressionProvider(compressionType pb.CompressionType) (
+	provider compression.Provider, ok bool) {
+	if pc.compressionProviders == nil {
+		pc.compressionProviders = map[pb.CompressionType]compression.Provider{
```
cckellogg, line 874: Will each consumer each have their own reference to all compressions providers? If so why not just preallocate when the partition consumer is created?

rueian: Not all providers are thread safe. For example currently the zstdProvider reusing its encoder when calling `Compress(data)`. Therefore I think each consumer and each producer should have their own set of compression providers to avoid racing on the global shared providers. And yes, it will have overhead.

Yes, preallocating will make the code here simpler, I will change that.

cckellogg: Do you see a race condition? Looking at the source code it seems like it is thread safe?
https://github.com/klauspost/compress/blob/master/zstd/encoder.go#L415

Am I missing something?

rueian: Oh, Sorry. It is my mistake that doesn't check on the source. The EncodeAll is thread safe.
I think I can close this PR and issue now.

PR #249, opened 2020-05-16
[Issue 242][pulsar-client-go] feature: add send timeout 

### LINKED ISSUE #242: send add timeout
#### Expected behavior

I want add timeout on producer send func, the send func need pass a context parameter，so I pass a WithTimeout context 

#### Actual behavior

I have test and find that the WithTimeout context seem not work on producer send func.The producer send func not return after timeout.

#### my code

`

	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
	defer cancel()

	id, err := s.config.DcSyncProducer.Send(ctx, msg)
	if err != nil{
		...
	}
`
above Send func is interface Producer Send method
`

	// Send a message
	// This call will be blocking until is successfully acknowledged by the Pulsar broker.
	// Example:
	// producer.Send(ctx, pulsar.ProducerMessage{ Payload: myPayload })
	Send(context.Context, *ProducerMessage) (MessageID, error)

	// SendAsync a message in asynchronous mode
	// The callback will report back the message being published and
	// the eventual error in publishing
	SendAsync(context.Context, *ProducerMessage, func(MessageID, *ProducerMessage, error))

`

#### System configuration
Pulsar version: 2.5.0
pulsar-client-go: lastest

WJL3333:
Fixes #242 

### Motivation
add send timeout

### Modifications

1. ProduceOptions
add SendTimeout and SendTimeoutCheckInterval in ProducerOptions

2. BlockingQueue
unexported BlockingQueue.Iterator
and add methods to avoid race condition.
BlockingQueue.IterateIfNonEmpty and BlockingQueue.PollIfSatisfy

3. partitionProducer
add a ticker to check if pendingItem in pendingQueue has expired.

### Verifying this change

add some tiny test in producer_test.go

  - The public API: no
  - The schema:  no 
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/internal/batch_builder.go
+++ b/pulsar/internal/batch_builder.go
@@ -57,6 +57,8 @@ type BatchBuilder struct {
 	callbacks   []interface{}
 
 	compressionProvider compression.Provider
+	createAt            time.Time
+	sendTimeout         time.Duration
 }
 
 // NewBatchBuilder init batch builder and return BatchBuilder pointer. Build a new batch message container.
@@ -105,7 +107,7 @@ func (bb *BatchBuilder) hasSpace(payload []byte) bool {
 
 // Add will add single message to batch.
 func (bb *BatchBuilder) Add(metadata *pb.SingleMessageMetadata, sequenceID uint64, payload []byte,
-	callback interface{}, replicateTo []string, deliverAt time.Time) bool {
+	callback interface{}, replicateTo []string, deliverAt time.Time, createAt time.Time, sendTimeout time.Duration) bool {
 	if replicateTo != nil && bb.numMessages != 0 {
 		// If the current batch is not empty and we're trying to set the replication clusters,
 		// then we need to force the current batch to flush and send the message individually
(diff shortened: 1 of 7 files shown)


merlimat:
The `Send()` is already taking a `Context` object. If the context has a timeout, it will be used.

merlimat:
I think the correct fix for #242 would be to also check the context  **before** attempting to publish the message.

WJL3333:
yeah, i think it should be, but when user pass different context with different sendtimeout to Send and SendAsync , this will introduce a lot of complicity. the message is send as batch. and each message in batch with different timeout

WJL3333:
i'm confused about the above use case, need some guide on the real popuse of passing an context here. and where do you prefer to check the context？ just before send or both before send and waiting for response.

yarthur1:
I hope that both before send and time wait for response should check timeout.I don't want to wait send func always,because I meet this situation where pulsar broker has something wrong and send func always wait @

WJL3333:
closed and goto https://github.com/apache/pulsar-client-go/pull/252

PR #253, opened 2020-05-18
Improve go client sending performance

zr-hebo:
Improve the sending performance to 50000 message per second, by

  - Add buffer pool, avoid make slice every time need getting buffer;
  - Increased buffer size in various place.

### DIFF
--- a/pulsar/internal/batch_builder.go
+++ b/pulsar/internal/batch_builder.go
@@ -34,7 +34,7 @@ const (
 	// MaxBatchSize will be the largest size for a batch sent from this particular producer.
 	// This is used as a baseline to allocate a new buffer that can hold the entire batch
 	// without needing costly re-allocations.
-	MaxBatchSize = 128 * 1024
+	MaxBatchSize = 1024 * 1024
 	// DefaultMaxMessagesPerBatch init default num of entries in per batch.
 	DefaultMaxMessagesPerBatch = 1000
 )
@@ -66,7 +66,7 @@ func NewBatchBuilder(maxMessages uint, producerName string, producerID uint64,
 		maxMessages = DefaultMaxMessagesPerBatch
 	}
 	bb := &BatchBuilder{
-		buffer:       NewBuffer(4096),
+		buffer:       NewBuffer(MaxBatchSize),
 		numMessages:  0,
 		maxMessages:  maxMessages,
 		producerName: producerName,
@@ -162,7 +162,7 @@ func (bb *BatchBuilder) Flush() (batchData []byte, sequenceID uint64, callbacks 
 	compressed := bb.compressionProvider.Compress(bb.buffer.ReadableSlice())
 	bb.msgMetadata.UncompressedSize = &uncompressedSize
 
-	buffer := NewBuffer(4096)
+	buffer := NewBuffer(MaxBatchSize)
 	serializeBatch(buffer, bb.cmdSend, bb.msgMetadata, compressed)
 
 	callbacks = bb.callbacks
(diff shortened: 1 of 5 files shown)


pulsar/internal/batch_builder.go
```diff
@@ -34,7 +34,7 @@ const (
 	// MaxBatchSize will be the largest size for a batch sent from this particular producer.
 	// This is used as a baseline to allocate a new buffer that can hold the entire batch
 	// without needing costly re-allocations.
-	MaxBatchSize = 128 * 1024
+	MaxBatchSize = 1024 * 1024
```
merlimat, line 37: The reason for using 128 KB is that we measured that there's no advantage in using bigger batches, as in we can easily saturate the network bandwidth already.

pulsar/internal/batch_builder.go
```diff
@@ -66,7 +66,7 @@ func NewBatchBuilder(maxMessages uint, producerName string, producerID uint64,
 		maxMessages = DefaultMaxMessagesPerBatch
 	}
 	bb := &BatchBuilder{
-		buffer:       NewBuffer(4096),
+		buffer:       NewBuffer(MaxBatchSize),
```
merlimat, line 69: That's increasing the min memory size for each producer from 4KB to 1MB. That would be very bad for overall memory usage.

pulsar/internal/buffer.go
```diff
@@ -77,13 +86,43 @@ type buffer struct {
 
 // NewBuffer creates and initializes a new Buffer using buf as its initial contents.
 func NewBuffer(size int) Buffer {
+	var memSeg []byte
+	memSegmentQueue := getMemorySegmentQueue(size)
```
merlimat, line 90: This will introduce a contention point on a global mutex, plus reading from a global channel, which again uses a mutex. I'd like to see some data indicating how this change is going to perform compared to the current code.

pulsar/internal/connection.go
```diff
@@ -166,7 +166,7 @@ func newConnection(logicalAddr *url.URL, physicalAddr *url.URL, tlsOptions *TLSO
 		connectionTimeout:    connectionTimeout,
 		logicalAddr:          logicalAddr,
 		physicalAddr:         physicalAddr,
-		writeBuffer:          NewBuffer(4096),
+		writeBuffer:          NewBuffer(512 * 1024),
```
merlimat, line 169: The buffer is auto-expanding as needed. I don't think increasing the minimum size to 0.5MB is going to be making any difference other than increasing the overall memory usage.

pulsar/internal/connection_reader.go
```diff
@@ -36,7 +36,7 @@ func newConnectionReader(cnx *connection) *connectionReader {
 	return &connectionReader{
 		cnx:    cnx,
 		reader: bufio.NewReader(cnx.cnx),
-		buffer: NewBuffer(4096),
+		buffer: NewBuffer(4 * 1024 * 1024),
```
merlimat, line 39: Again, the buffer is auto-expanding. Increasing here from 4 KB to 4 MB is going to just blow up the memory with no advantage.

merlimat:
> Improve the sending performance to 50000 message per second, by

Have you tried to benchmark the client using non-persistent topics?

zr-hebo:
> > Improve the sending performance to 50000 message per second, by
> 
> Have you tried to benchmark the client using non-persistent topics?

I did benchmark and code before this PR: https://github.com/apache/pulsar-client-go/pull/209, in order to get better performance I did some thing. After reading your comments, I think the optimization I did is not necessary.

zr-hebo:
I will do more benchmark, and commit PR if necessary.

PR #254, opened 2020-05-18
Fix send async comments

### LINKED ISSUE #208: Support `producer.SendAsync` not blocking if queue is full
Currently, `producer.SendAsync` can quickly become a blocking call if either:
1. The queue (set with maxPendingMessages) becomes full
2. The event channel (for enqueueing, handling disconnects, etc) becomes full (with a hardcoded 10 messages)

This can be really undesirable as it is unexpected that `SendAsync` becomes a blocking call. 

The golang producer needs an option equivalent to the java `producerBuilder.blockIfQueueFull` call. In this mode, any `SendAsync` call against a full queue (case 1 above) should result in an error.  The semaphore implementation around the maxPendingMessages does not have a timeout or tryAcquire, but if it did, it should be easy to make work (perhaps consider switching to golang/x/sync/semaphore package?) and in the event the semaphore can't be acquired, then pop an error.

It seems to also make sense that if the internal eventChannel is also full, we should also error (case 2 above). For example, this can happen when a producer is disconnected and is waiting for a re-connect. The call to re-connect is blocking and (AFAICT) blocks further messages being processed on the eventChannel, which quickly fills up and then results in a blocking behavior on `SendAsync`. A non-blocking send into the eventChannel would probably work here, though we may need to increase the `eventChannel` size a bit

None

wolfstudy:
Fixes #208 

### Motivation

- Fix send async comments

### DIFF
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -128,7 +128,7 @@ type Producer interface {
 	Send(context.Context, *ProducerMessage) (MessageID, error)
 
 	// SendAsync a message in asynchronous mode
-	// This call will is blocked when the `event channel` becomes full (default: 10) or the
+	// This call is blocked when the `event channel` becomes full (default: 10) or the
 	// `maxPendingMessages` becomes full (default: 1000)
 	// The callback will report back the message being published and
 	// the eventual error in publishing
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -128,6 +128,8 @@ type Producer interface {
 	Send(context.Context, *ProducerMessage) (MessageID, error)
 
 	// SendAsync a message in asynchronous mode
+	// This call will be blocking when the `event channel` becomes full (default: 10) and the
+	// maxPendingMessages becomes full (default: 1000)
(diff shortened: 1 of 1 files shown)


pulsar/producer.go
```diff
@@ -128,6 +128,8 @@ type Producer interface {
 	Send(context.Context, *ProducerMessage) (MessageID, error)
 
 	// SendAsync a message in asynchronous mode
+	// This call will be blocking when the `event channel` becomes full (default: 10) or the
```
Anonymitaet, line 131:
```suggestion
	// This call is blocked when the `event channel` becomes full (default: 10) or the
```

What's the unit for "10"? 10 MB or  10 messages?

pulsar/producer.go
```diff
@@ -128,6 +128,8 @@ type Producer interface {
...
+	// maxPendingMessages becomes full (default: 1000)
```
Anonymitaet, line 132:
```suggestion
	// `maxPendingMessages` becomes full (default: 1000)
```

PR #255, opened 2020-05-19
Fix perf-produce cannot be closed

wolfstudy:

### Motivation

Due to the use of goroutine, `stop <-chan struct {}` cannot be seen.

### Modifications

- Fix perf-produce cannot be closed

### DIFF
--- a/perf/perf-producer.go
+++ b/perf/perf-producer.go
@@ -106,7 +106,7 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
 
 		for {
 			select {
-			case <-stop:
+			case <-stopCh:
 				return
 			default:
 			}
--- a/perf/perf-producer.go
+++ b/perf/perf-producer.go
@@ -98,7 +98,7 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
 
 	ch := make(chan float64)
 
-	go func() {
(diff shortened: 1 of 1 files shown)


wolfstudy:
@cckellogg PTAL

perf/perf-producer.go
```diff
@@ -98,7 +98,7 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
 
 	ch := make(chan float64)
 
-	go func() {
+	go func(stop <-chan struct{}) {
```
cckellogg, line 101: maybe rename this the stopCh so it's clear in the go routine which one it's using? Just curious what are the steps for it getting stuck?

wolfstudy: > maybe rename this the stopCh so it's clear in the go routine which one it's using?

ok

> Just curious what are the steps for it getting stuck?

```
func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
    .....
    go func() {
		for {
			select {
			case <-stop:
				return
			
			....
             }
	}()
    ......
}
```

In gorutine, the external `stop` parameter is not visible to it, so the program is blocked inside this gorutine

cckellogg: Please remove the `perf/pulsar-perf-go` bin file

wolfstudy:
run integration tests

PR #276, opened 2020-06-11
Allow to have multiple connections per broker

merlimat:
### Motivation

Similar to what we do in Java client, in certain conditions it might be useful to have more than one TCP connection to each broker. This increase the parallelism in client (more Go routines), brokers (connections get assigned to different IO threads) and also can improve the throughput at TCP level when the client is talking to a small number of brokers.


srkukarni:
Do we want some tests as well?

PR #284, opened 2020-06-17
Introduced lifecycle for compression providers

merlimat:
### Motivation

Use individual instances of compression providers for each producer/consumer, with a defined lifecycle, instead of a static singleton instance. 

Some compression implementation can use private state to improve the performance. Typically these need to be used by a single go-routine. For this, it's better to have private instances.


pulsar/consumer_partition.go
```diff
@@ -846,11 +844,15 @@ func getPreviousMessage(mid *messageID) *messageID {
...
-		err := fmt.Errorf("unsupported compression type: %v", msgMeta.GetCompression())
-		pc.log.WithError(err).Error("Failed to decompress message.")
-		return nil, err
+		var err error
+		if provider, err = pc.initializeCompressionProvider(msgMeta.GetCompression()); err != nil {
+			pc.log.WithError(err).Error("Failed to decompress message.")
+			return nil, err
+		}
+
+		pc.compressionProviders[msgMeta.GetCompression()] = provider
```
cckellogg, line 855: Do we need to close these providers when we the consumer is closed?

merlimat: Yes, it's done at line 683 above

Issue #288, opened 2020-06-19
failed to create consumer with ReceiverQueueSize set to -1


shohi:
#### Expected behavior

create the consumer successfully. According to the doc, ReceiverQueueSize set to `-1` only disables prefetching in consumer.

#### Actual behavior
Panic when creating consumer with the setting.

```
panic: makechan: size out of range
```

Test code
```go
func TestConsumer_NegativeQueueSize(t *testing.T) {
	client, err := pulsar.NewClient(pulsar.ClientOptions{
		URL: "pulsar://localhost:6650",
	})

	if err != nil {
		t.FailNow()
	}
	defer client.Close()

	consumer, err := client.Subscribe(pulsar.ConsumerOptions{
		Topic:             "topic",
		SubscriptionName:  "my-sub",
		ReceiverQueueSize: -1,
	})
	defer func() {
		if consumer != nil {
			consumer.Close()
		}
	}()

	if err != nil {
		t.FailNow()
	}
}
```

#### System configuration
**Pulsar version**: 2.5.2
**pulsar-client-go**: latest(861d7af)

wolfstudy:
@shohi Sorry, this comment may be wrong, `receiveQueueSize` is essentially the size of golang channel buffer.  Can you help fix the issue?

shohi:
Sure

PR #289, opened 2020-06-19
fix panic when creating consumer with ReceiverQueueSize set to -1

shohi:
Fixes https://github.com/apache/pulsar-client-go/issues/288

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -70,13 +70,8 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		return nil, newError(SubscriptionNotFound, "subscription name is required for consumer")
 	}
 
-	if options.ReceiverQueueSize == 0 {
+	if options.ReceiverQueueSize <= 0 {
 		options.ReceiverQueueSize = 1000
-	}
-
-	// disable receiver queue if queue size is negative
-	if options.ReceiverQueueSize < 0 {
-		options.ReceiverQueueSize = 0
 	}
 
 	// did the user pass in a message channel?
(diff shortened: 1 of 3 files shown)


wolfstudy:
> 	// Set to -1 to disable prefetching in consumer
	ReceiverQueueSize int

@shohi Can you remove the comments from `ReceiverQueueSize`?

shohi:
> > // Set to -1 to disable prefetching in consumer
> > ReceiverQueueSize int
> 
> @shohi Can you remove the comments from `ReceiverQueueSize`?

Or change the description?

pulsar/consumer_impl.go
```diff
@@ -74,6 +74,11 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		options.ReceiverQueueSize = 1000
 	}
 
+	// disable receiver queue if queue size is negative
+	if options.ReceiverQueueSize < 0 {
+		options.ReceiverQueueSize = 0
+	}
```
wolfstudy, line 80: Maybe we can fix it using follows way:

```
	if options.ReceiverQueueSize <= 0 {
		options.ReceiverQueueSize = 1000
	}
```

wolfstudy:
> > > // Set to -1 to disable prefetching in consumer
> > > ReceiverQueueSize int
> > 
> > 
> > @shohi Can you remove the comments from `ReceiverQueueSize`?
> 
> Or change the description?

Yes.

pulsar/consumer_impl.go
shohi: unbuffered chan may be useful for some cases.

shohi: not sure whether it's a good idea to remove this option. If negative value is totally disallowed (should be reflected in comments), i think it's ok to panic though.

wolfstudy: This is related to the flow mechanism of messages in pulsar. By default, the value of the receiveQueueSize is 1000

shohi: Ok, i'll update the modification and comments according to your suggestions.

PR #292, opened 2020-06-21
Used pooled buffering for compression and batch serialization

merlimat:
### Motivation

 * Used pooled buffers when serializing a batch of messages
 * Perform in-place compression into the batch buffer

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -859,7 +859,7 @@ func (pc *partitionConsumer) Decompress(msgMeta *pb.MessageMetadata, payload int
 		pc.compressionProviders[msgMeta.GetCompression()] = provider
 	}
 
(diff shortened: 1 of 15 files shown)


pulsar/consumer_partition.go
```diff
@@ -859,7 +859,7 @@ func (pc *partitionConsumer) Decompress(msgMeta *pb.MessageMetadata, payload int
 		pc.compressionProviders[msgMeta.GetCompression()] = provider
 	}
 
-	uncompressed, err := provider.Decompress(payload.ReadableSlice(), int(msgMeta.GetUncompressedSize()))
+	uncompressed, err := provider.Decompress(nil, payload.ReadableSlice(), int(msgMeta.GetUncompressedSize()))
```
aahmed-se, line 862: have a method definition that does not require nil as a parameter.

merlimat: It’s an established convention in Go APIs. For example all the compression implementations are using that.

PR #299, opened 2020-06-26
Support producer name for Message

### LINKED ISSUE #295: Provide producer name
According to Pulsar documentation, a message should have a "producer name":
image

However, in this library there is no way to get the producer name upon reception of a message.

It'd be lovely to have. What I want to do here is to avoid to read my own messages, so I would check who is the producer of a message and if is the same name as me, then I'd do not do anything with that message.

wolfstudy:

Fixes #295 

### Motivation

Support producer name for Message interface, allows the user to get the `producerName` of the message in the received message.

```
if err != nil {
	log.Fatal(err)
}
	
log.Println(msg.ProducerName())
```

### Modifications

- add `producerName` filed to `message` struct


wolfstudy:
@merlimat Hello, currently, the Jenkinns CI still running, Do we need to do anything else for this?

wolfstudy:
@JoseFMP PTAL thanks

pulsar/message.go
```diff
@@ -64,6 +64,9 @@ type Message interface {
 	// Topic get the topic from which this message originated from
 	Topic() string
 
+	// ProducerName will get producer name from broker
```
merlimat, line 67: Saying "from broker" can be misleading since we're not fetching it from the broker, rather the name comes attached to the message metadata.

In Java client doc we have: 
```
Get the producer name who produced this message.
```

```suggestion
	// ProducerName returns the name of the producer that has published the message
```

PR #303, opened 2020-07-03
Close msg channel when consumer close

taiyang-li:
## Motivation
When consumer close,  message channel is not closed. It will cause the loop receiving messages by consumer.Chan() can't be break.
I fix it in this PR.

## Modifications
Add close message channel in consumer.Close()


wolfstudy:
@taiyang-li Can you merge master code to track the CI again?

pulsar/consumer_impl.go
```diff
@@ -374,6 +374,7 @@ func (c *consumer) Close() {
 		}
 		wg.Wait()
 		close(c.closeCh)
+		close(c.messageCh)
```
cckellogg, line 377: If the user passes in a channel and the channel is shared close could be called multiple times and there will be a panic.

taiyang-li: Yes. I already found this problem. It is hard to decide where to close the message channel.

taiyang-li: Maybe I should close message channel in multiTopicConsumer.Close() and  regexConsumer.Close(), and remove it in internal consumer.

taiyang-li: > Maybe I should close message channel in multiTopicConsumer.Close() and regexConsumer.Close(), and remove it in internal consumer.
 
Done. Commit again

pulsar/consumer_singletopic.go
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package pulsar
+
+type singleTopicConsumer struct {
```
cckellogg, line 20: What is the purpose of this?

pulsar/consumer_impl.go
cckellogg: If a users passes a message channel in the client should not close that channel.

cckellogg:
Looking at the code example reading messages from the channel after the consumer has unsubscribed or closed is kind of useless since you can't acknowledge those messages and they will be resent when the client connects again. I personally think it's better to have your own close flag to know when you should stop processing messages. 

Closing the message channel in the client is a little tricky since multiple go routines could be sending on it and the multi topic and regex consumer are built on the base consumer. There would probably need to be some sort of flag/lock to know when and who should close the message channel.

taiyang-li:
> Looking at the code example reading messages from the channel after the consumer has unsubscribed or closed is kind of useless since you can't acknowledge those messages and they will be resent when the client connects again. I personally think it's better to have your own close flag to know when you should stop processing messages.
> 
> Closing the message channel in the client is a little tricky since multiple go routines could be sending on it and the multi topic and regex consumer are built on the base consumer. There would probably need to be some sort of flag/lock to know when and who should close the message channel.

Indeed this is a workaround of the problem I mentioned above.  But new users must remember that he must break the consume loop himself, instead of client closing message channel. At least this make some confusion for me when I first use this library

pulsar/consumer_singletopic.go
taiyang-li: since I can't close message channel directly in internal consumer. So I wrap internal consumer for single topic subscription scene.

cckellogg: What happens if this is shared message channel passed in by the application?

taiyang-li: Usually,  it is better channel closed by producer, not consumer.  When application new a message channel and passes it to singleTopicConsumer/multiTopicConsumer/regexConsumer, we can still close the channel in interface Close() exactly once.

taiyang-li:
> @taiyang-li Can you merge master code to track the CI again?

Already merged.

pulsar/consumer_impl.go
flowchartsman: I just came across this issue, and allowing the user to pass in a message channel seems like a mistake to me, because it inverts control. If the user wants to multiplex messages from different consumers onto a channel they control, it should be up to them to handle that with a select clause, [runtime.Select](https://pkg.go.dev/reflect#Select), or a combination of concurrency primitives in their code.

RobertIndie: > I just came across this issue, and allowing the user to pass in a message channel seems like a mistake to me, because it inverts control. If the user wants to multiplex messages from different consumers onto a channel they control, it should be up to them to handle that with a select clause, [runtime.Select](https://pkg.go.dev/reflect#Select), or a combination of concurrency primitives in their code.

Yes, I agree that this is a bad idea for users to pass their message channel. We should remove it in the Go client 1.0.0. This feature makes it difficult to develop and fix the consumer.

PR #308, opened 2020-07-06
Remove MessageChannel from consumer options

wolfstudy:
### Motivation

It is a dangerous operation to expose the `MessageChannel` to the user, and the user may write data and other series of custom operations to the chan. This may interfere with `messageCh` correctly receiving data from the broker. We should recommend users to use `consumer.Chan()` instead of it, instead of directly exposing a readable and writable channel

### Modifications

remove MessageChannel from consumer options

### DIFF
--- a/examples/consumer-listener/consumer-listener.go
+++ b/examples/consumer-listener/consumer-listener.go
@@ -42,7 +42,6 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-
 	defer consumer.Close()
 
 	// Receive messages from channel. The channel returns a struct which contains message and the consumer from where
--- a/examples/consumer-listener/consumer-listener.go
+++ b/examples/consumer-listener/consumer-listener.go
@@ -32,15 +32,11 @@ func main() {
 
 	defer client.Close()
 
-	channel := make(chan pulsar.ConsumerMessage, 100)
-
 	options := pulsar.ConsumerOptions{
 		Topic:            "topic-1",
 		SubscriptionName: "my-subscription",
 		Type:             pulsar.Shared,
 	}
-
-	options.MessageChannel = channel
 
 	consumer, err := client.Subscribe(options)
 	if err != nil {
@@ -52,7 +48,7 @@ func main() {
 	// Receive messages from channel. The channel returns a struct which contains message and the consumer from where
 	// the message was received. It's not necessary here since we have 1 single consumer, but the channel could be
 	// shared across multiple consumers as well
-	for cm := range channel {
+	for cm := range consumer.Chan() {
 		msg := cm.Message
 		fmt.Printf("Received message  msgId: %v -- content: '%s'\n",
 			msg.ID(), string(msg.Payload()))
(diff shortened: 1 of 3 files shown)


wolfstudy:
@cckellogg PTAL thanks

cckellogg:
The idea was to give the option to limit the memory and or allow a shared channel when receiving messages. Would making the channel read only work? That would be an api breaking change though.

merlimat:
> It is a dangerous operation to expose the MessageChannel to the user, and the user may write data and other series of custom operations to the chan. This may interfere with messageCh correctly receiving data from the broker. 

Barring the API incompatibility, there are many ways with which an application can uses the API in the "wrong" way, impacting the proper behavior of the consumer. In all the cases, this would just impact the client itself, not the broker, so I don't think it's an issue.

> We should recommend users to use consumer.Chan() instead of it, instead of directly exposing a readable and writable channel

The `Chan()` method just returns the internal channel, so there would be nothing preventing me from writing into that channel, no?

wolfstudy:
> The Chan() method just returns the internal channel, so there would be nothing preventing me from writing into that channel, no?

The `MessageChannel` in consumer options is defined as a channel that can be read and written at the same time, and the channel returned in `consumer.Chan()` is a read-only channel, and the user can only read data from it.

```
MessageChannel chan ConsumerMessage
```
vs
```
Chan() <-chan ConsumerMessage
```

Yes, this is an action that will break API compatibility, but I still think that directly exposing a readable and writable channel is a dangerous action.  

> In all the cases, this would just impact the client itself, not the broker, so I don't think it's an issue.

Yes, Because we currently have a safer way to replace the behavior of the `MessageChannel` option. In other words, we should design to ensure that users are exposed to unsafe operations as little as possible, right?

wolfstudy:
Or, I want to know in terms of functions, which operations are consumer.Chan() can not replace MessageChannel. My understanding is that we expose a channel that can be read and written at the same time. In fact, we want users to read the data in this channel rather than write data to the channel.

cckellogg:
I think the message channel option should remain because there are valid use cases for it. The option needs to be read/write channel because the client needs to write messages. When a user passes a message channel only the client should be writing to that channel.  The documentation for the option is here

// Sets a `MessageChannel` for the consumer
// When a message is received, it will be pushed to the channel for consumption

It seems clear to me that you should not write to this channel. We could add another statement saying you should not try to write to this channel?


> I want to know in terms of functions, which operations are consumer.Chan() can not replace MessageChannel

I'm confused by this what is replacing the message channel?

wolfstudy:
> I think the message channel option should remain because there are valid use cases for it. The option needs to be read/write channel because the client needs to write messages. When a user passes a message channel only the client should be writing to that channel. 

Sorry, @cckellogg About this, i don't think so. In what kind of scenarios, the user needs to write a message to the `MessageChannel`? 

> I'm confused by this what is replacing the message channel?

As I said above, although we exposed MessageChannel, in more cases, we expect users to read data from this channel instead of writing data, so this parameter is exposed in consumer options Instead of producer's options.

> The option needs to be read/write channel because the client needs to write messages.

Yes, agree with you, but we can choose not to expose it to the outside world.

merlimat:
> Sorry, @cckellogg About this, i don't think so. In what kind of scenarios, the user needs to write a message to the MessageChannel?

If you are consuming from multiple consumers, you can aggregate all the messages into a single channel, supplied by the application.

PR #309, opened 2020-07-06
[Issue 306] Change connection failed warn log to error and print error message

### LINKED ISSUE #306: Should clear Error/Warn log
#### Expected behavior

Got detail error message when connection failed.

#### Actual behavior

Just got `WARN[0000] Failed to perform initial handshake - Expecting 'Connected' cmd, got 'ERROR'   local_addr="192.168.1.11:49262" remote_addr="pulsar://192.168.1.11:6560"`

#### Steps to reproduce

#### System configuration
**Pulsar version**: x.y

#### Other

connection.go#doHandshake
```
	if cmd.Connected == nil {
		c.log.Warnf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s' ",
			cmd.Type)
		return false
	}
```

Can  we use `cmd.GetError().GetMessage()`     instead of  `cmd.Type` ? just like :  

```
	if cmd.Connected == nil {
		c.log.Warnf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s' ",
			cmd.GetError().GetMessage())
		return false
	}
```

 and it can let me got `WARN[0000] Failed to perform initial handshake - Expecting 'Connected' cmd, got 'Unable to authenticate'  local_addr="192.168.1.11:50334" remote_addr="pulsar://192.168.1.11:6560"`. Maybe it should be `Error`   instead of  `Warn`.

liangyuanpeng:
Fixes #306 

### Motivation

### Modifications

Print a clean log when the connection failed.

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -287,8 +287,8 @@ func (c *connection) doHandshake() bool {
 	c.cnx.SetDeadline(time.Time{})
 
 	if cmd.Connected == nil {
-		c.log.Warnf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s'",
-			cmd.Type)
+		c.log.Errorf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s'",
+			cmd.Error.GetMessage())
 		return false
 	}
 	if cmd.Connected.MaxMessageSize != nil {
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -287,7 +287,7 @@ func (c *connection) doHandshake() bool {
 	c.cnx.SetDeadline(time.Time{})
 
 	if cmd.Connected == nil {
-		c.log.Errorf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s'",
+		c.log.Warnf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s'",
 			cmd.Error.GetMessage())
 		return false
 	}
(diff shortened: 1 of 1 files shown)


pulsar/internal/connection.go
```diff
@@ -287,8 +287,8 @@ func (c *connection) doHandshake() bool {
 	c.cnx.SetDeadline(time.Time{})
 
 	if cmd.Connected == nil {
-		c.log.Warnf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s'",
-			cmd.Type)
+		c.log.Errorf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s'",
```
wolfstudy, line 290: About log level I think `Warnf` is more appropriate, it has not reached the Error level

liangyuanpeng: How do we define the error level ? The client can't not working when this error happend and why just got warn log?

merlimat: The message itself need to be changed though. It was basically a "debug" message. In this case, the user doesn't know what's a `Connected cmd`.

pulsar/internal/connection.go
```diff
@@ -288,7 +288,7 @@ func (c *connection) doHandshake() bool {
 
 	if cmd.Connected == nil {
 		c.log.Warnf("Failed to perform initial handshake - Expecting 'Connected' cmd, got '%s'",
-			cmd.Type)
+			cmd.Error.GetMessage())
```
merlimat, line 291: Please change the original message since with this change it wouldn't make sense anymore:

```
c.log.Warnf("Failed to establish connection with broker: '%s'",
```

PR #310, opened 2020-07-06
Share buffer pool across all partitions

merlimat:
### Motivation

When a producer is publishing on many partitions, there can be significant memory overhead in maintaining a per-partition pool. Instead, there's not significant perf impact in using a single shared buffer pool.

### DIFF
--- a/perf/perf-producer.go
+++ b/perf/perf-producer.go
@@ -36,7 +36,7 @@ type ProduceArgs struct {
 	Topic              string
 	Rate               int
 	BatchingTimeMillis int
-	BatchingMaxSize int
+	BatchingMaxSize    int
 	MessageSize        int
 	ProducerQueueSize  int
 }
--- a/perf/perf-producer.go
+++ b/perf/perf-producer.go
@@ -36,6 +36,7 @@ type ProduceArgs struct {
 	Topic              string
 	Rate               int
 	BatchingTimeMillis int
+	BatchingMaxSize int
 	MessageSize        int
 	ProducerQueueSize  int
 }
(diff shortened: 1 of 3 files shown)


perf/perf-producer.go
```diff
@@ -62,6 +63,8 @@ func newProducerCommand() *cobra.Command {
 		"Publish rate. Set to 0 to go unthrottled")
 	flags.IntVarP(&produceArgs.BatchingTimeMillis, "batching-time", "b", 1,
 		"Batching grouping time in millis")
+	flags.IntVarP(&produceArgs.BatchingMaxSize, "batching-max-size", "", 128,
+		"Max size of a batch (in KB)")
```
wolfstudy, line 67: In `pulsar-perf produce` about `--batch-max-messages` is abbreviated and the default value is as follows, can I consider keeping the two consistent?


``` 
   -bm, --batch-max-messages
       Maximum number of messages per batch
       Default: 1000
```

merlimat: This are 2 different settings. One is the number of messages in 1 batch and the other is the max size of the batch

merlimat:
> just only some concerns whether the use of sync.pool here will cause GC pressure because sync.Pool cannot specify a size, which is only subject to the GC threshold.

Using the pool will not causing GC pressure itself: it's actually there to avoid the GC pressure. 

The pool having no max size is not a big problem. The max amount of memory is still determined by the "pending" messages whose payloads are buffered until they get acknowledged by the broker. 

This change is to use a single pool to avoid that each pool will have a few buffers that cannot be immediately reused. After this change, the memory usage is the same as without the pooling.

PR #313, opened 2020-07-08
Support oauth2 authentication for pulsar-client-go

zymap:
*Motivation*

Support oauth2 authentication for pulsar-client-go


pulsar/internal/auth/oauth2.go
```diff
@@ -0,0 +1,153 @@
...
+
+	oauth2Type := params[ConfigParamType]
+	keyFile := params[ConfigParamKeyFile]
+	return NewAuthenticationOAuth2(issuer, st, oauth2Type, keyFile)
+}
+
+func NewAuthenticationOAuth2(
+	issuer oauth2.Issuer,
+	store store.Store,
+	oauth2Type, keyFilePath string) (Provider, error) {
```
EronWright, line 65: I see that you changed this constructor from `NewAuthenticationOAuth2(issuer, store)` to `NewAuthenticationOAuth2(issuer, store, oauth2type, keyfile)`.   Let me explain why I think that's not a good change.

The authentication provider requires an authorization grant to be able to obtain access tokens over time.  The grant is a long-lived credential, such as a keyfile or a refresh token.  In all cases, the +application+ obtains the grant beforehand and then stores it in a store.  The store might be persistent.  Later, the application creates an authorization provider with the store as a parameter.  The provider loads the grant from the store, and uses it over time to obtain access tokens.

I am trying to explain that the provider is not responsible for obtaining or storing new grants.  Hence the signature `NewAuthenticationOAuth2(issuer, store)`. 

Now we discuss the special case of initializing the provider from parameters - `NewAuthenticationOAuth2WithParams`.  One could imagine a combination of parameters that would open a pre-existing store but this would require more design work.  A simple scenario that is possible now is the keyfile scenario in combination with an in-memory store (where the access token would be discarded when the process terminates).  This scenario is implemented above.

Would you revert the change to this method that is seen in https://github.com/apache/pulsar-client-go/pull/313/commits/97d98f753995fc7879e69472efec6fb698ebed99?

go.mod
```diff
@@ -1,20 +1,25 @@
...
 	github.com/pierrec/lz4 v2.0.5+incompatible
-	github.com/pkg/errors v0.8.1
+	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.4.1
 	github.com/spaolacci/murmur3 v1.1.0
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3 // indirect
 	github.com/stretchr/testify v1.4.0
 	github.com/yahoo/athenz v1.8.55
+	k8s.io/utils v0.0.0-20200619165400-6e3d28b6ed19
```
EronWright, line 22: This dependency was due to the mock clock right?  I recall that you forked it elsewhere, would that work here too?

zymap: Yes. Let me replace it.

PR #314, opened 2020-07-08
[Issue 52]Add interceptor

snowcrumble:
### Motivation

Add A chain of interceptors for Producer and Consumer as an option, these interceptors will be called at some points, it can be used for tracing, metrics, and so on.

### Modifications

* Add two files for interceptor definition.
* Call interceptor's methods at appropriate position.

### Verifying this change

- Added unit tests for all interceptor's methods.

  - The public API: yes
  - The schema: no
  - The wire protocol: no

### Documentation


pulsar/consumer_test.go
```diff
@@ -1342,3 +1344,130 @@ func TestProducerName(t *testing.T) {
...
+func TestConsumerWithInterceptors(t *testing.T) {
+	client, err := NewClient(ClientOptions{
+		URL: lookupURL,
+	})
+
+	assert.Nil(t, err)
+	defer client.Close()
+
+	rand.Seed(time.Now().Unix())
+	topic := fmt.Sprintf("persistent://public/default/test-topic-interceptors-%d", rand.Int())
```
wolfstudy, line 1400: Maybe we can use the `newTopicName()` to create a random topic.

pulsar/consumer_test.go
```diff
@@ -1342,3 +1344,130 @@ func TestProducerName(t *testing.T) {
...
+			nackIds = append(nackIds, msg.ID())
+		}
+	}
+	assert.Equal(t, int32(5), atomic.LoadInt32(&metric.ackn))
+
+	for i := range nackIds {
+		consumer.NackID(nackIds[i])
+	}
+
+	time.Sleep(time.Second * 3) // waiting for nack actual perform
```
wolfstudy, line 1471: Please avoid using sleep in the test, because the status of each server is uncertain.

wolfstudy:
@snowcrumble Can you merge the master code and fix conflict?

PR #319, opened 2020-07-09
Do not allocate MessageIDs on the heap

dferstay:
Passing a function parameter by pointer (or writing a pointer into a
channel) will cause the Go escape analysis to allocate the parameter on
the heap.

This change passes messageID struct instances by value instead of by
pointer; this keeps messageID structs on the stack.

Each message produced or consumed by the library is associated with
a MessageID; keeping instances of the MessageID structure on the stack
reduces heap allocation and associated GC cost.

### Motivation

Reduce the amount of per-Message heap allocation performed by the library

### Modifications

This change modifies the Consumer and Producer code paths to pass messageID struct instances by value instead of by
pointer; this keeps the messageID structures on the stack and off of the heap.

### Verifying this change

This change is already covered by existing tests, such as:
* pulsar/consumer_multitopic_test.go
* pulsar/consumer_partition_test.go
* pulsar/consumer_regex_test.go
* pulsar/consumer_test.go
* pulsar/impl_message_test.go
* pulsar/negative_acks_tracker_test.go
* pulsar/producer_test.go
* pulsar/reader_test.go

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -36,8 +36,8 @@ var ErrConsumerClosed = errors.New("consumer closed")
 const defaultNackRedeliveryDelay = 1 * time.Minute
 
 type acker interface {
-	AckID(id *messageID)
-	NackID(id *messageID)
+	AckID(id messageID)
+	NackID(id messageID)
 }
 
 type consumer struct {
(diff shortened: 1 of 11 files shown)


dferstay:
Below is the alloc_space view of a pprof heap profile collected during a large scale test of a high-throughput service that uses pulsar-client-go to send messages to Pulsar.

message-id-alloc-space

PR #321, opened 2020-07-10
Add Name method to Consumer interface

shohi:
## Motivation

Supports the `Name()` interface on the consumer side, allowing users to customize their own consumer name. If the user does not specify, a default consumer name will be randomly generated

## Modifications

- add `consumerName` field in consumer struct
- add `consumerName` field in multiConsumer struct
- add `consumerName` field in regexConsumer struct


wolfstudy:
Thanks @shohi work for this, the change LGTM +1, I added some descriptive information about this pull request, PTAL

Issue #228, opened 2020-04-17
ClientOptions should get a Logger field to allow injecting of an external Logger


cornelk:
It is available in the old cgo based go client library:

```
	// Provide a custom logger implementation where all Pulsar library info/warn/error messages will be routed
	// By default, log messages will be printed on standard output. By passing a logger function, application
	// can determine how to print logs. This function will be called each time the Pulsar client library wants
	// to write any logs.
	Logger func(level log.LoggerLevel, file string, line int, message string)
```

merlimat:
@cornelk On the CGo library, this was done to plug into a go logger into the C++ library logger.

In the new library, we're using Logrus for logging. You can easily configure Logrus (https://github.com/sirupsen/logrus)

For example: https://github.com/apache/pulsar-client-go/blob/master/perf/pulsar-perf-go.go#L53

The problem with the logger func, is that we'd be losing the structured logging information.

cornelk:
This works but requires the users of the library to import the logger library that is used and set the required log levels and formatter in an `init()` function compared to passing a logger to the library, which would be the better pattern here.

Also please consider using a different logger library while this library is still young, for example [zap](https://github.com/uber-go/zap), which is faster, does less memory allocations and is even recommended by logrus itself: `Logrus would look like those, had it been re-designed with what we know about structured logging in Go today. Check out, for example, Zerolog, Zap, and Apex.`

jonyhy96:
Considering give the ability of config the logger to user may be useful if you don't want the user chose there own logger implements.

WIZARDISHUNGRY:
It may be preferable to allow injection of an interface as a logger ([like this](https://github.com/dexidp/dex/blob/master/pkg/log/logger.go)) but we'd lose the structured logging features. We use logrus most places right now, so even allowing us to inject our own logrus would help as a first step.

taiyang-li:
Pulsar-client-go use default logger `std` allaround. 

In most cases, this will work , because users are able to `SetOutput`  `SetFormatter`  `SetLevel`  (refer to https://github.com/sirupsen/logrus/blob/master/exported.go)  

But if users already use default logrus logger `std` before introduce the dependency of pulsar-client-go,  and they want to set different output/formatter/level for these 2 cases,  there would be a problem.

So I think it is better to use a  diffrent default logrus logger(not `std`) for pulsar-go-client, which will provide much more flexibility of logging for users

bschofield:
[...]

PR #322, opened 2020-07-10
support set logger of pulsar

taiyang-li:
From Issue   https://github.com/apache/pulsar-client-go/issues/228  

Modifications:
pulsar-client-go new a default logger instead of use `std`.

### DIFF
--- a/examples/consumer-listener/consumer-listener.go
+++ b/examples/consumer-listener/consumer-listener.go
@@ -22,7 +22,19 @@ import (
 	"log"
 
 	"github.com/apache/pulsar-client-go/pulsar"
+	"github.com/natefinch/lumberjack"
 )
+
+func init() {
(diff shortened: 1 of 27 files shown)


merlimat:
I'm not sure how is this different than applications just using Logrus APIs to configure the logging?

taiyang-li:
Pulsar-client-go use default logger std allaround.

In most cases, this will work , because users are able to SetOutput SetFormatter SetLevel (refer to https://github.com/sirupsen/logrus/blob/master/exported.go)

But if users already use default logrus logger std before introduce the dependency of pulsar-client-go, and they want to set output/formatter/level different from pulsar-client-go does, there would be a problem.

So I think it is better to use a diffrent default logrus logger(not std) for pulsar-go-client, which will provide much more flexibility of logging for users

PR #323, opened 2020-07-13
define logger interface and add Logger field to ClientOptions

shohi:
### Motivation
enable users to configure the logger used by the client and use their own implementation. If no logger is provided, a wrapped `logrus.StandardLogger()` will be used. <s>This PR only solved part of the problem mentioned in the issue https://github.com/apache/pulsar-client-go/issues/228.</s>

### Modifications

* define `Logger` and `Entry` interfaces used by the client
* add `Logger` field to ClientOptions
* add `logger` field to internal structures
* provide a logger implementation backed by logrus
* implement a no-op logger


wolfstudy:
@shohi It seems that you and #322  are solving the same thing, right?

shohi:
@wolfstudy Not the same approach. This PR enables users to pass a given logger when creating clients, while https://github.com/apache/pulsar-client-go/pull/322 adds methods to alter the global logger.

pulsar/client.go
```diff
@@ -91,6 +92,9 @@ type ClientOptions struct {
 
 	// Max number of connections to a single broker that will kept in the pool. (Default: 1 connection)
 	MaxConnectionsPerBroker int
+
+	// Configure the logger used by the client. (Default: logrus.StandardLogger())
+	Logger *logrus.Logger
```
merlimat, line 97: I like this approach, though in this case we'd be exposing Logrus as part of our API, something that we cannot later rollback. 

If we want to take a logger option in, we should have a custom logger interface (with default impl bound to logrus).

shohi: Yeah, i agree with you. It may need some effort to extract a reasonable logger interface, how about resolving it in another PR?

merlimat: My main concern is that once we expose in API, people can depend on it. So we should rather first define the logger API and then apply this change.

wsoh: Agreed, will add logger and entry interface later.

wsoh:
@merlimat `Logger` and `Entry` interfaces are defined in `log` package, plz take a took :-)

wsoh:
also implemented a logrus backed logger and no-op logger

pulsar/client.go
shohi: it makes sense, i defined the `Logger` and `Entry` interfaces at `pulsar/log` package, please take a look :-)

pulsar/log/log.go
```diff
@@ -0,0 +1,51 @@
...
+func (l nopLogger) WithField(name string, value interface{}) Entry { return nopEntry{} }
+func (l nopLogger) Debug(args ...interface{})                      {}
+func (l nopLogger) Info(args ...interface{})                       {}
+func (l nopLogger) Warn(args ...interface{})                       {}
+func (l nopLogger) Error(args ...interface{})                      {}
+func (l nopLogger) Debugf(format string, args ...interface{})      {}
+func (l nopLogger) Infof(format string, args ...interface{})       {}
+func (l nopLogger) Warnf(format string, args ...interface{})       {}
+func (l nopLogger) Errorf(format string, args ...interface{})      {}
+
```
snowcrumble, line 38: why no Panic and Fatal?

pulsar/log/wrapper_logrus.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
+// logrusWrapper implements Logger interface
+// based on underlying logrus.FieldLogger
+type logrusWrapper struct {
+	l logrus.FieldLogger
+}
```
snowcrumble, line 28: If interface Logger's method is a sub set of `*logrus.Logger`, you can directly new a `logrus.Logger` as interface `Logger` to be the default logger without define this wrapper.

pulsar/log/log.go
shohi: To make the interfaces concise and small, I don't think exiting and panicing are in the scope of a logging library. (inspired by [logur/logur](https://github.com/logur/logur/blob/master/logger.go#L39)

pulsar/log/wrapper_logrus.go
shohi: thanks for your suggestion :-). Actually, the method set of interfaces is not a subset of logrus.Logger
*  the logger/entry interfaces defined in pulsar-client-go are independent of any existing logger implementations.
* the logger has its own `Fields` type even though it's similar with `logrus`, but they are not the same.

pulsar/log/logger.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Fields type, used to pass to `WithFields`.
+type Fields map[string]interface{}
+
+// Logger describes the interface that must be implemeted by all loggers.
+type Logger interface {
+	SubLogger(fields Fields) Logger
+
+	WithFields(fields Fields) Entry
+	WithField(name string, value interface{}) Entry
```
ernado, line 30: IMO there should be also `WithError(err) Entry` method, `WithField("cause", err)`  is not 1:1 replacement and more error-prone (e.g. somebody can make a typo field name).

pulsar/producer_partition.go
```diff
@@ -89,7 +88,7 @@ type partitionProducer struct {
 	state  int32
 	client *client
 	topic  string
-	log    *log.Entry
+	logger log.Logger
```
ernado, line 91: Looks like excessive rename. `log` field is OK too.

pulsar/producer_impl.go
```diff
@@ -61,7 +60,7 @@ type producer struct {
 	ticker        *time.Ticker
 	tickerStop    chan struct{}
 
-	log *log.Entry
```
ernado, line 64: Same excessive field rename. Can you please reduce review scope? ^^

pulsar/log/logger.go
```diff
@@ -0,0 +1,57 @@
...
+	Error(args ...interface{})
+
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Warnf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+}
+
+// Entry describes the interface for the logger entry.
+type Entry interface {
```
ernado, line 44: Is it possible to leave one interface? E.g. `SubLogger(fields Fields) Logger` is semantically close to  `WithFields(fields Fields) Entry`, so probably we can leave just `Entry`?

pulsar/producer_partition.go
shohi: `log` is usually used for the package name, the logger is the one who does the logging job, i think it's more accurate using `logger` than `log`

pulsar/log/logger.go
shohi: Yeah, `Logger` and `Entry` share many methods, but their responsibilities are different. In general, `Entry` just represents one log entry, once fired it should be released. Separating entry from logger enables users to fine-grained control the implementation, such as pooling the entries. I'm not sure whether it's more friendly and easier to use by combining the two interfaces into one.

pulsar/log/logger.go
shohi: Make sense, it will be handy and less error-prone if the `WithError` is available for pulsar-client-go, although the logging features are already complete.

pulsar/producer_impl.go
shohi: as mentioned above, i think using `logger` as field name is more idiomatic.

pulsar/producer_partition.go
ernado: It's true, but it makes review scope unfortunately bigger :(

pulsar/log/logger.go
ernado: Currently there is no pooling, so having two entities just complicates implementation and slightly confuses.
The `SubLogger(fields Fields)` result is technically `logrus.Entry`.

pulsar/producer_partition.go
ernado: `log` were selected initially by package maintainers, IMO there is no reason to change naming in scope of this PR.
It is really hard to review.

pulsar/internal/commands.go
```diff
@@ -196,20 +194,21 @@ func baseCommand(cmdType pb.BaseCommand_Type, msg proto.Message) *pb.BaseCommand
...
-func addSingleMessageToBatch(wb Buffer, smm *pb.SingleMessageMetadata, payload []byte) {
+func addSingleMessageToBatch(wb Buffer, smm *pb.SingleMessageMetadata, payload []byte, logger log.Logger) {
 	metadataSize := uint32(smm.Size())
 	wb.WriteUint32(metadataSize)
 
 	wb.ResizeIfNeeded(metadataSize)
 	_, err := smm.MarshalToSizedBuffer(wb.WritableSlice()[:metadataSize])
 	if err != nil {
-		log.WithError(err).Fatal("Protobuf serialization error")
+		logger.WithError(err).Error("Protobuf serialization error")
```
ernado, line 210: Can we leave `Fatal` here? It is different from Error + panic.

pulsar/producer_partition.go
shohi: Yeah, renaming makes the review scope bigger, but i don't think it's hard to review, it's just a little annoying. The field name using `log` really confuses me, perhaps it's better to open another PR for naming refactor, how about putting a TODO marker first?

pulsar/log/logger.go
shohi: For `logrus` backed implementation, it's true, but the logger interfaces are not targeted only for logrus, they are more for users to customize the implementation to cope with their own scenarios. 

Using two entities makes logging concepts clear and gives users more control on the implementation, in some cases,  logging performance is critical and entry pooling is necessary. Logger is usually used to hold the log level, common fields and its own entry pool, where the entry holds its private fields, formats all its fields into bytes and emits the bytes out.  I think implementation based on these concepts is much straightforward. The popular `logrus` provides both logger-level and entry-level methods, but they are not based on interfaces. Another the performant logging library `zap` provides only Logger level methods, but all loggers share a global entry pool.

shohi:
> @shohi , please take a look at #354
> 
>     1. Reduced interface scope to single `Logger`
> 
>     2. Allow `logrus.FieldLogger` as `NewLoggerWithLogrus` argument (this will allow passing logrus entries as new Logger so user can populate needed fields)
> 
>     3. Use `log` instead of `logger` field name.
> 
> 
> IMO we can rename `log` to `logger` in separate PR to reduce merge conflicts and PR scope.

shohi:
> @shohi , please take a look at #354
> 
>     1. Reduced interface scope to single `Logger`
> 
>     2. Allow `logrus.FieldLogger` as `NewLoggerWithLogrus` argument (this will allow passing logrus entries as new Logger so user can populate needed fields)
> 
>     3. Use `log` instead of `logger` field name.
> 
> 
> IMO we can rename `log` to `logger` in separate PR to reduce merge conflicts and PR scope.

I have some concerns about point 1 - see my comments above, and agree with the others.

pulsar/producer_partition.go
ernado: TODO marker is ok

pulsar/log/logger.go
ernado: Agree, but it seems like logrus approach with fields (`map[string]interface{}`) will reduce benefits from pooling.

Moreover, attempt to abstract logging is already limits performance optimizations due to usage of interfaces instead of structures, so I'm not sure that having both interfaces will simplify something performance-related.

There is another point: we are pre-v1, so it is OK to break backward compatibility at some point. If one decide to switch to zap or zerolog and employ it's performance benefits, they will be forced to change interfaces (but I still think that it is nearly impossible to create both abstracted and performant logger interfaces)

pulsar/internal/connection.go
```diff
@@ -188,21 +188,30 @@ type connection struct {
...
-	connectionTimeout time.Duration, auth auth.Provider) *connection {
+// ConnectionOptions defines configurations for creating connection.
+type ConnectionOptions struct {
+	LogicalAddr       *url.URL
+	PhysicalAddr      *url.URL
+	TLS               *TLSOptions
+	ConnectionTimeout time.Duration
+	Auth              auth.Provider
+	Logger            log.Logger
+}
```
wolfstudy, line 199: If `ConnectionOptions` is only used as a parameter of `newConnection()`, can we consider it as a private?

wolfstudy (approved):
Cool, LGTM +1

wolfstudy:
@shohi Can you help provide the docs in `docs` folder? How users can use this feature to customize their logs.

shohi:
Sure, will update it

pulsar/internal/connection.go
shohi: Sure, will update it

shohi:
The logger interfaces are much straightforward - almost same approach applied by logrus or zap, users can refer these implementations for some hints.

wolfstudy:
@shohi Can you help merge master code and fix conflict?

wsoh:
> @shohi Can you help merge master code and fix conflict?

Sure, i'll apply the code change from the master branch, but not sure I have the permission to merge codes

shohi:
@wolfstudy Sure, i'll apply the code change from the master branch, but not sure i have the permission to merge codes.

shohi:
@wolfstudy the codes have been updated, but i don't have the write access to the repo, could you please help merge the PR?

wolfstudy:
Thanks @shohi work for this

PR #326, opened 2020-07-17
Fix producer block

LvBay:
When the client creates a producer and a consumer with the same id, then the consumer processing handleCloseConsumer() will cause the producer with the same id to be unable to handle handleSendReceipt() and handleCloseProducer() in the future. The most direct impact is the Send() method Will always block.


LvBay:
@merlimat @wolfstudy 
Please look look~

PR #337, opened 2020-07-29
Avoid producer deadlock on connection closing

merlimat:
### Motivation

There's a condition in which a producer can remain deadlocked in the event of a connection failure. 

The sequence goes like: 
 1. Producer (or multiple producers) have several outstanding request to write on a connection
 2. The channel (used to pass buffers to write) has the buffer full and thus it's blocking
 3. When the connection is closed the channel is not drained
 4. The connection tries to notify the producers that it's time to reconnect
 5. The producer go-routine is not able to process that notification, since it's blocked on the connection channel

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -416,8 +416,10 @@ func (c *connection) WriteData(data Buffer) {
 			// The channel is either:
 			// 1. blocked, in which case we need to wait until we have space
 			// 2. the connection is already closed, then we need to bail out
+			c.log.Debug("Couldn't write on connection channel immediately")
 			state := connectionState(atomic.LoadInt32(&c.state))
 			if state != connectionReady {
+				c.log.Debug("Connection was already closed")
 				return
 			}
 		}
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -107,14 +107,14 @@ type ConsumerHandler interface {
 	ConnectionClosed()
 }
 
-type connectionState int
+type connectionState int32
 
 const (
-	connectionInit connectionState = iota
-	connectionConnecting
-	connectionTCPConnected
-	connectionReady
-	connectionClosed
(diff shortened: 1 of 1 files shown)


pulsar/internal/connection.go
```diff
@@ -729,7 +756,7 @@ func (c *connection) Close() {
 
 func (c *connection) changeState(state connectionState) {
 	c.Lock()
-	c.state = state
+	atomic.StoreInt32(&c.state, int32(state))
```
wolfstudy, line 759: In here, Is it necessary for us to `c.Lock()` and `c.Unlock()` actions on an atomic operation?

merlimat: I think it's still needed to ensure that the mutex condition is broadcasted. To trigger a condition you need to have a lock on the associated mutex.

wolfstudy: I am wondering if there are other problems with this approach, because for an atomic primitive, we rarely see the operation of Lock and Unlock it.


```
atomic.StoreInt32(&c.state, int32(state))
c.Lock()
c.cond.Broadcast()
c.Unlock()
```

merlimat: There are 2 different usages here: 
 * On one hand we want other go routines to be able to check the state, without taking a lock
 * On the other hand, we still need to maintain the atomic state update and notification

wolfstudy: yes, i agree with you, The main difference between us here is whether we need to lock to protect the atomic primitive, I think it is not needed, atomic itself is a synchronization primitive, so here, we can reduce the granularity of the lock and only lock sync.cond. The code example is as follows:

```
atomic.StoreInt32(&c.state, int32(state))
c.Lock()
c.cond.Broadcast()
c.Unlock()
```

wolfstudy: Please ignore me

PR #341, opened 2020-08-03
double check

yarthur1:
### Motivation

 To avoid delete other goroutine established connection, connectionPool's GetConnection function may be better by using double check


wolfstudy:
It seems that there is no need to go to double check here, you can try the latest master code to see if there is still connection competition.

I close this pr first, if there are any problems, we can reopen this pr for discussion

PR #343, opened 2020-08-04
Support specified the oauth2 private key with prefix 'file://' and 'data://'

zymap:
---

*Motivation*

Make the oauth2 read the private key can handle with  'file://' schema and 'data://' schema.


zymap:
I added the new test case here: https://github.com/apache/pulsar-client-go/pull/343/files#diff-7e18f94aa1cb5bdae4886ab8052f0c14R107. Also I added the `data://` schema test case.

PR #359, opened 2020-08-24
Support retry letter topic 

### LINKED ISSUE #353: Support retry logic in dead letter topic
For many online business systems, a message needs to be re-consumed because any exception occurs in the business logic processing.

More information reference to http://pulsar.apache.org/docs/en/concepts-messaging/#retry-letter-topic

wuYin:
Fixes #353 

### Motivation

Follow [pulsar#6449](https://github.com/apache/pulsar/pull/6449) to support retry letter topic in go client

### Modifications

- Add `retryRouter` for sending reconsume messages to retry letter topic
- Add `ReconsumeLater(msg Message, delay time.Duration)` to Consumer interface
- Add configureable retry letter topic name in `DLQPolicy`
    ```go
	type DLQPolicy struct {
		// ...
		// Name of the topic where the retry messages will be sent.
		RetryLetterTopic string
	}
	```
    enable it explicitly while creating consumer, default unenable
   
     ```go
    type ConsumerOptions struct {
	    // ...
		// Auto retry send messages to default filled DLQPolicy topics
		RetryEnable bool
	}
    ```
- Add 2 `TestRLQ*`  test cases

### Verifying this change

### Documentation
[pulsar doc: retry-letter-topic](http://pulsar.apache.org/docs/en/concepts-messaging/#retry-letter-topic)

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -176,6 +176,7 @@ type connection struct {
 	closeCh            chan interface{}
 	writeRequestsCh    chan Buffer
 
+	pendingLock sync.Mutex
 	pendingReqs map[uint64]*request
 	listeners   map[uint64]ConnectionListener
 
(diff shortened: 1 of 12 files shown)


wuYin:
run integration tests

pulsar/internal/connection.go
```diff
@@ -356,24 +357,35 @@ func (c *connection) run() {
...
+
+			case req := <-c.incomingRequestsCh:
+				if req == nil {
+					return // TODO: this never gonna be happen
+				}
+				c.internalSendRequest(req)
+			}
+		}
+	}()
+
```
wolfstudy, line 383: @wuYin Why do we need to do this?

wolfstudy (commented):
Thanks @wuYin work for this, the change LGTM, just one comment, PTAL.

pulsar/internal/connection.go
wuYin: Background:
- Retry topic producer, User topic consumer, they two use same connection instance to interact with broker.
- Request channel, Response channel, they two are listened in a [single `select{}` loop](https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/connection.go#L367) by a connection.

When message need to be redelivery:
1. Retry topic producer called `SendAsync()`, wrapped User topic consumer AckID logic as **callback** into message handler.
2. connection send cmd `SEND` to broker.
3. Broker send cmd  `SEND_RECEIPT` to connection.
4. connection received and call  **callback**
5. **callback** called `AckID()`, then connection want to send cmd `ACK`  to broker.
in the meantime, `select{}` was stucked in step 3 and  executing `SEND_RECEIPT` handler, `incomingRequestsCh` never be chosen, when it fulled with 10 Ack requests, finnally lead to dead block.

Asynchronized reuquest & response channels in connection, this block will be resolved.


deadblock

wolfstudy: Cool, LGTM +1

wolfstudy (approved):

PR #360, opened 2020-08-25
[Issue 257][feat] Support limit the retry number of reconnectToBroker

### LINKED ISSUE #257: Retry number limit in reconnectToBroker
Once the connection is closed the reconnectToBroker logic of both producer and consumer will try to reconnect to the broker infinatly.

Add a field in the `Options` represent the max number of retries and add a default number.

The client will stop retry reconnect to the broker under a limit number  of retries.

**Additional context**
I can make a pr of this If we all agree with this is a needed feature

jonyhy96:
Fixes #257

### Motivation

Once the connection is closed the reconnectToBroker logic of both producer and consumer will try to reconnect to the broker infinatly.

### Modifications

Add a field in the Options represent the max number of retries and not break current behavior if this field is not fullfilled.

### Verifying this change

  - The public API: (no)
  - The schema: (yes)
  - The wire protocol: (no)

### Documentation


pulsar/consumer.go
```diff
@@ -140,6 +140,9 @@ type ConsumerOptions struct {
 
 	// A chain of interceptors, These interceptors will be called at some points defined in ConsumerInterceptor interface.
 	Interceptors ConsumerInterceptors
+
+	// MaxReconnectToBroker set the maximum retry number of reconnectToBroker. (default: ultimate)
+	MaxReconnectToBroker *uint
```
shohi, line 145: IMO using `int` type is more friendly,  the default value `0` represents unlimited.

pulsar/producer_partition.go
```diff
@@ -236,8 +236,18 @@ func (p *partitionProducer) ConnectionClosed() {
...
+	var (
+		maxRetry int
+		backoff  = internal.Backoff{}
+	)
+
+	if p.options.MaxReconnectToBroker == nil {
+		maxRetry = -1
+	} else {
+		maxRetry = int(*p.options.MaxReconnectToBroker)
+	}
```
shohi, line 248: The handling for uninitialized maxRetry seems not correct, it may be better to extract the retry test into a separate function, e.g `shouldRetry`.

pulsar/consumer.go
jonyhy96: > IMO using `int` type is more friendly, the default value `0` represents unlimited.

How aboud `0` represents no retry but nil means retry ultimate times? Then this changes won't break current behavior of reconnectToBroker.

pulsar/producer_partition.go
jonyhy96: > The handling for uninitialized maxRetry seems not correct, it may be better to extract the retry test into a separate function, e.g `shouldRetry`.

```go
func (p *partitionProducer) limitRetry() bool {
 return p.options.MaxReconnectToBroker != nil
}

...
 var maxRetry = -1
 if p.limitRetry() {
   maxRetry = int(*p.options.MaxReconnectToBroker)  
 }
...
}
```

Do you mean something like this?

shohi: my bad, the handling logic is correct :-) 

What i mean is to wrap the test logic, like following

```go
func (p *partitionProducer) shouldRetry() bool {
	maxReconnects := p.opts.MaxReconnectToBroker

	switch {
	case maxReconnects < 0:
		return true
	case maxReconnects == 0:
		return false
	default:
		p.reconnectAttempts++
		if p.reconnectAttempts > maxReconnects {
			return false
		}

		return true
	}
}

	for p.shouldRetry() {
		// retry logic
	}
}
```

pulsar/consumer.go
shohi: I see your point, but it's common to use int for this kind of configuration, such as [go-redis](https://github.com/go-redis/redis/blob/master/options.go#L59) / [nats.go](https://github.com/nats-io/nats.go/blob/master/nats.go#L266). 

* 0 for no-try
* negative number for retry forever
* positive number for retry attempts

As current library is still in pre-1.0, i think it's ok to make a break change.

pulsar/producer_partition.go
jonyhy96: Great, This looks more clear. 
I think it would be better if we decide the type of p.opts.MaxReconnectToBroker first.
Besides, i think `shouldRetry()` logic maybe a little bit slower than the origin one, we can reduce the `shouldRetry()` call times by move some common logic outside of the for loop.

pulsar/consumer.go
jonyhy96: > I see your point, but it's common to use int for this kind of configuration, such as [go-redis](https://github.com/go-redis/redis/blob/master/options.go#L59) / [nats.go](https://github.com/nats-io/nats.go/blob/master/nats.go#L266), we can make the following convention for the config to cope with different scenarios,
> 
> * 0 for no retry
> * negative number for retry forever
> * positive number to set retry attempts
> 
> As the current library is still in pre-1.0, i think it's ok to make a break change.

Totally agree, Do we need to ask for other maintainer's opinions?

shohi: invite @wolfstudy to review?

jonyhy96: > invite @wolfstudy to review?

Sounds great!

wolfstudy:
Thanks @jonyhy96 Here, I have only one confusion. When users use it, they need to set it on the producer and consumer at the same time, right? Can we consider exposing this parameter to the user in ClientOptions so that the user only needs to set it once

jonyhy96:
> Thanks @jonyhy96 Here, I have only one confusion. When users use it, they need to set it on the producer and consumer at the same time, right? Can we consider exposing this parameter to the user in ClientOptions so that the user only needs to set it once

That would be more convenience for users to use, But the user loses more detailed control over the program.
By the way,do you mean add an option at [here](https://github.com/apache/pulsar-client-go/blob/master/pulsar/client.go#L77) and [here](https://github.com/apache/pulsar-client-go/blob/master/pulsar/client_impl.go#L40),in reconnectToBroker use p.clinet.maxReconnectToBroker to decide the retry logic?

jonyhy96:
> > Thanks @jonyhy96 Here, I have only one confusion. When users use it, they need to set it on the producer and consumer at the same time, right? Can we consider exposing this parameter to the user in ClientOptions so that the user only needs to set it once
> 
> That would be more convenience for users to use, But the user loses more detailed control over the program.
> By the way,do you mean add an option at [here](https://github.com/apache/pulsar-client-go/blob/master/pulsar/client.go#L77) and [here](https://github.com/apache/pulsar-client-go/blob/master/pulsar/client_impl.go#L40),in reconnectToBroker use p.clinet.maxReconnectToBroker to decide the retry logic?

/cc @wolfstudy @shohi  

Any further information about this? Maybe we can implement this feature before octuber 😄

pulsar/consumer.go
wolfstudy: Thanks @jonyhy96,  @shohi proposal LGTM +, can you help fix this?

wolfstudy (approved):
LGTM +1, just a little comment, @jonyhy96 please help check, thanks.

jonyhy96:
> LGTM +1, just a little comment, @jonyhy96 please help check, thanks.

Thanks for review! Already fix this.

pulsar/producer.go
```diff
@@ -138,6 +138,9 @@ type ProducerOptions struct {
 
 	// A chain of interceptors, These interceptors will be called at some points defined in ProducerInterceptor interface
 	Interceptors ProducerInterceptors
+
+	// MaxReconnectToBroker set the maximum retry number of reconnectToBroker. (default: no retry)
+	MaxReconnectToBroker int
```
wolfstudy, line 143: @jonyhy96 Can we not modify the original default behavior, and we will always retry by default. When we set the maximum number of retries, we will turn off the retry logic after reaching the specified threshold.

Because the current behavior will break the backward compatibility of the version

wolfstudy (changes requested):

pulsar/producer.go
jonyhy96: Sure, neither do I want to break the current behavior. The original commit of this pr will fulfill our request, please check out the [comments](https://github.com/apache/pulsar-client-go/pull/360#discussion_r476231877) above. @shohi may have different opinion on this due to his(her) proposal. Please take a review.

wolfstudy: We can consider setting -1 as the default value, and then check whether MaxReconnectToBroker is set, and if so, use the user-defined value.

wolfstudy: At present, I see that you have added the option of MaxReconnectToBroker to the relevant test cases. This may not be very user-friendly. We need to ensure that the interface is as simple as possible.

jonyhy96: > We can consider setting -1 as the default value, and then check whether MaxReconnectToBroker is set, and if so, use the user-defined value.

I understand,but how can we determine whether MaxReconnectToBroker is set or not when it's default value is meaningful. e.g. user set this option to 0.

> At present, I see that you have added the option of MaxReconnectToBroker to the relevant test cases. This may not be very user-friendly. We need to ensure that the interface is as simple as possible.

I totally agree, how about we use the origin commit of this pr to implement this feature?

wolfstudy: ok, agree with you

jonyhy96: Should i reset this pr to original commit or keep it this way?

wolfstudy: @jonyhy96 Sorry for the reply later, i think we should reset this pull request to the original commit for backward compatibility.

jonyhy96: > @jonyhy96 Sorry for the reply later, i think we should reset this pull request to the original commit for backward compatibility.

Already reset this pr to original commit.
By the way, the go.mod file seems a little bit dirty,you can reproduce this by run comamnd `go mod tidy` locally.

```shell
➜  pulsar-client-go git:(feat-max-reconnect-to-broker) go mod tidy -v        
unused github.com/modern-go/concurrent
unused github.com/modern-go/reflect2
```

perhaps add script below to ci for check if go.mod has been commit correctly

```shell
go mod tidy
git diff --exit-code -- go.mod go.sum
```

wolfstudy (approved):

PR #69, opened 2019-10-01
[ISSUE #68][feat]add Option config for init (#68)

xujianhai666:
Change-Id: I85a9c9f20e61e126b617eab919d2405a3ebda087

Master Issue: #68 

### Motivation

simple initialization of Producer

### Modifications

add ProducerOption for newProducer of client

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (yes / no)
  - The public API: (**yes** / no)
  - The schema: (yes / no / don't know)
  - The default values of configurations: (yes / no)
  - The wire protocol: (yes / no)

### Documentation

  - Does this pull request introduce a new feature? (yes / no)
  - If yes, how is the feature documented? (not applicable / docs / GoDocs / not documented)
  - If a feature is not applicable for documentation, explain why?
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

PR #362, opened 2020-09-08
[Producer]Add functional options

lhui1991:
### Motivation

*Using the Functional Options Pattern to make producer constructor more elegant and explicit.*

### Verifying this change

  - The public API: yes  (CreateProducer)
  - The schema:no
  - The default values of configurations:  no
  - The wire protocol:  no

### Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ client, err := pulsar.NewClient(pulsar.ClientOptions{
 
 defer client.Close()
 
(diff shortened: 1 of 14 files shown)


lhui1991:
> There was a similar PR (#69) a while back. I'm not sure this adds anything and the names would have to be prefixed to avoid clashing.

okay，Looks like I did the same thing.I will close this.

PR #368, opened 2020-09-21
Add schema logic in producer and consumer

### LINKED ISSUE #344: Support Schema Registry feature in Go Client
This is the Schema Registry design doc For Go Client: https://gist.github.com/wolfstudy/e0c9b349cd405715a9e04dd8658510bc

- [x] Basic classes on supporting schema (Primitive, PROTO, JSON, AVRO) #368
- [x] Support schema on Producer @wolfstudy #368
- [x] Support schema on Consumer @wolfstudy #368
- [ ] Add the schema version support - Multiple schema on a single topic (*) 
- [ ] Support schema on Reader
- [ ] Advanced Schema
   - [ ] KeyValue Schema
   - [ ] AutoSchema

wolfstudy:

Fixes #344 

### Motivation

Type safety is extremely important in any application built around a message bus like Pulsar. Currently, Apache Pulsar supports the function of Schema Registry. And Java, CPP and Python clients already support schema registry related functions, In order to further improve the function of Go Client, we need to support the function of Schema Registry in Go Client.

### Modifications

- Add schema logic in producer
- Add schema logic in consumer
- Package basic tool functions for schema logic

---

### Currente state

- Encapsulates all basic types of decode and encode logic
- Encapsulates the use of all schema types
- A new interface has been added for producer and consumer to be compatible with schema functions


wolfstudy:
@merlimat @cckellogg PTAL thanks.

pulsar/schemaDef_test.go
```diff
@@ -0,0 +1,57 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
cckellogg, line 1: rename file schema_test.go. We should use snake_case to be consistent with the rest of the files. Same for other camelCase files

pulsar/schemaDef_test.go
```diff
@@ -0,0 +1,57 @@
...
+package pulsar
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSchemaDef(t *testing.T) {
+	errSchemaDef := "{\"type\":\"record\",\"name\":\"Example\",\"namespace\":\"test\"," +
```
cckellogg, line 27: Would ` marks make these easier to maintain? We would then not have to scape the " characters

pulsar/schema.go
```diff
@@ -0,0 +1,502 @@
...
+type Int16Schema struct {
+	SchemaInfo
+}
+
+func NewInt16Schema(properties map[string]string) *Int16Schema {
+	int16Schema := new(Int16Schema)
+	int16Schema.SchemaInfo.Properties = properties
+	int16Schema.SchemaInfo.Name = "INT16"
+	int16Schema.SchemaInfo.Type = INT16
+	int16Schema.SchemaInfo.Schema = ""
```
cckellogg, line 332: Why is the schema "" for all primitives?

pulsar/primitiveSerDe.go
```diff
@@ -0,0 +1,316 @@
...
+const (
+	IoMaxSize     = 1024
+	maxBorrowSize = 10
+)
+
+var (
+	littleEndian = binary.LittleEndian
+)
+
+type BinaryFreeList chan []byte
```
cckellogg, line 36: What's the purpose of this?

pulsar/schema.go
```diff
@@ -0,0 +1,502 @@
...
+	"reflect"
+	"unsafe"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/linkedin/goavro"
+)
+
+type SchemaType int
```
cckellogg, line 33: Should we add a String() method on the enum then for the names we can call JSON.String(), etc... and it's useful for logging and or debugging

pulsar/message.go
```diff
@@ -27,6 +27,9 @@ type ProducerMessage struct {
 	// Payload for the message
 	Payload []byte
 
+	//Value and payload is mutually exclusive, `Value interface{}` for schema message.
+	Value interface{}
```
cckellogg, line 31: Should this be package private?

pulsar/schemaDef_test.go
wolfstudy: Yes, this test case is just to test whether the logic of the schemaDef is correct. When we implement the schema function on the producer and consumer respectively, we will add separate test cases for each schema type. At this time, the file will be named `schema_test.go`.

pulsar/message.go
wolfstudy: Should not be private, this interface is to be exposed to users, for example:

```
err := producer.Send(context.Background(), ProducerMessage{
		Value: &testJson{
			ID:   100,
			Name: "pulsar",
		},
	})
```

pulsar/schemaDef_test.go
wolfstudy: Good ideas will fix it

pulsar/schema.go
wolfstudy: Maybe we can use SchemaInfo.Name replace it, right?

pulsar/schema.go
wolfstudy: For primitives types, the schema field should be `""`

pulsar/schemaDef_test.go
cckellogg: we should be consistent with name. all the files are snake_case and I suggest we keep it that way for consistency. schema_def_test.go. Same for the other files in the PR.

cckellogg (commented):
I think this PR should be split into 2. First add the structs/interfaces for the schema and then add a follow up PR to implement the schema in the client. This will keep the PRs more self contained and easier to review.

skyrocknroll:
@wolfstudy is there any specific reason we are using v1.0.5 of goavro. According to README `"github.com/linkedin/goavro/v2"` has better support for namespaces and performance improvements. Can we please switch to v2 ?

wolfstudy:
> @wolfstudy is there any specific reason why we are using v1.0.5 of goavro ? According to README `"github.com/linkedin/goavro/v2"` has better support for namespaces and performance improvements. Can we please switch to v2 ?

Cool, I think we can update the `gpavro` to `v2`

wolfstudy:
@skyrocknroll done PTAL again, thanks. 

ping @sijie @massakam If time is convenient, please review it, thanks.

jiazhai (approved):
lgtm

PR #371, opened 2020-09-27
Support ack timeout

wuYin:
### Motivation
Follow doc [acknowledgement-timeout](https://pulsar.apache.org/docs/en/concepts-messaging/#acknowledgement-timeout) to support ack timeout feature.

### Modifications

- Add `AckTimeout` to ConsumerOptions.
- Add `unackedMessageTracker` to track messages which received by user.
- Add 5 `TestAckTimeout*` test cases.

### Verifying this change

- `TestAckTimeout*`
    Track ack timeout in single topic consumers, with `Exclusive, Failover, Shared` subscription.
- `TestAckTimeoutSharedMultiAndRegexTopics` 
    Track ack timeout in multi topics consumers, with `Shared` subscription.

### Different default behavior
- Java Client: if consumer enabled DLQPolicy, added 30s `ackTimeout` by default, see [apache/pulsar #3104](https://github.com/apache/pulsar/pull/3014)
- Go Client: I think it's no longer necessary to do this, so ignored.

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -132,6 +132,9 @@ type ConsumerOptions struct {
 	// Set the consumer name.
 	Name string
 
+	// Timeout of unacked messages.
+	AckTimeout time.Duration
+
 	// If enabled, the consumer will read messages from the compacted topic rather than reading the full message backlog
 	// of the topic. This means that, if the topic has been compacted, the consumer will only see the latest value for
 	// each key in the topic, up until the point in the topic message backlog that has been compacted. Beyond that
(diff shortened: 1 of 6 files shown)


wolfstudy:
Thanks @wuYin support for this logic, in https://github.com/apache/pulsar-client-go/pull/197, we tried to support this feature, but after discussion, we decided to discard this feature. In many cases, how long the ack timeout is set is a problem, so we recommend users to use nack instead.

wolfstudy (changes requested):
Negative ack are the preferred way to solve that problem and ack timeout should be regarded as legacy.

wuYin:
Thank you for reply, make sense to me, I'll close this PR.

uestczshen:
if network is disconnected, NACK cannot reaches the pulsar server. Then how does NACK  work?  so maybe ack-timeout can work in this situation.

YorkHwang:
@wuYin 
Go Client: I think it's no longer necessary to do this, so ignored. why？

PR #376, opened 2020-10-01
Fix deadlock when connection closed

jiazhai:
Fixes #366

### Motivation

In current code of `pulsar/internal/connection.go` we have 2 channels, closeCh and incomingRequestsCh. when the connection closes, the current mis-use of these 2 channels may have a deadlock. 
PR #366 has detailed steps to reproduce and the root cause [analysis](https://github.com/apache/pulsar-client-go/pull/366#issuecomment-696759873) .
This PR tries to fix the deadlock.

### Modifications
- make the close logic independent, not in the same loop of normal events handling.
- when the connection closed, handle the existing requests in the channel and return an error to avoid deadlock.

### Verifying this change
passed the tests in #366 
current ut passed

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,8 @@ require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/klauspost/compress v1.10.8
 	github.com/kr/pretty v0.2.0 // indirect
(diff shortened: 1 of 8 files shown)


wolfstudy (approved):
Cool, LGTM+1

jiazhai:
Thanks @wolfstudy for the help

PR #394, opened 2020-11-18
Support sendTimeout

wuYin:
### Motivation

Support `SendTimeout` and `DisableBlockIfQueueFull ` for producer.

### Modifications

- Add 2 configurations into `ProducerOptions`
- Check and fail timeout messages before flushing batch.
- Check `DisableBlockIfQueueFull` before acquiring publish semaphore while sending messages.
- Add `TestSendTimeout` to verify sendTimeout effective whille backlog exceeded on topic.

### Others
client-go and client-java have different ways to disable `sendTimeout`
- client-java: set `sendTimeout` to 0 explicitly, to override default 30s configuration.
- client-go: set `sendTimeout` to -1, to bypass unset detection.

### Documentation
[configure-producer](https://pulsar.apache.org/docs/en/client-libraries-java/#configure-producer)

### Verifying this change

### DIFF
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -78,7 +78,7 @@ type ProducerOptions struct {
 
 	// SendTimeout set the timeout for a message that not be acknowledged by server since sent.
 	// Send and SendAsync returns an error after timeout.
-	// Default is 30 seconds, -1 to disable.
+	// Default is 30 seconds, negative such as -1 to disable.
 	SendTimeout time.Duration
 
 	// DisableBlockIfQueueFull control whether Send and SendAsync block if producer's message queue is full.
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -81,9 +81,9 @@ type ProducerOptions struct {
 	// Default is 30 seconds, -1 to disable.
 	SendTimeout time.Duration
 
-	// BlockIfQueueFull control whether Send and SendAsync return error if producer's message queue is full.
-	// Default is false.
-	BlockIfQueueFull bool
+	// NonBlockIfQueueFull control whether Send and SendAsync block if producer's message queue is full.
+	// Default is false, if set to true then Send and SendAsync return error when queue is full.
+	NonBlockIfQueueFull bool
 
 	// MaxPendingMessages set the max size of the queue holding the messages pending to receive an
 	// acknowledgment from the broker.
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -76,6 +76,15 @@ type ProducerOptions struct {
 	// This properties will be visible in the topic stats
 	Properties map[string]string
 
+	// SendTimeout set the timeout for a message that not be acknowledged by server since sent.
+	// Send and SendAsync returns an error after timeout.
+	// Default is 30 seconds, -1 to disable.
+	SendTimeout time.Duration
+
+	// BlockIfQueueFull control whether Send and SendAsync return error if producer's message queue is full.
+	// Default is false.
+	BlockIfQueueFull bool
+
 	// MaxPendingMessages set the max size of the queue holding the messages pending to receive an
 	// acknowledgment from the broker.
 	MaxPendingMessages int
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -81,9 +81,9 @@ type ProducerOptions struct {
 	// Default is 30 seconds, -1 to disable.
 	SendTimeout time.Duration
 
-	// NonBlockIfQueueFull control whether Send and SendAsync block if producer's message queue is full.
+	// DisableBlockIfQueueFull control whether Send and SendAsync block if producer's message queue is full.
 	// Default is false, if set to true then Send and SendAsync return error when queue is full.
-	NonBlockIfQueueFull bool
+	DisableBlockIfQueueFull bool
 
 	// MaxPendingMessages set the max size of the queue holding the messages pending to receive an
 	// acknowledgment from the broker.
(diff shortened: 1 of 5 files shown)


pulsar/producer.go
```diff
@@ -76,6 +76,15 @@ type ProducerOptions struct {
 	// This properties will be visible in the topic stats
 	Properties map[string]string
 
+	// SendTimeout set the timeout for a message that not be acknowledged by server since sent.
+	// Send and SendAsync returns an error after timeout.
+	// Default is 30 seconds, -1 to disable.
+	SendTimeout time.Duration
```
merlimat, line 82: Shouldn't this be taken the `ctx` passed in by the application at the moment of `Send()` ?

pulsar/producer.go
```diff
@@ -76,6 +76,15 @@ type ProducerOptions struct {
...
+
+	// BlockIfQueueFull control whether Send and SendAsync return error if producer's message queue is full.
+	// Default is false.
+	BlockIfQueueFull bool
```
merlimat, line 86: The logic should be negated here. The default should be to block if queue full.

wuYin: Thanks for review.

I intented to maintain the same default behavior as [client-java](https://pulsar.apache.org/docs/en/client-libraries-java/#configure-producer), but this does change the default behavior of current client-go.

How about renaming `BlockIfQueueFull` to `NonBlockIfQueueFull`? just like `DisableBatching` for enable batch by default.

I think the reason for doing this is, we couldn't check whether the user sets `false` explicitly or default `false`

pulsar/producer.go
wuYin: Thanks for review.
I see [PR#252](https://github.com/apache/pulsar-client-go/pull/252) implementing `sendTimeout` with context but progress has stalled.
I'm going to try context instead of timer.

pulsar/producer_partition.go
```diff
@@ -434,17 +445,79 @@ func (p *partitionProducer) internalFlushCurrentBatch() {
...
+		}
+		t.Reset(p.options.SendTimeout)
+		p.queueLock.Unlock()
+	}
+}
+
 func (p *partitionProducer) internalFlush(fr *flushRequest) {
 	p.internalFlushCurrentBatch()
 
+	p.queueLock.Lock()
```
cckellogg, line 519: Instead of having a separate go routine checking for timeouts can the message expiry check be performed here?

I don't think the timeout check needs to be exact? We just need to expires old messages at some point? This could simplify the code and reduces the need for another lock. We can just call this method `failTimeoutMessages` and expire messages if needed.

pulsar/producer.go
```diff
@@ -76,6 +76,15 @@ type ProducerOptions struct {
...
 	Properties map[string]string
 
+	// SendTimeout set the timeout for a message that not be acknowledged by server since sent.
+	// Send and SendAsync returns an error after timeout.
+	// Default is 30 seconds, -1 to disable.
+	SendTimeout time.Duration
+
+	// NonBlockIfQueueFull control whether Send and SendAsync block if producer's message queue is full.
+	// Default is false, if set to true then Send and SendAsync return error when queue is full.
+	NonBlockIfQueueFull bool
```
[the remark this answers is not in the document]
wuYin: Thanks for review.
Now I renamed NonBlockIfQueueFull to DisableBlockIfQueueFull.

pulsar/producer_partition.go
wuYin: Thanks for review, good idea.
In some case `eventLoop` will be stucked such as producer is reconnecting to broker, 2 ways:
- Check at independent goroutine: `sendTimeout` still effective, periodic check and fail pending messages.
- Check before flush: `sendTimeout` not effective in this case.

If no need to check message timeout precisely, the above case will be ignored.
This is reasonable, since sendTimeout now implemented at **producer level**, no need to be that precise, also avoid the cost of lock.

pulsar/producer_partition.go
```diff
@@ -553,9 +613,7 @@ func (p *partitionProducer) ReceivedSendReceipt(response *pb.CommandSendReceipt)
 	// lock the pending item while sending the requests
 	pi.Lock()
 	defer pi.Unlock()
-	if pi.sentAt > 0 {
```
wuYin, line 556: @cckellogg 
`sentAt` always set before puting into pending queue, so I removed this predicate.
Like [sendRequest.publishTime](https://github.com/apache/pulsar-client-go/pull/394/files#diff-8adb09af2175be5751840343e14df2220a2154ca4bc8de21157ed112be25b6b9L565) did

pulsar/producer_partition.go
```diff
@@ -430,26 +432,74 @@ type pendingItem struct {
...
 	batchData    internal.Buffer
 	sequenceID   uint64
-	sentAt       int64
+	sentAt       time.Time
 	sendRequests []interface{}
 	completed    bool
 }
 
 func (p *partitionProducer) internalFlushCurrentBatch() {
+	if p.options.SendTimeout > 0 {
```
cckellogg, line 441: How can this be disabled? Setting it to -1?

cckellogg (commented):
Looks ok to me

merlimat (approved):

pulsar/producer_partition.go
wuYin: Yes, negative to disable, such as -1 typically.
https://github.com/apache/pulsar-client-go/blob/master/pulsar/producer.go#L81

PR #399, opened 2020-11-20
[Issue 398][consumer] only update the permits when we try to receive messages

### LINKED ISSUE #398: ReceiverQueueSize of 0 doesn't work as intended
(This was already reported against the old driver here (by me): https://github.com/apache/pulsar/issues/6496 but it was noted this was fixed with the new go implementation, which doesn't seem the case. The case is identical.)

#### Expected behavior

When running pulsar as a queue system, with long/varying processing times per message, a client setting ReceiverQueueSize to 0 should allow a client to only request a new message when it's ready for new work.

When specifying a ReceiverQueueSize of 0, no messages should be pre-fetched, and the client should only request a message when it's ready with the previous message. This is also how it's documented at https://pulsar.apache.org/docs/en/cookbooks-message-queue/#client-configuration-changes.

I also tried a python implementation, and that actually works as it should be.

#### Actual behavior

Setting a ReceiverQueueSize of 0 sets the defaultQueueSize of 1000, which is not intended.

#### Steps to reproduce

Consumer and producer code below. Run 3 consumers: two at a 1 second sleeptime and one at a 10 second sleep. When injecting 10 messages, it should give you 1 message on the 10 second consumer, and 4 and 5 on both 1 second sleeptime consumers.

How I did it: Run a standalone docker pulsar, and open 4 terminal windows. In 2, `go run consumer.go`, in 1 `go run consumer.go --sleeptime=10` and once they're all connected, run `go run producer.go` in the last window.

Simplified output:

Expected:
- consumer_10_seconds: consumes message 0
- consumer_1_second: consumes messages 1,3,5,7,9
- consumer_1_second: consumes messages 2,4,6,8
total runtime: 10 seconds

Actual:
- consumer_10_seconds: consumes message 0,5
- consumer_1_second: consumes messages 1,3,7,9
- consumer_1_second: consumes messages 2,4,6,8
total runtime: 20 seconds

#### System configuration
**Pulsar version**: 2.6.2 (docker, standalone)
Latest pulsar-client-go driver (v0.3.0)

#### Things I already tried

I fiddled a bit in the pulsar-client-go, hoping to resolve this. 

Made sure that it accepts 0 by changing the check here: https://github.com/apache/pulsar-client-go/blob/master/pulsar/consumer_impl.go#L94-L96, and made sure that the initialPermits was at least 1. This didn't yield the expected result.

#### Code: 

Consumer:

```
package main

import (
	"context"
	"flag"
	"fmt"
	"log"
	"time"

	"github.com/apache/pulsar-client-go/pulsar"
)

func main() {
	var sleeptime int

	flag.IntVar(&sleeptime, "sleeptime", 1, "sleeptime")
	flag.Parse()

	client, err := pulsar.NewClient(pulsar.ClientOptions{URL: "pulsar://localhost:6650"})
	if err != nil {
		log.Fatal(err)
	}

	defer client.Close()

		Topic:             "testtopic",
		SubscriptionName:  "sharedtest",
		Type:              pulsar.Shared,
		ReceiverQueueSize: 0,
	})
	if err != nil {
		log.Fatal(err)
	}

	defer consumer.Close()

	for {
		if err != nil {
			log.Fatal(err)
		}

		fmt.Printf("Received message  msgId: %s -- content: '%s' - will now sleep for %d seconds\n", msg.ID(), string(msg.Payload()), sleeptime)

		time.Sleep(time.Duration(sleeptime) * time.Second)
		consumer.Ack(msg)
	}
}

```

Producer:

```
package main

import (
	"context"
	"fmt"

	log "github.com/apache/pulsar/pulsar-client-go/logutil"
	"github.com/apache/pulsar/pulsar-client-go/pulsar"
)

func main() {
	fmt.Printf("Started...")
		URL: "pulsar://localhost:6650",
	})

	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("got Client")

	defer client.Close()

		Topic: "testtopic",
	})
	if err != nil {
		log.Fatal(err)
	}

	defer producer.Close()

	ctx := context.Background()

	fmt.Printf("Ready for sending")
	for i := 0; i < 10; i++ {
		msgID, err := producer.SendAndGetMsgID(ctx, pulsar.ProducerMessage{
		})

		if err != nil {
			log.Fatal(err)
		}
		fmt.Printf("The message Id value is: [%v] \n", msgID)
	}
}

```

For reference: the python code that works: 

```
import pulsar
import time

client = pulsar.Client('pulsar://localhost:6650')
consumer = client.subscribe('persistent://public/default/testtopic', 'sharedtest', consumer_type=pulsar.ConsumerType.Shared, receiver_queue_size=0)
while True:
    msg = consumer.receive()
    print(msg.data())
    time.sleep(1)
    consumer.acknowledge(msg)
client.close()
```

(change the sleeptime in one run to 10)

wonko:
Fixes #398 

### Motivation

The client didn't handle a receiverqueuesize of 0 correctly. See the attached issue for the in-depth explanation of the issue. 

### Modifications

The problem lies in the fact that, while the message is popped from the internal messageChannel towards the client code, we should hold back on receiving the next message. 

I was able to "solve" this for my case by moving the logic to ask for new/more permits away from the passing of the message, and into the logic that fetched the next message, eventually delaying the update of the permits.

I'm pretty sure this isn't the best fix possible, but it highlights the problem area...

### Verifying this change

Sorry, no tests added, i'm just not good enough at golang :D

No

### Documentation

No

### DIFF
--- a/pulsar/client_impl.go
+++ b/pulsar/client_impl.go
@@ -52,7 +52,8 @@ func newClient(options ClientOptions) (Client, error) {
 	if options.Logger != nil {
 		logger = options.Logger
 	} else {
-		logger = log.NewLoggerWithLogrus(logrus.StandardLogger())
+		l := logrus.StandardLogger()
+		logger = log.NewLoggerWithLogrus(l)
 	}
 
 	if options.URL == "" {
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -91,7 +91,7 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		return nil, newError(SubscriptionNotFound, "subscription name is required for consumer")
 	}
 
-	if options.ReceiverQueueSize <= 0 {
+	if options.ReceiverQueueSize < 0 {
 		options.ReceiverQueueSize = defaultReceiverQueueSize
 	}
 
@@ -391,6 +391,10 @@ func (c *consumer) Unsubscribe() error {
 }
 
 func (c *consumer) Receive(ctx context.Context) (message Message, err error) {
+	for _, consumer := range c.consumers {
+		consumer.nudgePermits()
+	}
+
 	for {
 		select {
 		case <-c.closeCh:
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -179,7 +179,7 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 		partitionIdx:         int32(options.partitionIdx),
 		eventsCh:             make(chan interface{}, 10),
 		queueSize:            int32(options.receiverQueueSize),
-		queueCh:              make(chan []*message, options.receiverQueueSize),
+		queueCh:              make(chan []*message, 5),
 		startMessageID:       options.startMessageID,
 		connectedCh:          make(chan struct{}),
 		messageCh:            messageCh,
@@ -644,6 +644,9 @@ func (pc *partitionConsumer) dispatcher() {
 			// reset available permits
 			pc.availablePermits = 0
 			initialPermits := uint32(pc.queueSize)
+			if initialPermits <= 0 {
+				initialPermits = 1
+			}
 
 			pc.log.Debugf("dispatcher requesting initial permits=%d", initialPermits)
 			// send initial permits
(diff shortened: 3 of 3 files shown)


merlimat:
There are few issues with this PR: 
 1. It's changing the default receiver queue size to 0 for everyone
 2. The receiveQueueSize=0 in Java and C++ is a kind of special case that needs to be treated in different way. Also it doesn't work consuming either partitioned topics or multiple topics in general (since we cannot know which topic we want to consume from).

wonko:
I'm well aware this PR isn't up to spec, it's more a proof that the problem lies with the driver and how it handles the case of ReceiverQueueSize of 0. These changes made my test-case behave as it is documented (and how I need it).

I'm by no means a good enough golang developer, so the decent change is beyond my reach, tbh.

Feel free to close this PR.

wonko:
Closing this as I won't be able to get this PR up to spec but I would love to have the golang driver in line with the Java/C drivers, so they all behave the same.

PR #402, opened 2020-11-30
Add VPC（Virtual Private Cloud）support for go sdk

xixiss:
### Motivation
Add listenerName field in ClientOptions to support VPC（Virtual Private Cloud）net mode in cloud environment when create client.Then sdk can use vpc mode when lookup for topics.

### Modifications
- add "listenerName" in ClientOptions
- Add corresponding fields when lookup


wolfstudy (changes requested):
Thanks @xixiss work for this, can you merge master code and fix the Action CI? Please make sure the CI is OK.

wolfstudy:
@xixiss Move this change to 0.5.0

pulsar/internal/lookup_service_test.go
```diff
@@ -117,9 +117,10 @@ func TestLookupSuccess(t *testing.T) {
 
 		expectedRequests: []pb.CommandLookupTopic{
 			{
-				RequestId:     proto.Uint64(1),
-				Topic:         proto.String("my-topic"),
-				Authoritative: proto.Bool(false),
+				RequestId:              proto.Uint64(1),
+				Topic:                  proto.String("my-topic"),
+				Authoritative:          proto.Bool(false),
+				AdvertisedListenerName: proto.String(""),
```
wolfstudy, line 123: @xixiss Can you add new test case for this change? This test case is used to cover other scenarios, please do not verify our own logic in this test case.

PR #406, opened 2020-12-01
Add const client label to metrics

cckellogg:
Add const client label to metrics so when metrics are added to other clients we can filter the metrics by client language.

### DIFF
--- a/pulsar/internal/metrics.go
+++ b/pulsar/internal/metrics.go
@@ -17,12 +17,8 @@
 
 package internal
 
-import (
-	"fmt"
-)
-
 func constLabels() map[string]string {
 	return map[string]string{
-		"client": fmt.Sprintf("go"),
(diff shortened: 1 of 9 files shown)


pulsar/internal/metrics.go
```diff
@@ -0,0 +1,24 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package internal
+
+func constLabels() map[string]string {
+	return map[string]string{
+		"client": "go",
```
cckellogg, line 22: Eventually it might be good to add the client tag/commit sha too.

merlimat (approved):

PR #408, opened 2020-12-02
Read the clock fewer times during message routing

dferstay:
Previously, we read the system clock twice for each event unless we
switched partitions early due to reaching `maxBatchingMessages` or
`maxBatchingSize`.

Now, we read the clock once every `maxBatchingMessages / 100` messages.

This improves performance (especially for larger batch sizes) as the
router function is called for every message produced.

A bench test of the default router was added; results are below:
```
name             old time/op    new time/op    delta
DefaultRouter       106ns ± 0%      16ns ± 1%  -84.94%  (p=0.016 n=4+5)
DefaultRouter-4     106ns ± 0%      16ns ± 0%     ~     (p=0.079 n=4+5)
```

### Motivation

The default_router reads the system clock twice for every message routed in the production code path.  When testing at scale with larger batch sizes the amount of time spent reading the system clock during message routing shows up in pprof profiles.

### Modifications

This change modifies the default router to read the system clock once for every `maxBatchingMessages / 100` messages routed.
A bench test of the default router was added to verify that the change yields a speedup.

### Verifying this change

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/default_router.go
+++ b/pulsar/default_router.go
@@ -48,6 +48,7 @@ func NewDefaultRouter(
 		lastChangeTimestamp:    math.MinInt64,
 	}
 
+	readClockAfterNumMessages := uint32(maxBatchingMessages / 10)
 	return func(message *ProducerMessage, numPartitions uint32) int {
 		if numPartitions == 1 {
 			// When there are no partitions, don't even bother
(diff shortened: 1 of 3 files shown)


merlimat (approved):

wolfstudy:
@dferstay Can you fix the ci error?

wolfstudy (approved):

PR #428, opened 2020-12-22
support jwt and trusted cert for pulsar perf client

zzzming:
### Motivation

Add token/JWT and trusted certificate support in Pulsar perf go client. So that it can be used in production environment.

### Modifications

Implement pulsar-perf-go command line arguments for token-file and trusted-cert-file. These two arguments are optional. There is no change to the default non-TLS and no auth configuration.

### Verifying this change

This is a change to Pulsar perf go client. It should pass build and existing test cases.

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/perf/pulsar-perf-go.go
+++ b/perf/pulsar-perf-go.go
@@ -26,7 +26,6 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
-	"strings"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


perf/pulsar-perf-go.go
```diff
@@ -49,6 +53,23 @@ func NewClient() (pulsar.Client, error) {
...
+		tokenBytes, err := ioutil.ReadFile(clientArgs.TokenFile)
+		if err != nil {
+			log.WithError(err).Errorf("failed to read Pulsar JWT from a file %s", clientArgs.TokenFile)
+			os.Exit(1)
+		}
+		clientOpts.Authentication = pulsar.NewAuthenticationToken(string(tokenBytes))
+	}
+
+	if strings.HasPrefix(clientArgs.ServiceURL, "pulsar+ssl://") {
+		if clientArgs.TLSTrustCertFile == "" {
```
cckellogg, line 68: If the broker certificate is signed by a known verifier is the trust cert still needed?

zzzming: If the certificate is from trusted certificate authorities, you can use the default set of certificate authorities in OS environment. But the default CA bundle still needs to be set against TLSTrustCertFile.

But if mandatory requirement of a trust file path when `pulsar+ssl://` is detected from serviceURL concerns you, I'll update the code to remove this check, because Golang will spit out `failed to parse root CAs certificates` when root ca path is missing anyhow.

cckellogg (approved):

wolfstudy (approved):

PR #436, opened 2021-01-07
Add a bridge type to allow usage of zap logging by the pulsar client

bschofield:
### Motivation

This client library has pluggable logging (via https://github.com/apache/pulsar-client-go/pull/323) and a bridge to the popular _logrus_ library. Currently, there is no bridge to the (also popular) _zap_ logging library.

This PR adds a bridge to _zap_.

### Usage

Assume you have set up your logger at program init time, e.g.

    logger, err = zap.NewProduction()
    if err != nil {
        log.Fatalf("failed to set up zap: %v", err)
    }

    zap.ReplaceGlobals(logger)

Then, when creating a client, add the _zap_ sugared logger like this:

    client := pulsar.NewClient(pulsar.ClientOptions{
        URL:    pulsarURL,
        Logger: log.NewLoggerWithZap(zap.S()),
    })

where `log` is the package `github.com/apache/pulsar-client-go/pulsar/log`.

### Modifications

Added a new `zapWrapper` type to the `log` package.

This does introduce a dependency on _zap_. If that is undesirable, it might be simpler to put the `wrapper_zap.go` code (which is quite short) in the documentation somewhere. However, you already have a dependency on _logrus_, so an additional dependency on another very common logging library doesn't seem so bad to me.

### Verifying this change

  - Dependencies (does it add or upgrade a dependency): yes, dependency on `go.uber.org/zap`
  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation


bschofield:
It does feel a little weird to include _zap_ as a dependency of this library, when the library doesn't use it directly. I wonder if another option might be to include an empty `go.mod` in the `log` package directory, so that _zap_ is not added to the top-level `go.mod` (https://github.com/golang/go/issues/30058#issuecomment-543815369). Users of _zap_ will already have it included in their own modules file.

PR #450, opened 2021-01-25
Logging what really caused lookup failure

wuYin:
### Motivation

Clusters deployed over k8s, during broker pod restarting / terminating, it's service DNS `.cluster.local` will be removed, cause proxy can't resolve the restarting owner broker DNS name.
For now, client only logging `ServiceNotReady` error but without details which help troubleshoot.

### Modifications

Logging LookupResponse failure detail message, like this:

image

### Verifying this change

### DIFF
--- a/pulsar/internal/lookup_service.go
+++ b/pulsar/internal/lookup_service.go
@@ -19,7 +19,6 @@ package internal
 
 import (
 	"errors"
-	"fmt"
 	"net/url"
 
 	"github.com/gogo/protobuf/proto"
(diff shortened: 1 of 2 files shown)


pulsar/internal/lookup_service.go
```diff
@@ -140,12 +140,9 @@ func (ls *lookupService) Lookup(topic string) (*LookupResult, error) {
 			}, nil
 
 		case pb.CommandLookupTopicResponse_Failed:
-			errorMsg := ""
-			if lr.Error != nil {
-				errorMsg = lr.Error.String()
-			}
-			ls.log.Warnf("Failed to lookup topic: %s, error msg: %s", topic, errorMsg)
-			return nil, fmt.Errorf("failed to lookup topic: %s", errorMsg)
+			err := fmt.Errorf("failed to lookup topic: %s, error: %s, message: %s", topic, lr.GetError(), lr.GetMessage())
```
merlimat, line 143: We should preferably use the `WithField("error", lr.getGetError())` instead of formatting into a string

wuYin: Thanks for review, I'll follow this convention
Now I removed string formatting, message fields will be highlight in log and only print once

merlimat (approved):

Issue #448, opened 2021-01-25
Access to partitionConsumer state not thread-safe.


flowchartsman:
As part of my testing of the fix for #432, I have discovered a race condition in consumer_partition.go where access to the struct's `state` member is modified concurrently without synchronization:

```
WARNING: DATA RACE
Read at 0x00c00039a918 by goroutine 9:
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).clearQueueAndGetNextMessage()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:911 +0x68
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).clearReceiverQueue()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:932 +0x84
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).grabConn()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:861 +0x1312
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).reconnectToBroker()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:802 +0x238
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).runEventsLoop.func2()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:711 +0xab

Previous write at 0x00c00039a918 by goroutine 70:
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).internalClose()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:772 +0x68f
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).runEventsLoop()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:732 +0x29b

Goroutine 9 (running) created at:
  github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).runEventsLoop()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:704 +0x146

Goroutine 70 (finished) created at:
  github.com/apache/pulsar-client-go/pulsar.newPartitionConsumer()
      /Users/me/pulsar-client-go/pulsar/consumer_partition.go:184 +0xead
  github.com/apache/pulsar-client-go/pulsar.(*consumer).internalTopicSubscribeToPartitions.func1()
[...]

PR #451, opened 2021-01-25
Issue #448 make state thread safe in consumer_partition and connection

zzzming:
### Motivation
Address Issue: #448 and connection.go also missing a few sync against state

### Modifications

Use go sync/atomic to synchronize state and add to missing ones in connection.go

I did not use uber's atomic as suggested by @flowchartsman. The sync/atomic seems adequate and avoid another dependency. If uber's atomic is required, we should update other sync/atomic to be consistent. @merlimat @wolfstudy what do you think? thanks

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - If yes, how is the feature documented?


flowchartsman:
I'm not particularly attached to uber's atomic, I used it for expediency's sake, but direct calls to atomic always make me nervous if I'm doing them from multiple places, so  guarding it in a (probably inlined) method is usually the way I like to go. YMMV. But, if it works and it's consistent with other code, who am I to judge? ¯\_ʕ◔ϖ◔ʔ_/¯ Aside from that, LGTM

zzzming:
> I'm not particularly attached to uber's atomic, I used it for expediency's sake, but direct calls to atomic always make me nervous if I'm doing them from multiple places, so guarding it in a (probably inlined) method is usually the way I like to go. YMMV. But, if it works and it's consistent with other code, who am I to judge? ¯_ʕ◔ϖ◔ʔ_/¯ Aside from that, LGTM

@flowchartsman I share the same sentiment that I would tend to forget about add sync.atomic too. Let's wait for Matteo to weight in. It's a minor change either way. cheers

merlimat:
The `uber/atomic` looks indeed more readable and safer than the `sync/atomic`. It would be good to use that consistently

zzzming:
> The `uber/atomic` looks indeed more readable and safer than the `sync/atomic`. It would be good to use that consistently

updated all states with uber/atomic

pulsar/producer_partition.go
```diff
@@ -703,8 +707,22 @@ func (p *partitionProducer) Flush() error {
 	return cp.err
 }
 
+func (p *partitionProducer) getProduerState() producerState {
```
merlimat, line 710:
```suggestion
func (p *partitionProducer) getProducerState() producerState {
```

pulsar/producer_partition.go
```diff
@@ -263,7 +267,7 @@ func (p *partitionProducer) reconnectToBroker() {
 	}
 
 	for maxRetry != 0 {
-		if atomic.LoadInt32(&p.state) != producerReady {
+		if p.getProduerState() != producerReady {
```
merlimat, line 270:
```suggestion
		if p.getProducerState() != producerReady {
```

pulsar/producer_partition.go
```diff
@@ -443,7 +447,7 @@ func (p *partitionProducer) internalFlushCurrentBatch() {
 
 func (p *partitionProducer) failTimeoutMessages() {
 	// since Closing/Closed connection couldn't be reopen, load and compare is safe
-	state := atomic.LoadInt32(&p.state)
+	state := p.getProduerState()
```
merlimat, line 450:
```suggestion
	state := p.getProducerState()
```

pulsar/producer_partition.go
```diff
@@ -703,8 +707,22 @@ func (p *partitionProducer) Flush() error {
...
+
+// set a new consumerState and return the last state
+// returns bool if the new state has been set or not
+func (p *partitionProducer) casProducerState(oldState, newState producerState) bool {
+	return p.state.CAS(int32(oldState), int32(newState))
+}
+
 func (p *partitionProducer) Close() {
-	if atomic.LoadInt32(&p.state) != producerReady {
+	if p.getProduerState() != producerReady {
```
merlimat, line 725:
```suggestion
	if p.getProducerState() != producerReady {
```

merlimat (approved):

PR #454, opened 2021-01-26
Make backoff policy configurable and use uniformly

wuYin:
### Motivation

For now, client backoff interval is hardcoded to start from 100ms and increasing exponentially to 60s.
In some case, client need reconnect to restarted broker as soon as possible, rather than still waiting while broker is available.
Like this one, during the 51.2s backoff interval, broker already restarted and available.
image

### Modifications

- Modular backoff policy and make it configurable, retained mandatory stop policy which added in [pulsar#747](https://github.com/apache/pulsar/pull/747) 
- Added  `backoff_test.go` to test backoff's functionality
- Added 2 backoff configurations for Client and use uniformly in producer, consumer reconnection
    ```go
	// Set the initial duration of time for a backoff interval (default: 100ms)
	StartingBackoffInterval time.Duration
	// Set the maximum duration of time for a backoff interval (default: 60s)
	MaxBackoffInterval time.Duration
    ```

### Verifying this change


pulsar/internal/backoff.go
```diff
@@ -18,28 +18,64 @@
...
+	if b.mandatoryStop > 0 && !b.mandatoryStopped {
+		now := systemNow()
+		tried := time.Duration(0)
+		if curGap == b.initGap {
+			b.firstRetry = now
+		} else {
+			tried = now.Sub(b.firstRetry)
+		}
+		if tried+curGap > b.mandatoryStop {
+			// reached mandatory stop, reset current retry gap
```
wuYin, line 70: If mandatoryStop set to 0, this always be true in first time, means there is no mandatory stop which is on expected.
So I rule it out in above `if b.mandatoryStop > 0`, more readable.

wuYin:
@merlimat Help review if you have time, thanks a lot.

merlimat:
The problem I see with this change is that it allows misconfigured clients that retry too quickly, to overwhelm brokers with a flood of requests.

wuYin:
> The problem I see with this change is that it allows misconfigured clients that retry too quickly, to overwhelm brokers with a flood of requests.

Thanks for review.
I understand this concern, for now backoff intervals kept default in `[100ms, ..., 60s]`
I assumed user knows what they're doing, for me, I may need reduce `MaxBackoffInterval` to 20s for reducing recovery time.
Should I add a WARNING log if `MaxBackoffInterval` less than 1s, or something else?

wuYin:
@merlimat Help review if you have time, thanks a lot.

Now I added 2 thresholds for backoff
-  `minInitBackoff` which is 1ms
- `minMaxBackoff` which is 1sec

If user-provided backoff config less than these, create client will be failed with a invalid config error, checked by  `TestMisconfiguredBackoff`

wuYin:
@merlimat

wuYin:
@merlimat

wuYin:
@merlimat 
If this features is intentionally restrained to avoid the risk of misconfiguration, it's ok to close this PR, I'll do stricter timeout checking in my application.
Thanks for the review and looking forward to reply.

wolfstudy:
> @merlimat
> If this features is intentionally restrained to avoid the risk of misconfiguration, it's ok to close this PR, I'll do stricter timeout checking in my application.
> Thanks for the review and looking forward to reply.

@wuYin Close this pull request?

PR #457, opened 2021-01-29
Fix unsubscribe blocked when consumer is closing or has closed

wuYin:
### Motivation

For the present consumer, `Close()` and `Unsubscribe()` handled by the same eventloop goroutine.
The eventloop exited after `Close()`, then unsubscribe event wouldn't be selected and handled anymore, lead to block.

example:
```go
func main() {
	client, err := pulsar.NewClient(pulsar.ClientOptions{URL: "pulsar://localhost:6650"})
	if err != nil {
		log.Fatal(err)
	}
	defer client.Close()

		Topic:            "topic-1",
		SubscriptionName: "my-sub",
	})
	if err != nil {
		log.Fatal(err)
	}
	
	defer consumer.Unsubscribe() // unintentional
	defer consumer.Close()
}
```

`Unsubscribe()` blocked:

image

### Modifications

Check consumer state before send unsubscribe event, if consumer is closing or has closed, just logging it

### Verifying this change

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -47,6 +47,21 @@ const (
 	consumerClosed
 )
 
+func (s consumerState) String() string {
+	switch s {
+	case consumerInit:
+		return "Initializing"
+	case consumerReady:
+		return "Ready"
+	case consumerClosing:
+		return "Closing"
+	case consumerClosed:
+		return "Closed"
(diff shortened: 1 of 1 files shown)


freeznet (approved):
LGTM

wolfstudy (approved):
LGTM +1

PR #459, opened 2021-02-02
Improve error log for frame size too big and maxMessageSize

zzzming:
### Motivation
This PR is to improve error log for `frame size too big` on the consumer. We have seen a number of frame size too big error and I would like to trace what the exact maxMessageSize is set to that is not default. Debug log was not able to be enabled on the production system.

### Modifications

Log improvement mostly

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - If yes, how is the feature documented? no

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -348,7 +348,7 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
 		p.log.WithError(errMessageTooLarge).
 			WithField("size", len(payload)).
 			WithField("properties", msg.Properties).
-			Error()
+			Errorf("MaxMessageSize %d", int(p.cnx.GetMaxMessageSize()))
 		p.metrics.PublishErrorsMsgTooLarge.Inc()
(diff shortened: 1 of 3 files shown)


pulsar/producer_partition.go
```diff
@@ -348,7 +348,7 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
 		p.log.WithError(errMessageTooLarge).
 			WithField("size", len(payload)).
 			WithField("properties", msg.Properties).
-			Error()
+			Errorf("MaxMessageSize %d", int(p.cnx.GetMaxMessageSize()))
```
merlimat, line 351: Should we keep this in a structured form instead? 
```suggestion

			WithField("max-message-size", int(p.cnx.GetMaxMessageSize())).
			Error()
```

freeznet (approved):
LGTM

wolfstudy (approved):
LGTM +1

Issue #462, opened 2021-02-05
goroutine leak


yuanweikang2020:
#### Expected behavior

```
func (c *connection) start() {
	// Each connection gets its own goroutine that will
	go func() {
		if c.connect() {
			if c.doHandshake() {
				c.run()
			} else {
				c.changeState(connectionClosed)
			}
		} else {
			c.changeState(connectionClosed)
		}
	}()
}
```
this code, If the pulsar doesn't work properly, it will cause a goroutine leak.

freeznet:
thanks @yuanweikang2020 for this issue, can you provide more context how to reproduce or verify this issue?

yuanweikang2020:
> thanks @yuanweikang2020 for this issue, can you provide more context how to reproduce or verify this issue?

when pulsar is failure, pulsar client reconnect
```
func (p *partitionProducer) runEventsLoop() {
	for {
		select {
		case i := <-p.eventsChan:
			switch v := i.(type) {
			case *sendRequest:
				p.internalSend(v)
			case *connectionClosed:
				p.reconnectToBroker()
			case *flushRequest:
				p.internalFlush(v)
			case *closeProducer:
				p.internalClose(v)
				return
			}

		case <-p.batchFlushTicker.C:
			p.internalFlushCurrentBatch()
		}
	}
}
```
when reconnect pulsar Broker
```
func (p *partitionProducer) reconnectToBroker() {
	backoff := internal.Backoff{}
	for {
		if p.state != producerReady {
			// Producer is already closing
			return
		}

		d := backoff.Next()
		p.log.Info("Reconnecting to broker in ", d)
		time.Sleep(d)

		err := p.grabCnx()
		if err == nil {
			// Successfully reconnected
			p.log.WithField("cnx", p.cnx.ID()).Info("Reconnected producer to broker")
			return
		}
	}
}
```
get connection frequently
```
func (c *rpcClient) getConn(logicalAddr *url.URL, physicalAddr *url.URL) (Connection, error) {
	cnx, err := c.pool.GetConnection(logicalAddr, physicalAddr)
	backoff := Backoff{1 * time.Second}
	startTime := time.Now()
	var retryTime time.Duration
	if err != nil {
		for time.Since(startTime) < c.requestTimeout {
			retryTime = backoff.Next()
			c.log.Debugf("Reconnecting to broker in {%v} with timeout in {%v}", retryTime, c.requestTimeout)
			time.Sleep(retryTime)
[...]

PR #473, opened 2021-02-22
Fix infinite loop in connection.failLeftRequestsWhenClose

PowerStateFailure:
Master Issue: #462

### Motivation

As mentioned in [this comment](https://github.com/apache/pulsar-client-go/issues/462#issuecomment-780357151), when connection closed, one of its goroutine stucks within `failLeftRequestsWhenClose` method as `incomingRequestsCh` gets never closed. This simple PR swaps order of operation done in this method so that it first closes the channel (as buffered chan used, it's perfectly valid and no data ever lost) and then handles any message left in `incomingRequestsCh` channel.

### Modifications

Mentioned above.

### Verifying this change

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -331,10 +331,10 @@ func (c *connection) waitUntilReady() error {
 }
 
 func (c *connection) failLeftRequestsWhenClose() {
+	close(c.incomingRequestsCh)
 	for req := range c.incomingRequestsCh {
 		c.internalSendRequest(req)
 	}
-	close(c.incomingRequestsCh)
 }
 
 func (c *connection) run() {
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -547,12 +547,18 @@ func (c *connection) SendRequest(requestID uint64, req *pb.BaseCommand,
 	callback func(command *pb.BaseCommand, err error)) {
 	if c.getState() == connectionClosed {
 		callback(req, ErrConnectionClosed)
-	} else {
-		c.incomingRequestsCh <- &request{
-			id:       &requestID,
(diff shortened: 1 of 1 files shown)


pulsar/internal/connection.go
```diff
@@ -331,10 +331,10 @@ func (c *connection) waitUntilReady() error {
 }
 
 func (c *connection) failLeftRequestsWhenClose() {
+	close(c.incomingRequestsCh)
```
rueian, line 334: I don't think it is safe to close incomingRequestsCh here.
Won't it be panic on producer side of this channel?

PowerStateFailure: You're right, it may cause panic in `SendRequest`/`SendRequestNoWait` since channel state check and `incomingRequestsCh` is both not atomic.
As a dirty workaround i've added double check before write to `incomingRequestsCh` by ensuring `closeCh` not closed yet - see next commit

rueian: Sadly, the workaround won't works. You can test it with the following code with recent golang version. It will panic very quickly.

```golang
func main() {
	c1 := make(chan struct{})
	c2 := make(chan struct{})
	close(c1)
	close(c2)
	for {
		select {
		case <-c1:
		case c2 <- struct{}{}:
		}
	}
}

PowerStateFailure: My bad, completely miss that `case`s in `select` has undefined order of execution in general. At the moment I don't have any idea how to resolve data race here but wrap `SendRequest`/`SendRequestNoWait` with mutex, but I don't think it's appropriate solution.

Will reopen PR if find a suitable solution.

PR #474, opened 2021-02-22
Fix race condition/goroutine leak in partition discovery goroutine

PowerStateFailure:
Master issue: #462

### Motivation

In some cases consumer `Close` method might be called in nearly the same moment with background partiotion discovery routine tick. When `Close` acquires mutex first, it will release all underlying consumers, but that not guarantees backgound routine's not wait for the same mutex in `internalTopicSubscribeToPartitions`, so when `Close` completes, mutex gets acquired by not-finished-yet goroutine leading to new partition consumers gets created without any chance to close them.
The same logic is valid for producer partition discovery goroutine which is closed the same manner.

This PR modifies background discrovery goroutine the way it will be closed before acquiring mutex in `Close` method.

### Modifications

Described above.

### Verifying this change

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -49,12 +49,12 @@ type consumer struct {
 	// channel used to deliver message to clients
 	messageCh chan ConsumerMessage
 
-	dlq       *dlqRouter
-	rlq       *retryRouter
-	closeOnce sync.Once
-	closeCh   chan struct{}
-	errorCh   chan error
-	ticker    *time.Ticker
+	dlq           *dlqRouter
+	rlq           *retryRouter
+	closeOnce     sync.Once
+	closeCh       chan struct{}
+	errorCh       chan error
+	stopDiscovery func()
 
 	log     log.Logger
 	metrics *internal.TopicMetrics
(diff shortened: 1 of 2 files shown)


omnilight:
@PowerStateFailure I faced the same problem. My service is a proxy between foreign clients and pulsar. When foreign client connects/disconnects too often, they got another consumer for partition.

Could you please fix failed build?

@wolfstudy Could you please review this PR?

PowerStateFailure:
@omnilight

> Could you please fix failed build?

As i can see from [build log](https://github.com/apache/pulsar-client-go/pull/474/checks?check_run_id=1954445477#step:4:2573), there is race condition somewhere in connection acquire logic. As I don't change, i'm unsure if i should fix it.

BTW, it seems that producer has [the same issue](https://github.com/apache/pulsar-client-go/blob/21541d6d40a6324db7b3952d9e1d392d872ce231/pulsar/producer_impl.go#L132) with closing background partition discovery goroutine, so I'll try to fix it too.

PowerStateFailure:
Rebased to the latest master, thus avoiding undesirable data race error in tests.

merlimat (approved):
👍

PR #478, opened 2021-03-02
[fix #472] move GetPartitionedTopicMetadata to lookup service

### LINKED ISSUE #472: TCP socket opened leak using proxy
#### Expected behavior

When a client is connected, the proxy shouldn't continue to create a connection with the broker.

#### Actual behavior

Considering a client with one producer and one consumer,
A cluster with 20 proxies/20 brokers/7 ZK

##### With 500 clients

- java client conducts proxy to have 3500 opened TCP socket and it's stable over time.
- go client conducts proxy to have 4293 opened TCP socket and it keeps increasing rapidly to 5000, then slowly.

##### With 300k clients

Using go client,
The proxies continue to create connections indefinitely to brokers. 
Then it conducts the proxies nodes to use all socket the host could provide.
Then, the proxy becomes unable and another gets the same problem, etc.
Finally, the cluster becomes fully unreachable.

##### Notes

There is no debug logs from go client that it's reconnecting or something.

For each new TCP socket opened, a connection from proxy to broker is logged.

```
Feb 22 14:00:16 pulsar-c2-n24 pulsar-proxy[795547]: 14:00:16.796 [pulsar-proxy-io-2-11] INFO  org.apache.pulsar.client.impl.ConnectionPool - [[id: 0xf30e2d8f, L:/192.168.11.115:51920 - R:pulsar-c2-n36/192.168.11.60:2004]] Connected to server
```

#### Steps to reproduce

Create 500 clients using go client, for each creates a producer and a consumer to cluster with multiples proxies/brokers and follow the number of opened TCP connections on proxy using:

```bash
ss | grep ESTAB | wc -l
```

freeznet:
Fixes #472

### Motivation

#472 addressed a socket leakage in go client. with some debug works, a difference between java client and go client was found. one difference is go client send `PartitionedTopicMetadata` request with new connection, but java is using lookup service's connection. This pr added `GetPartitionedTopicMetadata` to lookup service to reuse the lookup service connection.

Through this pr, the leakage has been resolved with [this sample deployment](https://gist.github.com/freeznet/8681525ddd240877af88633363ba222f), the sockets are keep to 7 for 2 hours test. But still need to check with @KannarFr  to see if there are any other possible leakages since I cannot have a cluster with 20 proxies/20 brokers/7 ZK locally.

### Verifying this change


freeznet:
rerun failure checks

wolfstudy (approved):

wolfstudy:
> rerun failure checks

Hello @freeznet , currently the `rerun failure checks` not work, please check the failure test case.

michaeljmarshall:
@freeznet and @wolfstudy - the original issue references an issue with the proxy. Was this bug limited to connections with the proxy or does it include connections to brokers as well? Thanks.

freeznet:
> @freeznet and @wolfstudy - the original issue references an issue with the proxy. Was this bug limited to connections with the proxy or does it include connections to brokers as well? Thanks.

@michaeljmarshall according to my tests, the issue only appears on proxy, and I havnt noticed the similar socket leakage issue with standalone pulsar.

PR #492, opened 2021-03-24
[Issue 491][rpc_client]fix: timeout guarantee for RequestOnCnx

### LINKED ISSUE #491: Producer block may block forever
#### Expected behavior

If producer.Close are called,it will close within configured OperationTimeout

#### Actual behavior

producer.Close may block forever and won't time out after configured OperationTimeout

#### Steps to reproduce

Just run the perf program for a while and send SIGINT or ctrl+c to it. This situation may be triggered.

This is because the callback of [producer.SendAsyc](https://github.com/apache/pulsar-client-go/blob/master/perf/perf-producer.go#L132) contains a send behavior on a channel which under the same scope with [stopCh](https://github.com/apache/pulsar-client-go/blob/master/perf/perf-producer.go#L145-L162). When we trigger producer.Close, if `case <-stop:` happens before `case latency := <-ch:`, it will block forever because it may be block at [here](https://github.com/apache/pulsar-client-go/blob/master/pulsar/producer_partition.go#L700) and [RequestOnCnx](https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/rpc_client.go#L146-L165) has no time out guarantee.

#### System configuration
**Pulsar version**: any

jonyhy96:
Fixes #491 

### Motivation

see #491

### Modifications

add timeout guarantee to RequestOnCnx

### Verifying this change

  - **The public API: (yes)**
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/internal/rpc_client.go
+++ b/pulsar/internal/rpc_client.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"net"
 	"net/url"
-	"sync"
 	"sync/atomic"
 	"time"
 
@@ -30,6 +29,16 @@ import (
 	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
 	"github.com/gogo/protobuf/proto"
 )
+
+var (
+	// ErrRequestTimeOut happens when request not finished in given requestTimeout.
+	ErrRequestTimeOut = errors.New("request timed out")
+)
+
+type result struct {
+	*RPCResult
+	error
+}
 
 type RPCResult struct {
 	Response *pb.BaseCommand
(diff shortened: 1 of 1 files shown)


pulsar/internal/rpc_client.go
```diff
@@ -139,29 +144,30 @@ func (c *rpcClient) Request(logicalAddr *url.URL, physicalAddr *url.URL, request
...
 func (c *rpcClient) RequestOnCnx(cnx Connection, requestID uint64, cmdType pb.BaseCommand_Type,
 	message proto.Message) (*RPCResult, error) {
 	c.metrics.RPCRequestCount.Inc()
-	wg := sync.WaitGroup{}
-	wg.Add(1)
 
-	rpcResult := &RPCResult{
-		Cnx: cnx,
-	}
+	ch := make(chan result, 1)
```
wolfstudy, line 155: Can we consider adjusting the size of the channel buffer appropriately? It looks like 10 is a good choice?

wolfstudy (commented):
The change LGTM +1, just a little comments, please check

pulsar/internal/rpc_client.go
jonyhy96: 0 maybe enough for this because there will be only one result sends to `ch` . Set it to 1  to prevent block happens on the sender side.

omnilight:
@wolfstudy @jonyhy96 Hi. Could we proceed with this PR? Lack of the timeout on connection is a problem causing locking producer, would be great if we can release this fix

freeznet (approved):
LGTM

wolfstudy (approved):
Great job! LGTM +1

PR #497, opened 2021-03-28
[Issue 496] Support for correct reconnections limit in consumer

### LINKED ISSUE #496: MaxReconnectToBroker is useless for consumer
#### Expected behavior

When we set `ConsumerOptions.MaxReconnectToBroker` to some value other than infinite, then when reconnection count reaches this value consumer should return with an error.

#### Actual behavior

Consumer just stuck, because [reconnectToBroker](https://github.com/apache/pulsar-client-go/blob/master/pulsar/consumer_partition.go#L861) just exists leaving consumer useless

#### Steps to reproduce

1. Set option to `ConsumerOptions.MaxReconnectToBroker = 2`
2. Begin consuming messages
3. Do something with broker (say, stop it)
4. Messages would stop consuming
5. Start broker
6. Messages won't appear because consumer would not reconnect

#### System configuration
**Pulsar version**: 2.7

omnilight:
Fixes #496 

### Motivation

Ability to set maximum number of reconnections to broker for consumer without consumer been stuck

### Modifications

- Added new method `Closed() <-chan struct` to consumer interface to be able to control if consumer is closed when using its `Chan()` method
- For partitioned consumer added call of the parent's `Close()` method in case when number of reconnections is limited. This fixes problem when consumer stuck in disconnected state.

### Verifying this change

 - Added test for consumer closing on reaching max reconnection count
 - Added test for new `Closed()` method

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

  - If yes, how is the feature documented? (not documented)


pulsar/consumer.go
```diff
@@ -165,6 +165,9 @@ type Consumer interface {
 	// Unsubscribe the consumer
 	Unsubscribe() error
 
+	// Closed returns a channel indicating that consumer is closed
+	Closed() <-chan struct{}
+
```
wolfstudy, line 170: Why do we need to add this method, It looks like it has never been called?

wolfstudy (commented):
@omnilight Thanks for work on this, when `reconnectToBroker` fails, we should close the parentConsumer object. This pr is for the consumer side. Will the producer have the same problem?

pulsar/consumer.go
omnilight: Yep. Idea is to use it like that:
```
select {
  case msg := <- consumer.Chan();
     // handle msg
  case <-consume.Closed()
     return
}
```

So this method is used by user to check if consumer closed or not. We can not close message channel as it shared across partition conumers, so, this is the way of detecting it

rueian: Hi @omnilight, I am run into the same problem and thank you for providing a fix.

However, I think the new `Closed()` method is quite hard to use. why not just close the `Chan()` and let `Receive()` return an error?

rueian: Oh, I see https://github.com/apache/pulsar-client-go/pull/303 mentioned that the `Chan()` can't be closed.

BTW, can you also add a log message when `maxRetry` reached and saying that it will not continue retry?

omnilight: I think, yes, let's add

omnilight: @rueian done

omnilight:
Added fix to prevent possible panic when pulsar returns error in consumer initialization step

omnilight:
@wolfstudy Hi, could you please review this PR?

clawfinger:
Hello, do we have any ETA on this issue? Is it still planned for 0.8.0? Started using this lib recently and need this functionality badly

rueian:
I also think this PR is important. I frequently find my consumer stuck without knowing the maximum number of reconnection had been reached.

omnilight:
@wolfstudy Could you please release this fix within the 0.9.0 release?

omnilight:
@wolfstudy Sorry for annoying you, but is it possible to include this PR into next release?

omnilight:
Hello!

Any chance this PR will be accepted?

@RobertIndie sorry for dm

RobertIndie:
> #496

Hi @omnilight This PR looks good to me. But could you help resolve the conflicts? After that, I will review it again. Thanks.

omnilight:
> But could you help resolve the conflicts? After that, I will review it again. Thanks.

Yep, will fix that soon

omnilight:
@RobertIndie Hi! I've rebases on latest master, please take a look

PR #499, opened 2021-03-29
[issue #498]fix range channel deadlock error

### LINKED ISSUE #498: range channel causes deadlock
The usage of range channel in connection.go line 334 is not correct. The range channel will be deadlock.

Bug Code: https://github.com/apache/pulsar-client-go/blob/4d6c15114b9b6a0fc41b58993cf8acdd0e617b2f/pulsar/internal/connection.go#L334 

Reproduce method:
```go
package main

import (
	"fmt"
)

func main() {
	c := make(chan int, 10)
	for i := range c {
		fmt.Println(i)
	}
	close(c)
}
```

Output will be:
```
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive]:
main.main()
	/tmp/sandbox329282401/prog.go:9 +0xcb
```

It can't be seen in pulsar for the main goroutines has exit. But exactly it remains in memory and cause goroutine and memory leak.

godchen0212:
Signed-off-by: godchen0212 <qingxiang.chen@zilliz.com>

Fixes #498

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -331,8 +331,9 @@ func (c *connection) waitUntilReady() error {
 }
 
 func (c *connection) failLeftRequestsWhenClose() {
-	for req := range c.incomingRequestsCh {
-		c.internalSendRequest(req)
+	reqLen := len(c.incomingRequestsCh)
+	for i := 0; i < reqLen; i++ {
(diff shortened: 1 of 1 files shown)


wolfstudy (approved):
LGTM +1

merlimat (approved):

PR #502, opened 2021-04-06
Support listener name for go client

wolfstudy:
Signed-off-by: xiaolongran <rxl@apache.org>

### Motivation

In [PIP-61](https://github.com/apache/pulsar/wiki/PIP-61%3A-Advertised-multiple-addresses), we introduce a new feature `Advertised multiple addresses` and in Java Client, we have implemented this function. correspondingly, we should also support this function in Go client.

### Modifications

- Add listenerName filed in lookupService struct
- add test case for this feature

### Verifying this change


wolfstudy:
@codelipenghui @zymap @merlimat @gaoran10 @congbobo184 PTAL thanks

wolfstudy:
@freeznet PTAL thanks

codelipenghui (approved):

pulsar/internal/lookup_service.go
```diff
@@ -46,21 +46,25 @@ type LookupService interface {
 
 type lookupService struct {
 	rpcClient           RPCClient
-	serviceNameResolver ServiceNameResolver
+	serviceURL          *url.URL
```
zymap, line 49: Why you change this? Look like we need the `ServiceNameResolver` to resolve the multiple service address.

wolfstudy: This is just to adjust the position when resolving the conflict, but not removed

zymap (approved):

PR #507, opened 2021-04-09
Expose GetHashingFunction() method on producer

hunter2046:
### Motivation

Expose GetHashingFunction() so that the user could construct the hashFunc to call `NewDefaultRouter`.

### Modifications

Expose `GetHashingFunction()` and added unit tests.

### Verifying this change

- Added unit tests in producer_test.go

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -20,6 +20,8 @@ package pulsar
 import (
 	"context"
 	"time"
+
+	"github.com/apache/pulsar-client-go/pulsar/internal"
 )
 
 type HashingScheme int
@@ -198,4 +200,16 @@ type Producer interface {
 	// No more writes will be accepted from this producer. Waits until all pending write request are persisted. In case
 	// of errors, pending writes will not be retried.
 	Close()
+}
+
+// GetHashingFunction return the corresponding hashing function for the hashing scheme
+func GetHashingFunction(s HashingScheme) func(string) uint32 {
+	switch s {
+	case JavaStringHash:
+		return internal.JavaStringHash
+	case Murmur3_32Hash:
+		return internal.Murmur3_32Hash
+	default:
+		return internal.JavaStringHash
+	}
 }
--- a/pulsar/producer_impl.go
+++ b/pulsar/producer_impl.go
@@ -59,17 +59,6 @@ type producer struct {
 
 var partitionsAutoDiscoveryInterval = 1 * time.Minute
 
-func getHashingFunction(s HashingScheme) func(string) uint32 {
-	switch s {
-	case JavaStringHash:
-		return internal.JavaStringHash
-	case Murmur3_32Hash:
-		return internal.Murmur3_32Hash
-	default:
-		return internal.JavaStringHash
-	}
-}
-
 func newProducer(client *client, options *ProducerOptions) (*producer, error) {
 	if options.Topic == "" {
 		return nil, newError(InvalidTopicName, "Topic name is required for producer")
@@ -102,7 +91,7 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
 
 	if options.MessageRouter == nil {
 		internalRouter := NewDefaultRouter(
-			getHashingFunction(options.HashingScheme),
+			GetHashingFunction(options.HashingScheme),
 			options.BatchingMaxMessages,
 			options.BatchingMaxSize,
 			options.BatchingMaxPublishDelay,
--- a/pulsar/producer_test.go
+++ b/pulsar/producer_test.go
@@ -21,6 +21,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"reflect"
+	"runtime"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -39,6 +41,12 @@ func TestInvalidURL(t *testing.T) {
 	if client != nil || err == nil {
 		t.Fatal("Should have failed to create client")
 	}
+}
+
+func TestGetHashingFunction(t *testing.T) {
+	assertHashingFunctionEqual(t, internal.JavaStringHash, GetHashingFunction(-1))
+	assertHashingFunctionEqual(t, internal.JavaStringHash, GetHashingFunction(JavaStringHash))
+	assertHashingFunctionEqual(t, internal.Murmur3_32Hash, GetHashingFunction(Murmur3_32Hash))
 }
 
 func TestProducerConnectError(t *testing.T) {
@@ -1030,4 +1038,10 @@ func TestProducerWithInterceptors(t *testing.T) {
 
 	assert.Equal(t, 10, metric.sendn)
 	assert.Equal(t, 10, metric.ackn)
+}
+
+func assertHashingFunctionEqual(t *testing.T, func1, func2 func(string) uint32) {
+	funcName1 := runtime.FuncForPC(reflect.ValueOf(func1).Pointer()).Name()
+	funcName2 := runtime.FuncForPC(reflect.ValueOf(func2).Pointer()).Name()
+	assert.Equal(t, funcName1, funcName2)
 }


aahmed-se (approved):

merlimat:
In the `MessageRouter`, you're getting a `TopicMetadata` object which will tell you the number of partitions: https://github.com/apache/pulsar-client-go/blob/f17deac7ac47919aa9e46d11d33638672a24d126/pulsar/producer.go#L57

Wouldn't this already solve the problem?

hunter2046:
> In the `MessageRouter`, you're getting a `TopicMetadata` object which will tell you the number of partitions:
> 
> https://github.com/apache/pulsar-client-go/blob/f17deac7ac47919aa9e46d11d33638672a24d126/pulsar/producer.go#L57
> 
> Wouldn't this already solve the problem?

IIUC, the TopicMetadata gives us the number of partitions at the time the MessageRouter is invoked. For our use case, we need to keep recording the number of partitions when a new partition key is defined.

I probably shouldn't paste the link here but DSP-28866 is the jira ticket that this change is made for.

hunter2046:
I have updated the pull request to also expose the `GetHashingFunction()` method to make it possible for the user to call `NewDefaultRouter`.

The idea about how we plan to make use of the changes can be found at https://cd.splunkdev.com/data-availability/s2s-svc/-/merge_requests/937/diffs

pulsar/producer.go
```diff
@@ -199,3 +204,15 @@ type Producer interface {
 	// of errors, pending writes will not be retried.
 	Close()
 }
+
+// GetHashingFunction return the corresponding hashing function for the hashing scheme
+func GetHashingFunction(s HashingScheme) func(string) uint32 {
```
merlimat, line 209: Why do we need to expose this publicly? An application is free to use a custom message router function, at that point it can use any hashing function.

merlimat:
> IIUC, the TopicMetadata gives us the number of partitions at the time the MessageRouter is invoked. For our use case, we need to keep recording the number of partitions when a new partition key is defined.

I'm not convinced exposing `NumPartitions()` would help here. All the custom logic should be possible to implement within a custom message router. 

In any case, there's already a way to get the partitions for a topic, by using `Client.TopicPartitions()`

pulsar/producer.go
hunter2046: We would like to mimic what the default router does including the hash function. The only difference would be to use a snapshotted numPartitions instead of a latest one that is from `TopicMetadata`: https://github.com/apache/pulsar-client-go/blob/6ab17dcca28db389ab91a4a2ea6e5ae990d613da/pulsar/producer_impl.go#L103

hunter2046:
> > IIUC, the TopicMetadata gives us the number of partitions at the time the MessageRouter is invoked. For our use case, we need to keep recording the number of partitions when a new partition key is defined.
> 
> I'm not convinced exposing `NumPartitions()` would help here. All the custom logic should be possible to implement within a custom message router.
> 
> In any case, there's already a way to get the partitions for a topic, by using `Client.TopicPartitions()`

ACK, I didn't realize there is a method already. That should work. I removed changes related to `NumPartitions()`

wolfstudy (approved):
LGTM +1

wolfstudy:
The point is wrong, can MessageRouter not meet our functions?

e.g:

```
		Topic: "my-partitioned-topic",
		MessageRouter: func(msg *ProducerMessage, tm TopicMetadata) int {
			fmt.Println("Routing message ", msg, " -- Partitions: ", tm.NumPartitions())
			return 2
		},
	})
```

wolfstudy (changes requested):
Thanks @hunter2046 work for this, just a little comment, please check.

hunter2046:
> The point is wrong, can MessageRouter not meet our functions?
> 
> e.g:
> 
> ```
> 	producer, err := client.CreateProducer(ProducerOptions{
> 		Topic: "my-partitioned-topic",
> 		MessageRouter: func(msg *ProducerMessage, tm TopicMetadata) int {
> 			fmt.Println("Routing message ", msg, " -- Partitions: ", tm.NumPartitions())
> 			return 2
> 		},
> 	})
> ```

Thanks for taking a look. MessageRouter is what we need. We'd like to set the MessageRouter to a function that works similar to the defaultRouter with a customized number of partitions, e.g.

```
	internalRouter := pulsarClient.NewDefaultRouter(
		getHashingFunction(producerOptions.HashingScheme),
		producerOptions.BatchingMaxMessages,
		producerOptions.BatchingMaxSize,
		producerOptions.BatchingMaxPublishDelay,
		producerOptions.DisableBatching)
	messageRouter := func(message *pulsarClient.ProducerMessage, metadata pulsarClient.TopicMetadata) int {
                 // customNumberOfPartitions could be different from metadata.NumPartitions() here
		return internalRouter(message, customNumberOfPartitions)
	}
```

The problem is that `getHashingFunction` is not exposed so we can't use it to feed the `NewDefaultRouter` method.

wolfstudy:
> > The point is wrong, can MessageRouter not meet our functions?
> > e.g:
> > ```
> > 	producer, err := client.CreateProducer(ProducerOptions{
> > 		Topic: "my-partitioned-topic",
> > 		MessageRouter: func(msg *ProducerMessage, tm TopicMetadata) int {
> > 			fmt.Println("Routing message ", msg, " -- Partitions: ", tm.NumPartitions())
> > 			return 2
> > 		},
> > 	})
> > ```
> 
> Thanks for taking a look. MessageRouter is what we need. We'd like to set the MessageRouter to a function that works similar to the defaultRouter with a customized number of partitions, e.g.
> 
> ```
> 	internalRouter := pulsarClient.NewDefaultRouter(
> 		getHashingFunction(producerOptions.HashingScheme),
> 		producerOptions.BatchingMaxMessages,
> 		producerOptions.BatchingMaxSize,
> 		producerOptions.BatchingMaxPublishDelay,
> 		producerOptions.DisableBatching)
> 	messageRouter := func(message *pulsarClient.ProducerMessage, metadata pulsarClient.TopicMetadata) int {
>                  // customNumberOfPartitions could be different from metadata.NumPartitions() here
> 		return internalRouter(message, customNumberOfPartitions)
> 	}
> ```
> 
> The problem is that `getHashingFunction` is not exposed so we can't use it to feed the `NewDefaultRouter` method.

Sorry for the late reply. @hunter2046 Regarding all the options in NewDefaultRouter, we expose them in the form of parameters. You can set these options in producerOptions, and NewDefaultRouter is also the specific value obtained from these options.

hunter2046:
> > > The point is wrong, can MessageRouter not meet our functions?
> > > e.g:
> > > ```
> > > 	producer, err := client.CreateProducer(ProducerOptions{
> > > 		Topic: "my-partitioned-topic",
> > > 		MessageRouter: func(msg *ProducerMessage, tm TopicMetadata) int {
> > > 			fmt.Println("Routing message ", msg, " -- Partitions: ", tm.NumPartitions())
> > > 			return 2
> > > 		},
> > > 	})
> > > ```
> > 
> > 
> > Thanks for taking a look. MessageRouter is what we need. We'd like to set the MessageRouter to a function that works similar to the defaultRouter with a customized number of partitions, e.g.
> > ```
> > 	internalRouter := pulsarClient.NewDefaultRouter(
> > 		getHashingFunction(producerOptions.HashingScheme),
> > 		producerOptions.BatchingMaxMessages,
> > 		producerOptions.BatchingMaxSize,
> > 		producerOptions.BatchingMaxPublishDelay,
> > 		producerOptions.DisableBatching)
> > 	messageRouter := func(message *pulsarClient.ProducerMessage, metadata pulsarClient.TopicMetadata) int {
> >                  // customNumberOfPartitions could be different from metadata.NumPartitions() here
> > 		return internalRouter(message, customNumberOfPartitions)
> > 	}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The problem is that `getHashingFunction` is not exposed so we can't use it to feed the `NewDefaultRouter` method.
> 
> Sorry for the late reply. @hunter2046 Regarding all the options in NewDefaultRouter, we expose them in the form of parameters. You can set these options in producerOptions, and NewDefaultRouter is also the specific value obtained from these options.

Thanks. I understand that those options including MessageRouter are in producerOptions. My use case (as mentioned in https://github.com/apache/pulsar-client-go/pull/507#issuecomment-822802546) is that I need to set the MessageRouter in a customized way that is close to what the internalRouter does except using a different number of partitions to avoid causing producers to send messages to a different paritition when the number of partitions is being increased.

In order to do that, I need to initialize the internalRouter by my own by calling `NewDefaultRouter`. To call `NewDefaultRouter`, I could supply my own hashing function for sure. However, because I want to mimic what the internalRouter does, I would like using the same hashing function. That is why I want to see if `getHashingFunction` can be exposed so I could achieve the above thing. Hope that this makes sense.

PR #514, opened 2021-05-05
Make PartitionsAutoDiscoveryInterval configurable

hunter2046:
### Motivation
`partitionsAutoDiscoveryInterval` is currently hardcoded to 1 minute: https://github.com/apache/pulsar-client-go/blob/6ab17dcca28db389ab91a4a2ea6e5ae990d613da/pulsar/producer_impl.go#L60

I could think of two reasons for making it a configurable value:
* A faster or slower partitions discovery is desired. A faster one is more likely when users would like a partition increment to be effective more quickly.
* Allow a test to be able to set it to a much smaller time duration so that the test could exercise the logic within a reasonable short time

### Modifications

Added `PartitionsAutoDiscoveryInterval` into `ProducerOptions` with a default value of 1 minute.

### Verifying this change

### Documentation

  - If yes, how is the feature documented? (not applicable / docs / GoDocs / **not documented**)

### DIFF
--- a/pulsar/producer_impl.go
+++ b/pulsar/producer_impl.go
@@ -40,6 +40,9 @@ const (
 
 	// defaultMaxMessagesPerBatch init default num of entries in per batch.
 	defaultMaxMessagesPerBatch = 1000
+
+	// defaultPartitionsAutoDiscoveryInterval init default time interval for partitions auto discovery
+	defaultPartitionsAutoDiscoveryInterval = 1 * time.Minute
 )
 
 type producer struct {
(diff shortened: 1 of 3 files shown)


pulsar/producer_impl.go
```diff
@@ -87,6 +88,9 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
 	if options.BatchingMaxPublishDelay <= 0 {
 		options.BatchingMaxPublishDelay = defaultBatchingMaxPublishDelay
 	}
+	if options.PartitionsAutoDiscoveryInterval == 0 {
```
cckellogg, line 91: What happens if a negative value is passed in?

hunter2046: Thanks for the catch. Updated the check to include negative values.

cckellogg (approved):

merlimat (approved):

PR #517, opened 2021-05-13
Check ParseTopicName() return err

EAHITechnology:


pulsar/consumer_regex.go
```diff
@@ -383,11 +382,14 @@ func subscriber(c *client, topics []string, opts ConsumerOptions, ch chan Consum
 	return consumerErrorCh
 }
 
-func filterTopics(topics []string, regex *regexp.Regexp) []string {
+func filterTopics(topics []string, regex *regexp.Regexp) ([]string, error) {
 	matches := make(map[string]bool)
 	matching := make([]string, 0)
 	for _, t := range topics {
-		tn, _ := internal.ParseTopicName(t)
+		tn, err := internal.ParseTopicName(t)
```
cckellogg, line 389: I'm not sure if this good thing to error out here. The list of topic passed here usually comes from a response from the broker, right?

pulsar/producer_impl.go
```diff
@@ -88,12 +88,17 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
 		options.BatchingMaxPublishDelay = defaultBatchingMaxPublishDelay
 	}
 
+	ms, err := client.metrics.GetTopicMetrics(options.Topic)
```
cckellogg, line 91: This seems like a strange way of testing the topic. I don't think we should rely on calling a metrics function. Why not just call ` internal.ParseTopicName(topic)`. Same for the consumer_impl and tests.

pulsar/test_helper.go
```diff
@@ -153,15 +153,24 @@ func deleteTopic(topic string) error {
 
 func topicStats(topic string) (map[string]interface{}, error) {
 	var metadata map[string]interface{}
-	err := httpGet("admin/v2/persistent/"+topicPath(topic)+"/stats", &metadata)
+	tp, err := topicPath(topic)
+	if err != nil {
```
cckellogg, line 157: Not sure if this is needed since it's a test helper method. The test should fail without this check if a bad topic name is passed. Same for the topicPath function.

pulsar/producer_impl.go
EAHITechnology: The CreateProducer() passes an unchecked topic.And then use GetTopicMetrics() func (Use ParseTopicName func but uncheck err. When use as object transfer internal members ,you will get a runtime panic)

pulsar/consumer_regex.go
EAHITechnology: when c.client.namespaceTopics() return array length is 0 and err id nil.Does this situation exist?

pulsar/test_helper.go
EAHITechnology: This is just a protective measure to prevent modification of test cases. It's just that I will do it. I'm not sure that others are the same

pulsar/producer_impl.go
EAHITechnology: I resubmit here

EAHITechnology: I resubmit here

pulsar/consumer_regex.go
EAHITechnology: I'm not sure too.Although this is returned by the broker.

pulsar/producer_impl.go
EAHITechnology: If I verify the legitimacy on the outer layer. GetTopicMetrics() passes in *TopicName, is it reasonable? In the same way, the following operations do not need to verify the legitimacy

EAHITechnology: done!

pulsar/consumer_partition_test.go
```diff
@@ -92,12 +104,18 @@ func TestBatchMessageIDNoAckTracker(t *testing.T) {
 
 func TestBatchMessageIDWithAckTracker(t *testing.T) {
 	eventsCh := make(chan interface{}, 1)
+
+	tn, err := internal.ParseTopicName("topic")
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
wolfstudy, line 112: Why do we need to parse topic name here? In the `GetTopicMetrics` method, we have called the `ParseTopicName` method, is it enough to directly handle the error value returned by that method?

wolfstudy (commented):
Thanks @EAHITechnology work for this, thc change lgtm +1, just a little comments, please check.

wolfstudy:
@EAHITechnology Any update for this?

freeznet:
I think we already tested the topics with `func validateTopicNames(topics ...string) ([]*internal.TopicName, error)`, so could you please fill this PR's content to help us understand the context and what issue is trying to solve? thanks.

wolfstudy:
> I think we already tested the topics with `func validateTopicNames(topics ...string) ([]*internal.TopicName, error)`, so could you please fill this PR's content to help us understand the context and what issue is trying to solve? thanks.

hello @freeznet In ParseTopicName, we threw an error, but during the call, we ignored the error. such as:

in `filterTopics`

```
tn, _ := internal.ParseTopicName(t)
```

in `GetTopicMetrics`:

```
tn, _ := ParseTopicName(t)
```

PR #525, opened 2021-05-22
[bug] Fix reader latest position

addisonj:
Currently, using reader.latest fails because we discard
all messages less than the startPositionID, this causes an issue with
the reader as we filter all messages as all messages are less than
latest message id


wolfstudy:
@addisonj The change lgtm+1, please fix action CI, thanks.

zymap (approved):

PR #526, opened 2021-05-22
Add messageID from string and from parts

addisonj:
If we want to deal with string message ids, we need methods for being
able to create them from either a string or a using the individual parts
of the message id.

This exposes some constructors for this as well as an equals method to make testing more straight forward (but I am not sure if we want to expose equals...)

This is primarily to match the java implementation, but if we don't want to add the string version, we should at least expose the method to make up a message id from the individual parts.

This commit adds tests for the needed new methods

### DIFF
--- a/pulsar/message.go
+++ b/pulsar/message.go
@@ -18,11 +18,12 @@
 package pulsar
 
 import (
-	"github.com/pkg/errors"
 	"math"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 // ProducerMessage abstraction used in Pulsar producer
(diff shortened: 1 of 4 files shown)


pulsar/message.go
```diff
@@ -120,19 +124,65 @@ type Message interface {
 type MessageID interface {
 	// Serialize the message id into a sequence of bytes that can be stored somewhere else
 	Serialize() []byte
+	// String the message id represented as a string
+	String() string
```
cckellogg, line 128: This is not needed the struct just needs to implement the Stringer interface
https://golang.org/pkg/fmt/#Stringer

pulsar/message.go
```diff
@@ -120,19 +124,65 @@ type Message interface {
...
+	// Equals indicates to message IDs are equal
+	Equals(other MessageID) bool
```
cckellogg, line 130: I don't think we should add this either. We should avoid changing the interfaces since it's a breaking change and this can be accomplished with out doing that. We can add a util method like `messageIDsEqual or MessageIDsEquals`.

wolfstudy:
ping @addisonj any update for this?

PR #528, opened 2021-05-28
[issue 513] fix batch size limit validation

### LINKED ISSUE #513: BatchingMaxMessages option is ignored
#### Expected behavior

When I initialize client like the following:

```
   BatchingMaxMessages: 2
}
```

I expect it will send maximum 2 messages in the batch, even if I send using `SendAsync`.

#### Actual behavior

Parameter is ignored in batch builder. Both `batchContainer` and `keyBasedBatchContainer` during calling of their method `Add` use `hasSpace` method internally. But this method looks like the following:

```

// for simple batch:
func (bc *batchContainer) hasSpace(payload []byte) bool {
	return bc.numMessages > 0 && (bc.buffer.ReadableBytes()+msgSize) > uint32(bc.maxBatchSize)
}

// for key based:
func (bc *keyBasedBatchContainer) hasSpace(payload []byte) bool {
	return bc.numMessages > 0 && (bc.buffer.ReadableBytes()+msgSize) > uint32(bc.maxBatchSize)
}
```

So it does not check for the `maxMessages`. Option. Moreover. Is is obvious that this method should always return false at the first call of method `Add` because it has a condition `bc.numMessages > 0`, but this bug is compensated by another bug in method `Send` itself:

```
if replicateTo != nil && bc.numMessages != 0 {
		// If the current batch is not empty and we're trying to set the replication clusters,
		// then we need to force the current batch to flush and send the message individually
		return false
	} else if bc.msgMetadata.ReplicateTo != nil {
		// There's already a message with cluster replication list. need to flush before next
		// message can be sent
		return false
	} else if bc.hasSpace(payload) {
		// The current batch is full. Producer has to call Flush() to
		return false
	}
```

Calling method `bc.hasSpace` must be `!bc.hasSpace`.

As a result: batch limiting does not work as expected and the only way to fix that from the client code is to use `Disable batching option`

#### Steps to reproduce

Just set option `BatchingMaxMessages` to any value and send more message using `SendAsync` method. Batch would be finished only when size (in bytes) of bytes would be exceeded, not when number of messages would exceed limit

#### System configuration
**Pulsar version**: 2.7

zzzming:
Fixes #513

fix batch size limit validation

### Motivation

The current batch size validation is incorrect. 

### Modifications

The batch max size should be used to compare against the current size. The message should be flushed when the max size is reached.

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: ( no)

### Documentation

### DIFF
--- a/pulsar/internal/batch_builder.go
+++ b/pulsar/internal/batch_builder.go
@@ -153,7 +153,7 @@ func (bc *batchContainer) IsFull() bool {
 
 func (bc *batchContainer) hasSpace(payload []byte) bool {
 	msgSize := uint32(len(payload))
-	return bc.numMessages > 0 && (bc.buffer.ReadableBytes()+msgSize) > uint32(bc.maxBatchSize)
+	return bc.numMessages+1 > bc.maxMessages && (bc.buffer.ReadableBytes()+msgSize) > uint32(bc.maxBatchSize)
 }
 
 // Add will add single message to batch.
@@ -170,7 +170,7 @@ func (bc *batchContainer) Add(
 		// There's already a message with cluster replication list. need to flush before next
 		// message can be sent
 		return false
-	} else if bc.hasSpace(payload) {
+	} else if !bc.hasSpace(payload) {
 		// The current batch is full. Producer has to call Flush() to
 		return false
 	}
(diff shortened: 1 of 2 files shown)


wolfstudy:
@zzzming Thanks work for this, the change LGTM +1, please fix CI error.

freeznet (approved):
LGTM

wolfstudy:
@zzzming Any update for this change?

wolfstudy:
Please merge master code and fix CI error, thanks.

PR #535, opened 2021-06-08
Fix data race while accessing connection in partitionConsumer

dferstay:
The partitionConsumer maintains a few internal go-routines, two of which
access the underlying internal.Connection.  The main runEvenstLoop()
go-routine reads the connection field while a separate go-routine is used
to detect connnection loss, initiate reconnection, and sets the connection.

Previously, access to the conn field was not synchronized.

Now, the conn field is read and written atomically; avoiding race
conditions.

### Motivation

While attempting to submit a separate PR (https://github.com/apache/pulsar-client-go/pull/534) I found that the `pulsar/reader_test` consistently failed with the following data race.  This change in this PR is an attempt to fix it. 

```
2021-06-07T22:33:43.1825587Z ==================
2021-06-07T22:33:43.1825992Z WARNING: DATA RACE
2021-06-07T22:33:43.1826513Z Read at 0x00c0003de4a8 by goroutine 463:
2021-06-07T22:33:43.1828038Z   github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).requestGetLastMessageID()
2021-06-07T22:33:43.1829418Z       /pulsar-client-go/pulsar/consumer_partition.go:279 +0x27e
2021-06-07T22:33:43.1830873Z   github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).internalGetLastMessageID()
2021-06-07T22:33:43.1832427Z       /pulsar-client-go/pulsar/consumer_partition.go:270 +0xea
2021-06-07T22:33:43.1833799Z   github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).runEventsLoop()
2021-06-07T22:33:43.1835074Z       /pulsar-client-go/pulsar/consumer_partition.go:806 +0x2cb
2021-06-07T22:33:43.1835559Z 
2021-06-07T22:33:43.1836251Z Previous write at 0x00c0003de4a8 by goroutine 294:
2021-06-07T22:33:43.1837949Z time="2021-06-07T22:33:41Z" level=info msg="[Connected consumer]" consumerID=2 name= subscription=reader-kcnmq topic=my-topic-971826719
2021-06-07T22:33:43.1839441Z   github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).grabConn()
2021-06-07T22:33:43.1841513Z       /pulsar-client-go/pulsar/consumer_partition.go:974 +0x1875
2021-06-07T22:33:43.1842783Z   github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).reconnectToBroker()
2021-06-07T22:33:43.1844507Z       /pulsar-client-go/pulsar/consumer_partition.go:887 +0x2db
2021-06-07T22:33:43.1845873Z   github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).runEventsLoop.func2()
2021-06-07T22:33:43.1847183Z       /pulsar-client-go/pulsar/consumer_partition.go:791 +0xbe
2021-06-07T22:33:43.1847678Z 
2021-06-07T22:33:43.1848159Z Goroutine 463 (running) created at:
2021-06-07T22:33:43.1849312Z   github.com/apache/pulsar-client-go/pulsar.newPartitionConsumer()
2021-06-07T22:33:43.1850983Z       /pulsar-client-go/pulsar/consumer_partition.go:208 +0xf46
2021-06-07T22:33:43.1852092Z   github.com/apache/pulsar-client-go/pulsar.newReader()
2021-06-07T22:33:43.1853173Z       /pulsar-client-go/pulsar/reader_impl.go:105 +0x8ab
2021-06-07T22:33:43.1854294Z   github.com/apache/pulsar-client-go/pulsar.(*client).CreateReader()
2021-06-07T22:33:43.1855391Z       /pulsar-client-go/pulsar/client_impl.go:170 +0xcb
2021-06-07T22:33:43.1857422Z   github.com/apache/pulsar-client-go/pulsar.TestReaderLatestInclusiveHasNext()
2021-06-07T22:33:43.1858855Z       /pulsar-client-go/pulsar/reader_test.go:587 +0x946
2021-06-07T22:33:43.1859491Z   testing.tRunner()
2021-06-07T22:33:43.1860118Z       /usr/local/go/src/testing/testing.go:909 +0x199
2021-06-07T22:33:43.1860498Z 
2021-06-07T22:33:43.1860966Z Goroutine 294 (running) created at:
2021-06-07T22:33:43.1862630Z   github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).runEventsLoop()
2021-06-07T22:33:43.1864091Z       /pulsar-client-go/pulsar/consumer_partition.go:784 +0x174
2021-06-07T22:33:43.1865119Z ==================
```
### Modifications

Store the internal.Connection managed for the partitionConsumer in an `atomic.Value`: https://golang.org/pkg/sync/atomic/#Value

### Verifying this change

This change is already covered by existing tests that use `partitionConsumer` instances, such as `pulsar/reader_test`.

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -107,7 +107,7 @@ type partitionConsumer struct {
 	state          atomic.Int32
 	options        *partitionConsumerOpts
 
-	conn internal.Connection
+	conn atomic.Value
 
 	topic        string
 	name         string
@@ -238,7 +238,7 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
 		RequestId:  proto.Uint64(requestID),
 		ConsumerId: proto.Uint64(pc.consumerID),
 	}
-	_, err := pc.client.rpcClient.RequestOnCnx(pc.conn, requestID, pb.BaseCommand_UNSUBSCRIBE, cmdUnsubscribe)
+	_, err := pc.client.rpcClient.RequestOnCnx(pc._getConn(), requestID, pb.BaseCommand_UNSUBSCRIBE, cmdUnsubscribe)
 	if err != nil {
 		pc.log.WithError(err).Error("Failed to unsubscribe consumer")
 		unsub.err = err
@@ -248,7 +248,7 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
 		return
 	}
 
-	pc.conn.DeleteConsumeHandler(pc.consumerID)
+	pc._getConn().DeleteConsumeHandler(pc.consumerID)
 	if pc.nackTracker != nil {
 		pc.nackTracker.Close()
 	}
@@ -276,7 +276,7 @@ func (pc *partitionConsumer) requestGetLastMessageID() (trackingMessageID, error
 		RequestId:  proto.Uint64(requestID),
 		ConsumerId: proto.Uint64(pc.consumerID),
 	}
-	res, err := pc.client.rpcClient.RequestOnCnx(pc.conn, requestID,
+	res, err := pc.client.rpcClient.RequestOnCnx(pc._getConn(), requestID,
 		pb.BaseCommand_GET_LAST_MESSAGE_ID, cmdGetLastMessageID)
 	if err != nil {
 		pc.log.WithError(err).Error("Failed to get last message id")
@@ -326,7 +326,7 @@ func (pc *partitionConsumer) internalRedeliver(req *redeliveryRequest) {
 		}
 	}
 
-	pc.client.rpcClient.RequestOnCnxNoWait(pc.conn,
+	pc.client.rpcClient.RequestOnCnxNoWait(pc._getConn(),
 		pb.BaseCommand_REDELIVER_UNACKNOWLEDGED_MESSAGES, &pb.CommandRedeliverUnacknowledgedMessages{
 			ConsumerId: proto.Uint64(pc.consumerID),
 			MessageIds: msgIDDataList,
@@ -399,7 +399,7 @@ func (pc *partitionConsumer) requestSeekWithoutClear(msgID messageID) error {
 		MessageId:  id,
 	}
 
-	_, err = pc.client.rpcClient.RequestOnCnx(pc.conn, requestID, pb.BaseCommand_SEEK, cmdSeek)
+	_, err = pc.client.rpcClient.RequestOnCnx(pc._getConn(), requestID, pb.BaseCommand_SEEK, cmdSeek)
 	if err != nil {
 		pc.log.WithError(err).Error("Failed to reset to message id")
 		return err
@@ -435,7 +435,7 @@ func (pc *partitionConsumer) internalSeekByTime(seek *seekByTimeRequest) {
 		MessagePublishTime: proto.Uint64(uint64(seek.publishTime.UnixNano() / int64(time.Millisecond))),
 	}
 
-	_, err := pc.client.rpcClient.RequestOnCnx(pc.conn, requestID, pb.BaseCommand_SEEK, cmdSeek)
+	_, err := pc.client.rpcClient.RequestOnCnx(pc._getConn(), requestID, pb.BaseCommand_SEEK, cmdSeek)
 	if err != nil {
 		pc.log.WithError(err).Error("Failed to reset to message publish time")
 		seek.err = err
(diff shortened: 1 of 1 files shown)


freeznet (approved):

pulsar/consumer_partition.go
```diff
@@ -971,9 +972,9 @@ func (pc *partitionConsumer) grabConn() error {
 		pc.name = res.Response.ConsumerStatsResponse.GetConsumerName()
 	}
 
-	pc.conn = res.Cnx
+	pc._setConn(res.Cnx)
 	pc.log.Info("Connected consumer")
-	pc.conn.AddConsumeHandler(pc.consumerID, pc)
+	pc._getConn().AddConsumeHandler(pc.consumerID, pc)
```
cckellogg, line 977: What's the behavior if _getConn() ever returns nil? Same for all the other places.

dferstay: @cckellogg ,

Good question; `_getConn()` should never return `nil`.

An invariant in this code is that the `partitionConsumer.conn` field must be set and is never `nil`.
The `grabConn()` method sets the `partitionConsumer.conn` field; `grabConn()` is called from the `newPartitionConsumer` factory method which will fail construction of the partitionConsumer if grabConn() returns an error.

The above said, it is probably better for the cast in `_getConn()` to be unchecked and let the code `panic()` if the invariant is broken.  Thoughts?

dferstay: I've made the cast unchecked and added a comment that explains why.

cckellogg: There was change that moved the broker reconnect out of the events go routine (https://github.com/apache/pulsar-client-go/pull/376/files). This is now causing the data race issue. 

The question is if there is a reconnection what should happen with the pending events in the event channel? Right now they are processed using a stale/closed connection. Even with this fix it's possible events will try to get processed using a stale connection. Maybe that is ok but for me it makes the code difficult to follow and reason about. Ideally, we could come up with a cleaner way so we are never unintentionally using a stale/closed connection.

Thoughts?

dferstay: @cckellogg ,
I'm thinking that since [PR 376](https://github.com/apache/pulsar-client-go/pull/376) drains the connection incomingRequestsCh on close it should be possible remove the extra go-routine in the partitionConsumer and select from the connectionClosedCh in the partitonConsumer.runEventsLoop.  What do you think?

dferstay: @cckellogg ,
I've attempted the above with the following commit:
https://github.com/apache/pulsar-client-go/pull/535/commits/17335cd7846b71bd14592c944ea76f3fbcfdafd6

If this change is accepted I'll clean up this branch and update the PR description.

dferstay: In https://github.com/apache/pulsar-client-go/pull/535/commits/5dfcc1300eee30824efa5945bf463beb3bc3474e we also short-circuit broker re-connection attempts on consumer close.

dferstay: @cckellogg ,

I've reverted the behaviour of this PR back to the first approach, making access to the connection in the partitionConsumer atomic.

dferstay: The latest CI failure is going to be addressed by https://github.com/apache/pulsar-client-go/pull/544

pulsar/consumer_partition.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"math"
 	"sync"
+	syncAtomic "sync/atomic"
```
jonyhy96, line 24: perhaps you don't need to set alias to atomic package, just use atomic

pulsar/consumer_partition.go
```diff
@@ -1113,6 +1114,17 @@ func (pc *partitionConsumer) discardCorruptedMessage(msgID *pb.MessageIdData,
 		})
 }
 
+func (pc *partitionConsumer) _setConn(conn internal.Connection) {
+	pc.conn.Store(conn)
+}
+
+func (pc *partitionConsumer) _getConn() internal.Connection {
```
jonyhy96, line 1121: Maybe we don’t need _ as a function prefix

dferstay: @jonyhy96 ,

I wanted to make it clear that these methods should not be used by other components in the `pulsar` package, but I could remove this if you don't like it.  Admittedly this style of naming is not used anywhere else in the codebase.

pulsar/consumer_partition.go
dferstay: @jonyhy96 ,

Good suggestion; I've cut over to using [uber-go.atomic.Value](https://github.com/uber-go/atomic/blob/master/value.go)

pulsar/consumer_partition.go
jonyhy96: > I wanted to make it clear that these methods should not be used by other components in the pulsar package
Agree with this, but this is not a guarantee. Maybe we can add some comment on this function.

dferstay: Sure; comments added.

cckellogg (approved):

PR #541, opened 2021-06-14
#540 fix lookup service not implemented GetTopicsOfNamespace

### LINKED ISSUE #540: http lookup service not have namespace topics query implemented
#### Expected behavior

When using regex topics consumer with HTTP client URL, the client will fail to query namespace topics and throw failed handshake error. This is because the client sends protobuf requests to HTTP service. 

#### Actual behavior

get namespace topic lists from http lookup service with http URL.

freeznet:
Fixes #540

### Motivation

When using regex topics consumer with HTTP client URL, the client will fail to query namespace topics and throw failed handshake error. This is because the client sends protobuf requests to HTTP service.

### Modifications

add GetTopicsOfNamespace to LookupService interface
add related tests

### Verifying this change


congbobo184 (approved):

codelipenghui (approved):

pulsar/internal/namespace_name.go
```diff
@@ -0,0 +1,26 @@
...
+// under the License.
+
+package internal
+
+import "strings"
+
+func IsV2NameSpaceName(namespace string) bool {
+	parts := strings.Split(namespace, "/")
+	// Legacy namespace name that includes cluster name
+	return len(parts) != 3
```
cckellogg, line 25: should this be `len(parts) == 2` ? What if `ns` or  `ns/1/2/3` are passed in?
Let's be consistent and use `Namespace`

pulsar/consumer_regex.go
```diff
@@ -343,7 +343,7 @@ func (c *regexConsumer) unsubscribe(topics []string) {
 }
 
 func (c *regexConsumer) topics() ([]string, error) {
-	topics, err := c.client.namespaceTopics(c.namespace)
+	topics, err := c.client.lookupService.GetTopicsOfNamespace(c.namespace, internal.Persistent)
```
cckellogg, line 346: For regex are only persistent topics supported?

freeznet: not actually for other client like Java, but this PR does not tend to add `non-persistence` and `all` supporting, will do it in another PR.

wolfstudy (approved):
Thanks @freeznet , LGTM +1

wolfstudy:
@freeznet Please fix the CI error, thanks.

cckellogg (approved):

PR #548, opened 2021-06-22
regenerate the certs to work with Pulsar 2.8.0 and Java 11

freeznet:
### Motivation

Pulsar 2.8.0 moves to JDK 11 and no longer support SHA1 signature algorithm. ref: "[JEP 288: Disable SHA-1 Certificates](https://openjdk.java.net/jeps/288)".

To make pulsar-client-go CI works with the latest pulsar release, this PR re-generate certs, according to [Transport Encryption using TLS](http://pulsar.apache.org/docs/en/next/security-tls-transport/) and [Authentication using TLS](http://pulsar.apache.org/docs/en/next/security-tls-authentication/)

this PR is a rework for #544

### Verifying this change


pulsar/internal/http_client.go
```diff
@@ -52,7 +52,7 @@ func (c *httpClient) Close() {
 }
 
 type HTTPClient interface {
-	Get(endpoint string, obj interface{}) error
+	Get(endpoint string, obj interface{}, params map[string]string) error
```
merlimat, line 55: Is this change related to the certificate updates

pulsar/internal/lookup_service.go
```diff
@@ -325,9 +325,11 @@ func (h *httpLookupService) GetPartitionedTopicMetadata(topic string) (*Partitio
 
 	path := fmt.Sprintf(format, GetTopicRestPath(topicName))
 
+	h.log.Infof("Fetch topic{%s} partitioned metadata %s", topic, path)
```
merlimat, line 328: This we should maybe leave as debug level

pulsar/internal/http_client.go
freeznet: > Is this change related to the certificate updates

it is related to Pulsar 2.8.0, seems the `GetPartitionedTopicMetadata` with http lookup service has to pass `checkAllowAutoCreation=true` as params to get the partition num of the non-existed topic. So once we run CI with Pulsar 2.8.0, the tests will fail. [related with https://github.com/apache/pulsar/pull/10601]

pulsar/internal/lookup_service.go
freeznet: I have removed this line since I was trying to debug the reason why the tests are failed.

wolfstudy (approved):
Gread jobs

merlimat (approved):

PR #549, opened 2021-06-23
[Issue 515][pulsar-producer]Fix for Producer Send and SendAsyn is blocked forever when pulsar is down.

### LINKED ISSUE #515: Producer Send and SendAsyn is blocked for forever when pulsar is down
Producer Send and SendAsyn is blocked forever when pulsar is down if the MaxReconnectToBroker is set to ultimate. In case of pulsar down scenarios, within `runEventsLoop`in `producer_partition.go`, the call enters `reconnectToBroker` and remains in a forever loop until the pulsar broker connection is established. Due to this, no more events are consumed from `eventsChan` channel cause both Send and SendAsyn to be blocked. Due to this even the `SendTimeout` would also be not honoured. 

#### Expected behavior
Producer Send must not be blocked forever when the pulsar broker is down. It must honour the SendTimeout and return back with an error. 
Producer SendAsyn must never be blocked when the pulsar broker is down. It must honour the SendTimeout and call the callback function.

#### Actual behavior
Due the above mentioned issue Producer Send/SendAsyn blocks forever when the pulsar broker is down

#### Steps to reproduce
1. Create a pulsar producer and set the MaxReconnectToBroker as ultimate and SendTimeout as a fixed value
2. Send messages to pulsar using Send or SendAsyn API
3. Bring down pulsar broker or inject a connection error between broker and client
4. In case of Send, the call is blocker forever. In case of SendAsyn, the callback is never called and once the `pendingQueue` is filled, the call is blocked forever. 

#### System configuration
Pulsar client version - v0.4.0

mmahadikar-ns:
Issue link - https://github.com/apache/pulsar-client-go/issues/515

Signed-off-by: Megaraj Mahadikar <megarajtm@gmail.com>

Fixes #515

### Motivation
Producer `Send` and `SendAsyn` is blocked forever when pulsar is down if the MaxReconnectToBroker is set to unlimited retry. In case of pulsar down scenarios, within runEventsLoop in producer_partition.go, the call enters reconnectToBroker and remains in a forever loop until the pulsar broker connection is established. Due to this, no more events are consumed from eventsChan channel causing both `Send` and `SendAsyn` to be blocked. Due to this, the SendTimeout would also be not honoured.

### Modifications

In runEventsLoop, have a separate go-routine working on connectClosedCh channel. This way eventsChan is never blocked.

### Verifying this change
1. A new test is added in producer_test.go. `TestSendTimeoutWithUnlimitedRetry`

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -70,6 +70,7 @@ type partitionProducer struct {
 	// Channel where app is posting messages to be published
 	eventsChan      chan interface{}
 	connectClosedCh chan connectionClosed
+	doneCh          chan struct{}
 
 	publishSemaphore internal.Semaphore
 	pendingQueue     internal.BlockingQueue
@@ -106,6 +107,7 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		producerID:       client.rpcClient.NewProducerID(),
 		eventsChan:       make(chan interface{}, maxPendingMessages),
 		connectClosedCh:  make(chan connectionClosed, 10),
+		doneCh:           make(chan struct{}),
 		batchFlushTicker: time.NewTicker(batchingMaxPublishDelay),
 		publishSemaphore: internal.NewSemaphore(int32(maxPendingMessages)),
 		pendingQueue:     internal.NewBlockingQueue(maxPendingMessages),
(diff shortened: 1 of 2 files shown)


megarajtm:
Closing this PR. Will create a new PR for these changes

PR #552, opened 2021-06-24
Encryption support in Go client

GPrabhudas:
### Motivation

Added Go client changes to encrypt messages published to pulsar. Once encrypted, it only allows the consumer with the right set of keys to be able to decrypt the original message published by the producer. Only RSA encryption of data key is supported.

### Modifications
Changes are implemented as per [PIP-4](https://github.com/apache/incubator-pulsar/wiki/PIP-4:-Pulsar-End-to-End-Encryption)

Followed Java client [Encryption support changes](https://github.com/apache/pulsar/pull/731) 

Addresses issue [#333](https://github.com/apache/pulsar-client-go/issues/333)

PR #555, opened 2021-06-28
add base package for encryption

GPrabhudas:
Breakdown of PR  [#552 ](https://github.com/apache/pulsar-client-go/pull/552)

This PR includes base interface and default implementation.


pulsar/crypto/encryption_key_Info.go
```diff
@@ -0,0 +1,64 @@
...
+func NewEncryptionKeyInfo(key string, value []byte, metadata map[string]string) *EncryptionKeyInfo {
+	return &EncryptionKeyInfo{
+		metadata: metadata,
+		key:      key,
+		value:    value,
+	}
+}
+
+// GetKey get key
+func (eci *EncryptionKeyInfo) GetKey() string {
```
cckellogg, line 37: When i think of key i think of the actual bytes of the key what about just these functions
```
Name() string
Key() []byte
Meta() map[string]string

```

pulsar/crypto/message_metadata.go
```diff
@@ -0,0 +1,122 @@
...
+	GetEncryptionKeys() []EncryptionKeyInfo
+
+	// UpsertEncryptionkey add new or update existing EncryptionKeys in to the MessageMetadata
+	UpsertEncryptionkey(EncryptionKeyInfo)
+
+	// GetEncryptionParam read the ecryption parameter from the MessageMetadata
+	GetEncryptionParam() []byte
+
+	// SetEncryptionParam set encryption parameter in to the MessageMetadata
+	SetEncryptionParam([]byte)
```
cckellogg, line 36: Are setters need in this interface?

pulsar/crypto/message_metadata.go
```diff
@@ -0,0 +1,122 @@
...
+// under the License.
+
+package crypto
+
+import (
+	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
+)
+
+// MessageMetadataSupplier wrapper implementation around message metadata
+type MessageMetadataSupplier interface {
```
cckellogg, line 25: What is the goal of this class and am wondering if can come up with more meaningful name?

pulsar/crypto/pri_key_rsa.pem
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
cckellogg, line 1: move add these test keys file to a pulsar/crypto/testdata directory

https://golang.org/pkg/cmd/go/internal/test/
```
The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.
```

pulsar/crypto/crypto_key_reader.go
```diff
@@ -0,0 +1,28 @@
...
+package crypto
+
+// KeyReader implement this interface to read and provide public & private keys
+// key pair can be RSA, ECDSA
+type KeyReader interface {
+	// GetPublicKey get public key that is be used by the producer to encrypt data key
+	GetPublicKey(keyName string, metadata map[string]string) (*EncryptionKeyInfo, error)
+
+	// GetPrivateKey get private key that is used by the consumer to decrypt data key
+	GetPrivateKey(keyName string, metadata map[string]string) (*EncryptionKeyInfo, error)
```
cckellogg, line 27: Nitpick generally go code style does not use get for values. It would be PublicKey() and PrivateKey() instead. Same for other parts fo the code.

pulsar/crypto/crypto_failure_action.go
```diff
@@ -0,0 +1,46 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package crypto
+
+type ProducerCryptoFailureAction int
+type ConsumerCryptoFailureAction int
+
+const (
```
cckellogg, line 23: Since there aren't really enums in go to namespace things maybe we should use long names to make it easier to reason about in the code?

ProducerCryptoFailureActionFail or ProducerFailureActionSend
ProducerCryptoFailureActionSend or ProducerFailureActionFail

Thoughts?

pulsar/crypto/default_message_crypto.go
```diff
@@ -0,0 +1,379 @@
...
+		return err
+	}
+
+	parsedKey, err := d.loadPublicKey(keyInfo.GetValue())
+	if err != nil {
+		return err
+	}
+
+	// try to cast to RSA key
+	rsaPubKey, ok := parsedKey.(*rsa.PublicKey)
```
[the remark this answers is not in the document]
GPrabhudas: Yes currently only supporting RSA Keys and will update it accordingly

pulsar/crypto/crypto_key_reader.go
```diff
@@ -0,0 +1,28 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package crypto
+
+// KeyReader implement this interface to read and provide public & private keys
+// key pair can be RSA, ECDSA
+type KeyReader interface {
+	// GetPublicKey get public key that is be used by the producer to encrypt data key
+	GetPublicKey(keyName string, metadata map[string]string) (*EncryptionKeyInfo, error)
```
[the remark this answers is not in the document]
GPrabhudas: Key name is something user configures while creating producer in `ProducerOptions`

pulsar/crypto/encryption_key_Info.go
```diff
@@ -0,0 +1,64 @@
...
+	return eci.key
+}
+
+// SetKey set key
+func (eci *EncryptionKeyInfo) SetKey(key string) {
```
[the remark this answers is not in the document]
GPrabhudas: No need of setters. Removed them

pulsar/crypto/encryption_key_Info.go
GPrabhudas: done

pulsar/crypto/message_metadata.go
GPrabhudas: Yes they are needed.
1. Producer will update `MessageMetadata` with encryption parameter and encrypted data key.
2. Consumer will try to read the same info from the `MessageMetadata` to decrypt the payload.

There is need to have this setters here since we are not directly exposing `proto package`.

 If client(user) wants to have his own implementation of `MessageCrypto` interface, in such case he won't be able to access `MessageMetadata` directly because it is an `internal` package. 

So adding this interface and exposing necessary methods to read `encrypted data key and encryption parameter` from the `MessageMetadata` that are required for decryption of payload.

pulsar/crypto/message_metadata.go
GPrabhudas: Goal : Is to provide client(user) to set/read the required information in/from `MessageMetadata`.

Meaningful name => I think it is meaningful since it just providing a way to read some info from internal `MessageMetadata`

`MessageMetadata` => this is from the `proto` which is an internal package and client can access it directly and this class implemented required methods to read/write MessageMetadata.

pulsar/crypto/pri_key_rsa.pem
GPrabhudas: done.

pulsar/crypto/crypto_key_reader.go
GPrabhudas: Got it.

pulsar/crypto/crypto_failure_action.go
GPrabhudas: I tried to use them as enum, but yeah I think it make sense to have a long names instead.

GPrabhudas: And it is good have long name like `ProducerCryptoFailureActionFail` instead of `ProducerFailureActionSend`

pulsar/crypto/default_message_crypto.go
```diff
@@ -0,0 +1,379 @@
...
+
+	logger log.Logger
+
+	cipherLock sync.Mutex
+
+	encryptLock sync.Mutex
+}
+
+const (
+	errGeneratingNonce          = "failed to generate new nonce"
```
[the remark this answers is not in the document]
GPrabhudas: done.

pulsar/crypto/default_message_crypto.go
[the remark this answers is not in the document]
GPrabhudas: done

pulsar/crypto/crypto_failure_action.go
```diff
@@ -0,0 +1,43 @@
...
+// under the License.
+
+package crypto
+
+const (
+	// ProducerCryptoFailureActionFail this is the default option to fail send if crypto operation fails.
+	ProducerCryptoFailureActionFail = iota
+
+	// ProducerCryptoFailureActionFailSend ingnore crypto failure and proceed with sending unencrypted message.
+	ProducerCryptoFailureActionFailSend
```
cckellogg, line 25: nit `ProducerCryptoFailureActionSend`

pulsar/crypto/message_crypto.go
```diff
@@ -0,0 +1,39 @@
...
+
+// MessageCrypto implement this interface to encrypt and decrypt messages
+type MessageCrypto interface {
+	// AddPublicKeyCipher
+	AddPublicKeyCipher(keyNames []string, keyReader KeyReader) error
+
+	// RemoveKeyCipher remove the key indentified by the keyname from the list
+	RemoveKeyCipher(keyName string) bool
+
+	/*
```
cckellogg, line 28: Nitpick use `//` for each line in multiline comments

pulsar/crypto/message_metadata.go
cckellogg: Ok makes sense. For my understanding what's the difference between `UpsertEncryptionkey` and `SetEncryptionParam` and when would one or the other be used?

cckellogg (commented):
Just a few nits and questions. Also, thanks for working on this.

GPrabhudas:
> Ok makes sense. For my understanding what's the difference between UpsertEncryptionkey and SetEncryptionParam and when would one or the other be used?

[EncryptionParam](https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/pulsar_proto/PulsarApi.pb.go#L1361) =>  It is a  `nonce or padding`  used while encrypting and we need to pass same `nonce` while decrypting payload.  So `SetEncryptionParam` sets the `nonce`  in  `MessageMetadata`.

[EncryptionKeys](https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/pulsar_proto/PulsarApi.pb.go#L1357) => It is the Info about data key used to encrypt the payload . `UpsertEncryptionkey` sets/updates the encrypted `data key` and it's metadata if any in the `MessageMetadata`.


Producer uses `UpsertEncryptionkey` and  `SetEncryptionParam` to update MessageMetadata after encrypting payload.
Consumer uses `EncryptionKeys` and `EncryptionParam` to read from  MessageMetadata for decrypting payload.


@cckellogg

pulsar/crypto/message_crypto.go
GPrabhudas: Done.

pulsar/crypto/crypto_failure_action.go
GPrabhudas: done.

cckellogg (approved):

wolfstudy (approved):
LGTM +1

PR #557, opened 2021-06-28
[Issue 556]Add jitter to exponential backoff (#556)

### LINKED ISSUE #556: Backoff is calculated without jitter
The problem occurs when the connection to the broker is lost. The client will try to reconnect with an exponential backoff. The problem is that there is no jitter involved when calculating the backoff delay.

#### Expected behavior

Not all clients try to reconnect at the same time. 

#### Actual behavior

All clients try to reconnect at the same time to the brokers

#### Steps to reproduce

- Set up multiple clients connecting to the same set of brokers.
- Break the connection between the clients and the brokers

#### System configuration
**Pulsar version**: 0.5.0

dvergnes:
Fixes #556 

### Motivation
When reconnecting to a broker, an exponential backoff strategy is used. However, there is no randomness involved which means in case of a general outage impacting several clients, they will try to reconnect at the same time.

### Modifications

The modification consists of adding a jitter that adds 20% of delay after the delay is doubled on each retry.

### Verifying this change

This change added unit tests to verify the expected behavior.

  - The public API: no
  - The schema: no
  - The default values of configurations: no)
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/internal/backoff.go
+++ b/pulsar/internal/backoff.go
@@ -18,20 +18,27 @@
 package internal
 
 import (
+	"math/rand"
 	"time"
 )
 
-// Backoff
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+// Backoff computes the delay before retrying an action.
+// It uses an exponential backoff with jitter. The jitter represents up to 20 percents of the delay.
 type Backoff struct {
 	backoff time.Duration
 }
 
 const (
-	minBackoff = 100 * time.Millisecond
-	maxBackoff = 60 * time.Second
+	minBackoff       = 100 * time.Millisecond
+	maxBackoff       = 60 * time.Second
+	jitterPercentage = 0.2
 )
 
-// Next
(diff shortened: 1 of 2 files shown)


pulsar/internal/backoff.go
```diff
@@ -40,6 +43,7 @@ func (b *Backoff) Next() time.Duration {
 	} else if b.backoff.Nanoseconds() > maxBackoff.Nanoseconds() {
 		b.backoff = maxBackoff
 	}
+	jitter := rand.Float64() * float64(b.backoff) * jitterPercentage
```
cckellogg, line 46: Does the generator need to be seeded (https://golang.org/pkg/math/rand/#Seed)?

dvergnes: Yes, you are right. I updated the PR.

cckellogg (approved):

freeznet (approved):

wolfstudy (approved):

PR #560, opened 2021-07-08
Encryption support  producer

GPrabhudas:
Breakdown of PR [#552 ](https://github.com/apache/pulsar-client-go/pull/552)

This PR includes the encryption changes at producer side.


pulsar/internal/commands.go
```diff
@@ -269,6 +270,117 @@ func serializeBatch(wb Buffer,
 	wb.PutUint32(checksum, checksumIdx)
 }
 
+// copy of the method serializeBatch(....) with an extension to encrypt payload
+func serializeBatchWithEncryption(wb Buffer,
```
cckellogg, line 274: What are the differences between serializeBatch and serializeBatchWithEncryption? It looks like a lot of duplicate code?

pulsar/producer.go
```diff
@@ -163,6 +165,19 @@ type ProducerOptions struct {
 	// PartitionsAutoDiscoveryInterval is the time interval for the background process to discover new partitions
 	// Default is 1 minute
 	PartitionsAutoDiscoveryInterval time.Duration
+
+	// EncryptionKeys list of encryption key names to encrypt session key
+	EncryptionKeys []string
```
cckellogg, line 170: Should we wrap these into a struct? Maybe something like this?
```
type Encryption {
  Keys []string
  KeyReader crypto.KeyReader
  etc
]

```

go.mod
```diff
@@ -14,10 +14,7 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/klauspost/compress v1.10.8
```
cckellogg, line 16: Curious why is the mod and sum file changing? Can these changes be done in a separate PR?

pulsar/internal/commands.go
```diff
@@ -269,6 +270,117 @@ func serializeBatch(wb Buffer,
...
+	// Write checksum at created checksum-placeholder
+	frameEndIdx := wb.WriterIndex()
+	checksum := Crc32cCheckSum(wb.Get(metadataStartIdx, frameEndIdx-metadataStartIdx))
+
+	// Set Sizes and checksum in the fixed-size header
+	wb.PutUint32(frameEndIdx-frameStartIdx, frameSizeIdx) // External frame
+	wb.PutUint32(checksum, checksumIdx)
+}
+
+func encryptPayload(msgMetadata *pb.MessageMetadata,
```
cckellogg, line 347: Should this return an error?

pulsar/internal/commands.go
```diff
@@ -269,6 +270,117 @@ func serializeBatch(wb Buffer,
...
+	encryptedPayload := encryptPayload(msgMetadata,
+		msgCrypto,
+		KeyReader,
+		encryptionKeys,
+		compressedPayload,
+		cryptoFailureAction)
+
+	// there was a error in encrypting the payload and
+	// crypto failure action is set to crypto.ProducerCryptoFailureActionFail
+	if encryptedPayload == nil {
```
cckellogg, line 299: I'm not sure panic is the correct behavior here.

pulsar/internal/commands.go
```diff
@@ -269,6 +270,117 @@ func serializeBatch(wb Buffer,
...
+		return nil
+	}
+
+	// encrypt payload
+	encryptedPayload, err := msgCrypto.Encrypt(encryptionKeys,
+		KeyReader,
+		crypto.NewMessageMetadataSupplier(msgMetadata),
+		compressedPayload)
+
+	if err != nil {
```
cckellogg, line 371: Should this error be log somewhere?

pulsar/internal/commands.go
```diff
@@ -269,6 +270,117 @@ func serializeBatch(wb Buffer,
...
+	KeyReader crypto.KeyReader,
+	encryptionKeys []string,
+	msgCrypto crypto.MessageCrypto,
+	cryptoFailureAction int,
+) {
+	// Wire format
+	// [TOTAL_SIZE] [CMD_SIZE][CMD] [MAGIC_NUMBER][CHECKSUM] [METADATA_SIZE][METADATA] [PAYLOAD]
+
+	// compress the payload
+	compressedPayload := compressionProvider.Compress(nil, uncompressedPayload.ReadableSlice())
```
cckellogg, line 288: We want to compress before encrypting?

pulsar/producer_partition.go
```diff
@@ -148,6 +179,26 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
...
+	for t := range p.dataKeyTicker.C {
+		p.log.Infof("Refreshing data key :%v", t)
+		err := p.updateDataKey()
+		if err != nil {
+			p.log.Errorf("Error refreshing data key : %v", err)
+		}
+	}
+}
+
+func (p *partitionProducer) updateDataKey() error {
```
cckellogg, line 192: Does this need to be thread safe? Also, can we move the key refreshing to another MR.

pulsar/producer_partition.go
```diff
@@ -197,30 +248,32 @@ func (p *partitionProducer) grabCnx() error {
...
-			compression.Level(p.options.CompressionLevel),
-			p,
-			p.log)
-		if err != nil {
-			return err
-		}
+		provider, _ = GetBatcherBuilderProvider(DefaultBatchBuilder)
+
 	} else if p.batchBuilder == nil {
-		provider, err := GetBatcherBuilderProvider(p.options.BatcherBuilderType)
```
cckellogg, line 211: Why change this? The `p` masks the partitionProducer and make the code harder to follow.

pulsar/internal/batch_builder.go
```diff
@@ -128,24 +137,56 @@ func newBatchContainer(
...
 	bc := newBatchContainer(
 		maxMessages, maxBatchSize, producerName, producerID, compressionType,
-		level, bufferPool, logger,
+		level, bufferPool, logger, options...,
 	)
 
 	return &bc, nil
 }
 
+// UseEncryptionKeys encryption key names to use
```
cckellogg, line 162: What about removing these config options and passing an Encryptor interface/struct. I think the config functions work better when things are scoped in their own package - it make it name spacing clearer. Also, it doesn't really match the rest of the code.
```
type Encrypter interface {
 Encrypt(msgMetadata, payload []byte) ([]byte, error)
}
```
If encryption is not provided there can be a noop encryter

pulsar/internal/commands.go
GPrabhudas: As per java implementation => Yes compress and the encrypt.

pulsar/producer_partition.go
GPrabhudas: I can move it to next MR. 
Just ask:  why this to be moved to another MR?

go.mod
GPrabhudas: Let me recheck again :)

pulsar/producer.go
GPrabhudas: Make sense

pulsar/internal/commands.go
GPrabhudas: I preferred to put this in separate function instead of modifying the exiting one. Let me check if I can remove duplicate code

pulsar/internal/batch_builder.go
GPrabhudas: Got it.

pulsar/producer_partition.go
GPrabhudas: Just tried to remove duplicate code :) If it is not ok, I'll revert it.

wolfstudy:
@cckellogg @GPrabhudas Any update for this? We are ready to initiate a new version of the release, hoping to include this new change in the new version (because some of the code has been merged in), can we consider advancing the progress here.

GPrabhudas:
> @cckellogg @GPrabhudas Any update for this? We are ready to initiate a new version of the release, hoping to include this new change in the new version (because some of the code has been merged in), can we consider advancing the progress here.

I'll work on it and address the review suggestions.

wolfstudy:
> @wolfstudy what is your timeframe for a new release?

What I want is to start the release process after merging the change related to this function.

cckellogg:
> > @wolfstudy what is your timeframe for a new release?
> 
> What I want is to start the release process after merging the change related to this function. If the development of this feature takes some time, we can consider including it in the next version.

There are additional MRs (besides this one) to get this feature working. Currently, just the crypto package has been merged. It's probably better to cut a release now and then add the crypto implementation (producer and consumer) into the next release. To me that is better than rushing to get MRs in.

pulsar/internal/batch_builder.go
GPrabhudas: Done

pulsar/producer_partition.go
GPrabhudas: reverted

pulsar/producer_partition.go
GPrabhudas: done

pulsar/internal/commands.go
GPrabhudas: done

pulsar/internal/commands.go
GPrabhudas: Its better to throw an exception because client is setting `ProducerCryptoFailureActionFail` to fail.

go.mod
GPrabhudas: synced with master branch

pulsar/internal/commands.go
GPrabhudas: reused the `serializeBatch `

pulsar/internal/crypto/producer_encryptor.go
```diff
@@ -0,0 +1,78 @@
...
+		messageCrypto:               messageCrypto,
+		logger:                      logger,
+		producerCryptoFailureAction: producerCryptoFailureAction,
+	}
+}
+
+// Encrypt producer encryptor
+func (e *producerEncryptor) Encrypt(payload []byte, msgMetadata crypto.MessageMetadataSupplier) ([]byte, error) {
+	// encryption is enabled but KeyReader interface is not implemented
+	if e.keyReader == nil {
```
cckellogg, line 52: Should this be detected and an error raised while setting up the producer?

pulsar/producer_partition.go
```diff
@@ -126,6 +129,23 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		p.producerName = options.Name
 	}
 
+	encryption := options.Encryption
+	// add default message crypto if not provided
+	if encryption != nil && len(encryption.Keys) > 0 && encryption.MessageCrypto == nil {
+		logCtx := fmt.Sprintf("[%v] [%v] [%v]", p.topic, p.producerName, p.producerID)
+		messageCrypto, err := crypto.NewDefaultMessageCrypto(logCtx, true, logger)
+		if err != nil {
+			logger.WithError(err).Error("Unable to get MessageCrypto instance. Producer creation is abandoned")
```
cckellogg, line 138: Will there be more context in the err of why this failed?

pulsar/encryption.go
```diff
@@ -0,0 +1,36 @@
...
+// under the License.
+
+package pulsar
+
+import "github.com/apache/pulsar-client-go/pulsar/crypto"
+
+// ProducerEncryptionInfo encryption related fields required by the producer
+type ProducerEncryptionInfo struct {
+	// KeyReader read RSA public/private key pairs
+	Keyreader crypto.KeyReader
```
cckellogg, line 25: `KeyReader`

pulsar/internal/crypto/encryptor.go
```diff
@@ -0,0 +1,33 @@
...
+package crypto
+
+import (
+	"github.com/apache/pulsar-client-go/pulsar/crypto"
+	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
+)
+
+// Encryptor support encryption
+type Encryptor interface {
+	Encrypt([]byte, crypto.MessageMetadataSupplier) ([]byte, error)
```
cckellogg, line 27: Since this is internal we can pass the *pb.MessageMetadata and avoid having to create a supplier for each message when encryption is not being used.

pulsar/internal/commands.go
```diff
@@ -221,9 +222,21 @@ func serializeBatch(wb Buffer,
...
 	// [TOTAL_SIZE] [CMD_SIZE][CMD] [MAGIC_NUMBER][CHECKSUM] [METADATA_SIZE][METADATA] [PAYLOAD]
+
+	// compress the payload
+	compressedPayload := compressionProvider.Compress(nil, uncompressedPayload.ReadableSlice())
+
+	// encrypt the compressed payload
+	encryptedPayload, err := encryptor.Encrypt(compressedPayload, crypto.NewMessageMetadataSupplier(msgMetadata))
+	if err != nil {
+		// error occurred while encrypting the payload, ProducerCryptoFailureAction is set to Fail
+		panic(fmt.Sprintf("Encryption of message failed, ProducerCryptoFailureAction is set to Fail. Error :%v", err))
```
cckellogg, line 237: What do other clients do here? This could crash a server.

pulsar/producer_partition.go
GPrabhudas: Yes. Error is returned if generation of data key is failed. Here we are logging the same error with other information.

pulsar/internal/commands.go
GPrabhudas: Java client throws Exception [PulsarClientException](https://github.com/apache/pulsar/blob/34a04b5fa4692cb4716df61e740795467f76451b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L679)

pulsar/internal/crypto/encryptor.go
GPrabhudas: done

pulsar/encryption.go
GPrabhudas: done

pulsar/internal/crypto/producer_encryptor.go
GPrabhudas: done

pulsar/producer_partition.go
```diff
@@ -781,6 +821,17 @@ func (p *partitionProducer) internalClose(req *closeProducer) {
 	p.batchFlushTicker.Stop()
 }
 
+func (p *partitionProducer) generateDataKey() error {
+	if p.options.Encryption != nil {
+		if p.options.Encryption.KeyReader != nil {
+			return p.options.Encryption.MessageCrypto.AddPublicKeyCipher(p.options.Encryption.Keys,
```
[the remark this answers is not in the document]
GPrabhudas: No. It won't be nil.
If not provided, it is getting initialized with default one.
https://github.com/Fanatics/pulsar-client-go/blob/encryption-support-ext-producer/pulsar/producer_partition.go#L141

pulsar/producer_partition.go
```diff
@@ -126,6 +129,29 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
...
 
+	encryption := options.Encryption
+	// add default message crypto if not provided
+	if encryption != nil && len(encryption.Keys) > 0 {
+		if encryption.KeyReader == nil {
+			return nil, fmt.Errorf("encryption is enabled, KeyReader can not be nil")
+		}
+
+		if encryption.MessageCrypto == nil {
+			logCtx := fmt.Sprintf("[%v] [%v] [%v]", p.topic, p.producerName, p.producerID)
```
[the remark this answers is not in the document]
GPrabhudas: Sure

pulsar/internal/crypto/producer_encryptor.go
```diff
@@ -0,0 +1,72 @@
...
+		payload)
+
+	// error encryping the payload
+	if err != nil {
+		// error occurred in encrypting the payload
+		// crypto ProducerCryptoFailureAction is set to send
+		// send unencrypted message
+		if e.producerCryptoFailureAction == crypto.ProducerCryptoFailureActionSend {
+			e.logger.Errorf("Encryption of payload failed : %v", err)
+			e.logger.Warn("ProducerCryptoFailureAction is set to send, sending unecrypted message")
```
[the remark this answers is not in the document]
GPrabhudas: done.

pulsar/internal/commands.go
GPrabhudas: It depends on end user if they are catching the exception. Since it is java, it is possible to catch the exception.

I think it is proper action to `panic` here because  `ProducerCryptoFailureAction=Fail` set by the user in config. 

I do see other panics in the same function [serializeBatch](https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/commands.go#L220) and if they are recoverable then this panic can also be recovered.

jerrypeng: In java and c++, only the message is marked as failed to send and will NOT explicitly crash the application.  We should follow the same semantics here.

GPrabhudas: Thanks @jerrypeng  @cckellogg  Yes, I agree It is better to not crash the application.

But I'm just trying to understand how java client do not crash the application.

I found below steps where in it throws exception and possibly crash the application.

- [ProducerBase send](https://github.com/apache/pulsar/blob/34a04b5fa4692cb4716df61e740795467f76451b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerBase.java#L63)
- [TypedMessaggeBuilderImpl send](https://github.com/apache/pulsar/blob/34a04b5fa4692cb4716df61e740795467f76451b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TypedMessageBuilderImpl.java#L91)
- [ProducerImpl completable future](https://github.com/apache/pulsar/blob/34a04b5fa4692cb4716df61e740795467f76451b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L466)
- [ProducerImpl completing future with exception](https://github.com/apache/pulsar/blob/34a04b5fa4692cb4716df61e740795467f76451b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L318)
- [ProducerImpl encryptMessage](https://github.com/apache/pulsar/blob/34a04b5fa4692cb4716df61e740795467f76451b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L679)

GPrabhudas: And when compared  with java client , I don't see any existing way to mark message as failed in go client. 
So need suggestions  on achieving similar thing in go client.

cckellogg: This code will check if the message was added and if not it returns an error through the callback.
https://github.com/apache/pulsar-client-go/blob/master/pulsar/producer_partition.go#L434

I think we should we should look into changing the BatchBuilder Add message to return an error as opposed to a bool (this could be a separate MR). (https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/batch_builder.go#L47). There will be a little refactoring for some of the functions. If this returned an error then we could fail the message and return that to the client with a meaningful message. The bool will still work it's just the message returned is not very helpful.

pulsar/internal/key_based_batch_builder.go
```diff
@@ -203,21 +205,23 @@ func (bc *keyBasedBatchContainer) FlushBatches() (
 	sort.Strings(sortedKeys)
 	for _, k := range sortedKeys {
 		container := bc.batches.containers[k]
-		b, s, c := container.Flush()
+		b, s, c, err := container.Flush()
 		if b != nil {
 			batchesData[idx] = b
 			sequenceIDs[idx] = s
 			callbacks[idx] = c
+		} else if err != nil {
```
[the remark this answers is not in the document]
GPrabhudas: @cckellogg can this PR me merged if all looks good :)

cckellogg (approved):

PR #561, opened 2021-07-12
Fix possible race condition in connection pool

cckellogg:
- Use mutex instead of sync map in connection pool
- Create tickers in run function for the connection instead
of when the connection is created


wolfstudy (approved):
LGTM +1

PR #566, opened 2021-07-13
Add send error logic for connection

### LINKED ISSUE #564: Support handleSendError command in Go SDK

When the broker is processing the connection failure, it will throw a ServerError to the client. The current Go SDK does not handle this scenario, and we need to add the corresponding code logic for this

In `commands.go`, add this logic:

```
	case pb.BaseCommand_SEND_ERROR:
		cmd.SendError = msg.(*pb.CommandSendError)
```

wolfstudy:

Fixes #564 

### Motivation

Add command of SendError logic for connection

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -751,7 +751,6 @@ func (c *connection) handleSendError(sendError *pb.CommandSendError, cmdError *p
 	requestID := cmdError.GetRequestId()
 	producerID := sendError.GetProducerId()
 
-SendError:
 	switch *sendError.Error {
 	case pb.ServerError_NotAllowedError:
 		c.pendingLock.Lock()
@@ -768,24 +767,21 @@ SendError:
 
 		errMsg := fmt.Sprintf("server error: %s: %s", cmdError.GetError(), cmdError.GetMessage())
 		request.callback(nil, errors.New(errMsg))
-		break SendError
 	case pb.ServerError_TopicTerminatedError:
+		// TODO: no-op
+	default:
+		// By default, for transient error, let the reconnection logic
+		// to take place and re-establish the produce again
 		c.listenersLock.RLock()
 		producer, ok := c.listeners[producerID]
 		c.listenersLock.RUnlock()
-
 		if ok {
 			producer.ConnectionClosed()
 		} else {
 			c.log.
 				WithField("producerID", producerID).
-				Warn("[HandleSendError] connection closed")
+				Warn("[handleSendError] reconnection to broker.")
 		}
-		break SendError
-	default:
-		// By default, for transient error, let the reconnection logic
-		// to take place and re-establish the produce again
-		c.Close()
 	}
 }
 
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -536,7 +536,7 @@ func (c *connection) internalReceivedCommand(cmd *pb.BaseCommand, headersAndPayl
 		c.handleResponseError(cmd.GetError())
 
 	case pb.BaseCommand_SEND_ERROR:
-		c.handleSendError(cmd.GetSendError(), cmd.GetError())
+		c.handleSendError(cmd.GetError())
 
 	case pb.BaseCommand_CLOSE_PRODUCER:
 		c.handleCloseProducer(cmd.GetCloseProducer())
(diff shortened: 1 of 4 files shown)


wolfstudy:
ping @zymap PTAL

pulsar/internal/connection.go
```diff
@@ -744,6 +745,50 @@ func (c *connection) handleAuthChallenge(authChallenge *pb.CommandAuthChallenge)
...
+		break SendError
+	case pb.ServerError_TopicTerminatedError:
+		c.listenersLock.RLock()
+		producer, ok := c.listeners[producerID]
+		c.listenersLock.RUnlock()
+
+		if ok {
+			producer.ConnectionClosed()
+		} else {
+			c.log.
```
zymap, line 780: wrong indentation?

pulsar/internal/connection.go
```diff
@@ -744,6 +745,50 @@ func (c *connection) handleAuthChallenge(authChallenge *pb.CommandAuthChallenge)
...
+		errMsg := fmt.Sprintf("server error: %s: %s", cmdError.GetError(), cmdError.GetMessage())
+		request.callback(nil, errors.New(errMsg))
+		break SendError
+	case pb.ServerError_TopicTerminatedError:
+		c.listenersLock.RLock()
+		producer, ok := c.listeners[producerID]
+		c.listenersLock.RUnlock()
+
+		if ok {
+			producer.ConnectionClosed()
```
zymap, line 778: In the Java implementation, we change the state to terminate and send the error to all pending messages, does the producer connection closed will do the same thing with that?

pulsar/internal/connection.go
```diff
@@ -744,6 +745,50 @@ func (c *connection) handleAuthChallenge(authChallenge *pb.CommandAuthChallenge)
...
 }
 
+func (c *connection) handleSendError(sendError *pb.CommandSendError, cmdError *pb.CommandError) {
+	c.log.Warnf("Received send error from server: [%v] : [%s]", sendError.GetError(), sendError.GetMessage())
+
+	requestID := cmdError.GetRequestId()
+	producerID := sendError.GetProducerId()
+
+SendError:
+	switch *sendError.Error {
```
zymap, line 755: We also handle the `ChecksumError` in the java implementation, do we need to implement that?

pulsar/internal/connection.go
```diff
@@ -744,6 +745,50 @@ func (c *connection) handleAuthChallenge(authChallenge *pb.CommandAuthChallenge)
...
+
+		if ok {
+			producer.ConnectionClosed()
+		} else {
+			c.log.
+				WithField("producerID", producerID).
+				Warn("[HandleSendError] connection closed")
+		}
+		break SendError
+	default:
```
cckellogg, line 785: Is the pending request callback not needed in the default case?

cckellogg (commented):
How do the other clients handle this and is the correct behavior (send signal to producer, close producer, close connection, etc..)?

pulsar/internal/connection.go
wolfstudy: No, this is automatically formatted by golanglint, probably for aesthetic reasons

pulsar/internal/connection.go
wolfstudy: This will be supported in the next pr

pulsar/internal/connection.go
wolfstudy: Nothing here is to send a connection closed signal to runEventsLoop, try to trigger the logic of reconnection

wolfstudy:
@zymap @cckellogg 

There is currently no good idea to deal with TopicTerminatedError, so it is temporarily not supported.

pulsar/internal/pulsartracing/producer_interceptor.go
```diff
@@ -33,7 +33,9 @@ func (t *ProducerInterceptor) BeforeSend(producer pulsar.Producer, message *puls
 	buildAndInjectSpan(message, producer).Finish()
 }
 
-func (t *ProducerInterceptor) OnSendAcknowledgement(producer pulsar.Producer, message *pulsar.ProducerMessage, msgID pulsar.MessageID) {
+func (t *ProducerInterceptor) OnSendAcknowledgement(producer pulsar.Producer,
```
cckellogg, line 36: Can we revert these formatting changes since they have nothing to do with this change?

pulsar/internal/connection.go
cckellogg: The java client closes the connection in the default case.

https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java#L649

Should we do the same. What should the expected client behavior be here?

pulsar/internal/connection.go
```diff
@@ -744,6 +745,46 @@ func (c *connection) handleAuthChallenge(authChallenge *pb.CommandAuthChallenge)
...
 
+func (c *connection) handleSendError(sendError *pb.CommandSendError, cmdError *pb.CommandError) {
+	c.log.Warnf("Received send error from server: [%v] : [%s]", sendError.GetError(), sendError.GetMessage())
+
+	requestID := cmdError.GetRequestId()
+	producerID := sendError.GetProducerId()
+
+	switch *sendError.Error {
+	case pb.ServerError_NotAllowedError:
+		c.pendingLock.Lock()
```
cckellogg, line 756: This block of code is used in a couple of places now maybe we should add a function?  Something like this?

```
func deletePendingRequest(requestID) (request, bool) {
   c.pendingLock.Lock()
   defer c.pendingLock.Unlock()
   request, ok := c.pendingReqs[requestID]
    if ok {
	delete(c.pendingReqs, requestID)
   }
   return request, ok
		
}
```

pulsar/internal/pulsartracing/producer_interceptor.go
wolfstudy: This will cause Action CI to fail:

```
Error: pulsar/internal/pulsartracing/producer_interceptor.go:36: line is 136 characters (lll)
func (t *ProducerInterceptor) OnSendAcknowledgement(producer pulsar.Producer, message *pulsar.ProducerMessage, msgID pulsar.MessageID) {
Error: pulsar/internal/pulsartracing/producer_interceptor_test.go:58: line is 132 characters (lll)
func (p *mockProducer) SendAsync(context.Context, *pulsar.ProducerMessage, func(pulsar.MessageID, *pulsar.ProducerMessage, error)) {
Error: Process completed with exit code 1.
```

pulsar/internal/connection.go
wolfstudy: Good ideas for this.

wolfstudy: > The java client closes the connection in the default case.
> 
> https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java#L649
> 
> Should we do the same. What should the expected client behavior be here?

About this, Look at the comments like reconnect, look at the code like close the current connection:

```
            // By default, for transient error, let the **reconnection** logic
            // to take place and re-establish the produce again
            ctx.close();
```

zymap (approved):

PR #579, opened 2021-07-30
Ignore partitions shrink in client side

wuYin:
### Issue
This happened in out production env suddenly without any admin operation.

image

### Modifications

Since partitions shrink is not client's fault, we just ignore it temporarily, I'll try to find cause in server side.

### Verifying this change

### DIFF
--- a/pulsar/producer_impl.go
+++ b/pulsar/producer_impl.go
@@ -177,7 +177,14 @@ func (p *producer) internalCreatePartitionsProducers() error {
 
 	if oldProducers != nil {
 		oldNumPartitions = len(oldProducers)
-		if oldNumPartitions == newNumPartitions {
+		if newNumPartitions < oldNumPartitions {
+			p.log.WithFields(log.Fields{
+				"old_partitions": oldNumPartitions,
+				"new_partitions": newNumPartitions,
+			}).Error("Number of partitions shrank, ignored.")
+			// since it's not client's fault, we just ignore it temporarily
+			return nil
+		} else if newNumPartitions == oldNumPartitions {
 			p.log.Debug("Number of partitions in topic has not changed")
 			return nil
 		}


pulsar/producer_impl.go
```diff
@@ -177,7 +177,14 @@ func (p *producer) internalCreatePartitionsProducers() error {
 
 	if oldProducers != nil {
 		oldNumPartitions = len(oldProducers)
-		if oldNumPartitions == newNumPartitions {
+		if newNumPartitions < oldNumPartitions {
```
cckellogg, line 180: I didn't think it was possible for the number of partitions to decrease. What the topic deleted or did the broker return the wrong number of partitions? Any logs on the broker side to indicate that? If this is ignored will the next check fix it?

wuYin: Thank you for reply.
- There is no forced topic deletion or other admin operation this day.
- [brokerDeleteInactiveTopicsEnabled](https://github.com/apache/pulsar/blob/master/conf/broker.conf#L156) has configured as false.
- proxy and broker logs were checked, and there is not much valuable information, most is INFO level messages.
- only 20% producers has this error, it's weird.

maybe there is a concurrency issue in broker ownership cache that leads to data corruption?

In fact, client no need to do this detection, but the server needs to locate and solve the decrease problem.

pulsar/producer_impl.go
```diff
@@ -177,7 +177,14 @@ func (p *producer) internalCreatePartitionsProducers() error {
...
+			p.log.WithFields(log.Fields{
+				"old_partitions": oldNumPartitions,
+				"new_partitions": newNumPartitions,
+			}).Error("Number of partitions shrank, ignored.")
```
cckellogg, line 184: Maybe add a little more context in the message something like this? `Unexpected response from get partitions the number of partitions decreased`. 

By ignoring this do you see any potential issues?

PR #582, opened 2021-08-06
add epoch to handle create producer timeout

freeznet:
Master Issue: https://github.com/apache/pulsar/issues/5606

### Motivation

Handle create producer timeout on pulsar-client-go. To prevent duplicate producer when broker is not available.

### Modifications

- add `epoch` to `partitionProducer`
- add `userProvidedProducerName` to `partitionProducer`
- add tests

### Verifying this change

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -80,8 +80,7 @@ type partitionProducer struct {
 	partitionIdx     int32
 	metrics          *internal.TopicMetrics
 
-	epoch                    uint64
-	userProvidedProducerName bool
+	epoch uint64
 }
 
 func newPartitionProducer(client *client, topic string, options *ProducerOptions, partitionIdx int,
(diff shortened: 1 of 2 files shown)


codelipenghui (approved):

pulsar/producer_partition.go
```diff
@@ -79,6 +79,9 @@ type partitionProducer struct {
 	schemaInfo       *SchemaInfo
 	partitionIdx     int32
 	metrics          *internal.TopicMetrics
+
+	epoch                    uint64
+	userProvidedProducerName bool
```
cckellogg, line 84: Can this `userProvidedProducerName`  just be determined when creating the producer command from the name? Why create an extra variable?

pulsar/producer_partition.go
```diff
@@ -136,6 +142,7 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 	p.log = p.log.SubLogger(log.Fields{
 		"producer_name": p.producerName,
 		"producerID":    p.producerID,
+		"epoch":         atomic.LoadUint64(&p.epoch),
```
cckellogg, line 145: Will this always log the same the same value?

freeznet:
@cckellogg thanks for your comments, I have addressed your comments in the latest commit, PTAL when you have time.

pulsar/producer_partition.go
```diff
@@ -298,7 +306,8 @@ func (p *partitionProducer) reconnectToBroker() {
 		d := backoff.Next()
 		p.log.Info("Reconnecting to broker in ", d)
 		time.Sleep(d)
-
+		atomic.AddUint64(&p.epoch, 1)
+		p.log.WithField("epoch", atomic.LoadUint64(&p.epoch))
```
cckellogg, line 310: The epoch field is going to log the same value every time (the first value). Can epoch just be logged after the producer is created? Updating the field value requires creating a new logger and doing this would require locking to access the log which is probably not what we want.

freeznet: I see, just removed the epoch from the logger and add a one-line debug log to print the current epoch value.

pulsar/producer_partition.go
```diff
@@ -298,7 +305,9 @@ func (p *partitionProducer) reconnectToBroker() {
 		d := backoff.Next()
 		p.log.Info("Reconnecting to broker in ", d)
 		time.Sleep(d)
-
+		atomic.AddUint64(&p.epoch, 1)
+		p.log.WithField("epoch", atomic.LoadUint64(&p.epoch)).Debug(
```
cckellogg, line 309: This will log the epoch twice once as a field and once as a message. It might make more sense to add the epoch log after the producer is created? Curious what's the significance using and knowing the epoch?

freeznet: Updated, the reason to log the epoch is mainly for debug.

merlimat (approved):

cckellogg (approved):

PR #604, opened 2021-08-28
initial commit

### LINKED ISSUE #500: Make Topic Level Prometheus Metrics Optional
In `0.4.0`, https://github.com/apache/pulsar-client-go/pull/410 introduced topic level prometheus metrics. When consuming from or publishing to many topics, this creates many metrics and can cause problems for applications and/or prometheus instances. In my case, I saw up to 900k metrics produced by a single client, which caused memory problems in my application and temporarily brought down my prometheus server.

I would like to be able to enable or disable the topic level metrics. I think these topic level metrics should be disabled by default, since that was the previous behavior.

I looked into the code introduced in `0.4.0`, and I don't see any ways to prevent these metrics from being created currently.

**Additional context**
I like the metrics I get out of the client in `0.3.0`. I think it's great to have them out of the box.

Also, the cardinality of metrics is preventing me from upgrading my go client from `0.3.0` to `0.4.0`.

flowchartsman:
# Draft
No test exercise yet, but passes docker-ci.sh tests.

I am a bit dissatisfied with the use of the const/iota enum in the public package and the integers in the backend, but it's necessary to avoid import cycles. Otherwise the enum needs to be moved out of `pulsar` and placed into another package, or the enum needs to be mirrored in `internal`

Fixes #500

### Motivation

Allow selective metrics cardinality configuration

### Modifications

Allow log cardinality level to be set to none, tenant, namespace or topic.


flowchartsman:
Paging @merlimat for a quick sanity check on the WIP, since they did initial implementation.

pulsar/internal/metrics.go
```diff
@@ -21,9 +21,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-var topicLabelNames = []string{"pulsar_tenant", "pulsar_namespace", "topic"}
-
 type Metrics struct {
+	metricsLevel      int
```
merlimat, line 25: Would it make sense to define a 2nd enum for internal use instead of ints?

merlimat (commented):
Nice!

pulsar/internal/metrics.go
flowchartsman: It might, but that’s definitely a hack around some layout problems. But there probably aren’t any half-measures there, save an awkward new package, so I guess that’s a ticket for another day.

merlimat (approved):

pulsar/internal/metrics.go
flowchartsman: One option could be to make the consts be strings instead of integers, and then have `internal` do a switch on these. Then at least the hard-coded values would have more meaning than inscrutable integer values.

PR #610, opened 2021-09-03
Add reconsumer level later logic

wolfstudy:

### Motivation

As https://github.com/apache/pulsar/issues/11892 show, add reconsumer level later logic

### Modifications

- Add `ReconsumerLaterLevel()` logic
- Add `ReconsumerLaterLevelAsync()` logic

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -117,6 +117,9 @@ type ConsumerOptions struct {
 	// Default is false
 	RetryEnable bool
 
+	// Configuration for Retry delay level policy
+	DelayLevelUtil DelayLevelUtil
+
 	// Sets a `MessageChannel` for the consumer
 	// When a message is received, it will be pushed to the channel for consumption
 	MessageChannel chan ConsumerMessage
@@ -180,6 +183,14 @@ type Consumer interface {
 
 	// ReconsumeLater mark a message for redelivery after custom delay
 	ReconsumeLater(msg Message, delay time.Duration)
+
+	// Reconsume a message
+	ReconsumeLaterLevel(message Message, options ReconsumeOptions) error
+
+	// Reconsume a message Async
+	ReconsumeLaterLevelAsync(message Message,
+		options ReconsumeOptions,
+		callback func(MessageID, *ProducerMessage, error))
 
 	// Acknowledge the failure to process a single message.
 	//
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"strconv"
@@ -126,6 +127,10 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 			options.Topic = ""
 		} else if options.Topic == "" && len(options.Topics) > 0 {
 			options.Topics = append(options.Topics, options.DLQ.RetryLetterTopic)
+		}
+
+		if options.DelayLevelUtil == nil {
+			options.DelayLevelUtil = NewDelayLevelUtil(DefaultMessageDelayLevel)
 		}
 	}
 
@@ -244,6 +249,55 @@ func (c *consumer) runBackgroundPartitionDiscovery(period time.Duration) (cancel
 		close(stopDiscoveryCh)
 		wg.Wait()
 	}
+}
+
+func (c *consumer) ReconsumeLaterLevel(message Message, reconsumeOptions ReconsumeOptions) error {
+	if !c.options.RetryEnable {
+		return errors.New("This Consumer config retry disabled. ")
+	}
+	topicName, err := internal.ParseTopicName(message.Topic())
+	if err != nil {
+		return err
+	}
+	index := 0
+	if topicName.Partition >= 0 {
+		index = topicName.Partition
+	}
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	prod, prodMsg, desType, err := c.consumers[index].internalBeforeReconsume(message, reconsumeOptions)
+	if err != nil {
+		return err
+	}
+
+	c.consumers[index].internalReconsumeAsync(prod, message, prodMsg, desType, func(id MessageID, producerMessage *ProducerMessage, e error) {
+		err = e
+		wg.Done()
+	})
+	wg.Wait()
+	return err
+}
+
+func (c *consumer) ReconsumeLaterLevelAsync(message Message, reconsumeOptions ReconsumeOptions, callback func(MessageID, *ProducerMessage, error)) {
+	if !c.options.RetryEnable {
+		c.log.Error("This Consumer config retry disabled. ")
+		return
+	}
+	topicName, err := internal.ParseTopicName(message.Topic())
+	if err != nil {
+		c.log.Error(err)
+		return
+	}
+	index := 0
+	if topicName.Partition >= 0 {
+		index = topicName.Partition
+	}
+	prod, prodMsg, desType, err := c.consumers[index].internalBeforeReconsume(message, reconsumeOptions)
+	if err != nil {
+		c.log.Error(err)
+		return
+	}
+	c.consumers[index].internalReconsumeAsync(prod, message, prodMsg, desType, callback)
 }
 
 func (c *consumer) internalTopicSubscribeToPartitions() error {
--- a/pulsar/consumer_multitopic.go
+++ b/pulsar/consumer_multitopic.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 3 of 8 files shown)


pulsar/consumer.go
```diff
@@ -117,6 +117,9 @@ type ConsumerOptions struct {
 	// Default is false
 	RetryEnable bool
 
+	// Configuration for Retry delay level policy
+	DelayLevelUtil DelayLevelUtil
```
cckellogg, line 121: What is a delay level util? Why does a util need to be exposed as a public API?

pulsar/consumer_partition.go
```diff
@@ -271,6 +275,113 @@ func (pc *partitionConsumer) internalGetLastMessageID(req *getLastMsgIDRequest)
...
+		delayLevels = pc.options.delayLevelUtil.GetMaxDelayLevel()
+	}
+
+	if v, ok := propertiesMap["RECONSUMETIMES"]; ok {
+		reconsumeTimesUint64, err := strconv.ParseUint(v, 10, 64)
+		if err != nil {
+			return nil, nil, "", err
+		}
+		reconsumeTimes = uint32(reconsumeTimesUint64) + 1
+		propertiesMap["RECONSUMETIMES"] = strconv.FormatUint(uint64(reconsumeTimes), 10)
```
cckellogg, line 325: Maybe add constants for these keys?

pulsar/reconsume_later_util.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+	"strconv"
+	"strings"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const DefaultMessageDelayLevel = "1s 5s 10s 30s 1m 2m 3m 4m 5m 6m 7m 8m 9m 10m 20m 30m 1h 2h"
```
cckellogg, line 28: should this be private? why is this a string?

pulsar/reconsume_later_util.go
```diff
@@ -0,0 +1,145 @@
...
+	} else if level < 1 {
+		return 0
+	} else if level > d.maxDelayLevel {
+		return d.delayLevelTable[d.maxDelayLevel]
+	} else {
+		return d.delayLevelTable[level]
+	}
+}
+
+func (d *delayLevelUtil) ParseDelayLevel() bool {
```
cckellogg, line 67: How does this leveling work?

pulsar/consumer.go
```diff
@@ -181,6 +184,14 @@ type Consumer interface {
 	// ReconsumeLater mark a message for redelivery after custom delay
 	ReconsumeLater(msg Message, delay time.Duration)
 
+	// Reconsume a message
+	ReconsumeLaterLevel(message Message, options ReconsumeOptions) error
+
+	// Reconsume a message Async
+	ReconsumeLaterLevelAsync(message Message,
+		options ReconsumeOptions,
+		callback func(MessageID, *ProducerMessage, error))
```
cckellogg, line 193: Why is there a ProducerMessage in the callback?

pulsar/consumer_impl.go
```diff
@@ -246,6 +251,57 @@ func (c *consumer) runBackgroundPartitionDiscovery(period time.Duration) (cancel
...
+	if err != nil {
+		return err
+	}
+
+	c.consumers[index].internalReconsumeAsync(prod, message, prodMsg, desType,
+		func(id MessageID, producerMessage *ProducerMessage, e error) {
+			err = e
+			wg.Done()
+		})
+	wg.Wait()
```
cckellogg, line 278: How long could this block for?

pulsar/consumer_multitopic.go
```diff
@@ -155,6 +156,46 @@ func (c *multiTopicConsumer) ReconsumeLater(msg Message, delay time.Duration) {
 	consumer.ReconsumeLater(msg, delay)
 }
 
+func (c *multiTopicConsumer) ReconsumeLaterLevel(message Message, reconsumeOptions ReconsumeOptions) error {
+	if !c.options.RetryEnable {
+		return errors.New("[ReconsumeLater]This Consumer config retry disabled. ")
```
cckellogg, line 161: Let's remove [ReconsumeLater*] from logs and errors it's inconsistent with the rest of the code. Instead of `[ReconsumeLater]This Consumer config retry disabled. `  maybe `retries disabled for consumer unable to reconsume later`.

pulsar/consumer_partition.go
```diff
@@ -271,6 +275,113 @@ func (pc *partitionConsumer) internalGetLastMessageID(req *getLastMsgIDRequest)
...
+	}
+	prod := pc.dlq.producer
+	desType := "retry"
+	if reconsumeTimes > pc.dlq.policy.MaxDeliveries {
+		propertiesMap["REAL_TOPIC"] = pc.dlq.policy.DeadLetterTopic
+		prod = pc.dlq.getProducer() // Get dead topic producer
+		desType = "dead"
+	}
+	// Default use the retry letter topic producer
+	if prod == nil {
```
cckellogg, line 358: Should this check happen earlier?

pulsar/consumer_impl.go
```diff
@@ -246,6 +251,57 @@ func (c *consumer) runBackgroundPartitionDiscovery(period time.Duration) (cancel
...
+	if err != nil {
+		return err
+	}
+	index := 0
+	if topicName.Partition >= 0 {
+		index = topicName.Partition
+	}
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	prod, prodMsg, desType, err := c.consumers[index].internalBeforeReconsume(message, reconsumeOptions)
```
cckellogg, line 268: create a consumer variable and or pass it into the internal functions?

pulsar/consumer_multitopic.go
```diff
@@ -155,6 +156,46 @@ func (c *multiTopicConsumer) ReconsumeLater(msg Message, delay time.Duration) {
...
+		}
+	}
+	return errors.New("[ReconsumeLater]Topic not in multi topic consumer list. ")
+}
+
+func (c *multiTopicConsumer) ReconsumeLaterLevelAsync(message Message, reconsumeOptions ReconsumeOptions,
+	callback func(MessageID, *ProducerMessage, error)) {
+	if !c.options.RetryEnable {
+		c.log.Warn(errors.New("[ReconsumeLaterAsync]This Consumer config retry disabled. "))
+		return
```
cckellogg, line 181: Do callbacks need to happen before returning?

pulsar/consumer.go
wolfstudy: Because this needs to be exposed to users to customize the behavior of using ReconsumeLater

pulsar/consumer_partition.go
wolfstudy: Good idea! will fix

pulsar/reconsume_later_util.go
wolfstudy: > should this be private?

yes, It should be more reasonable to be private. The definition of string here is OK, as long as we parse and process according to specific rules.

pulsar/consumer.go
wolfstudy: When doing message sending, we need it to process the field information such as Payload and Properties

wolfstudy:
```
producerMsg := &ProducerMessage{
		Payload:             msg.Payload(),
		Properties:          propertiesMap,
		ReplicationClusters: msg.(*message).replicationClusters,
	}
```

pulsar/consumer_multitopic.go
wolfstudy: cool, will fix

pulsar/consumer.go
cckellogg: What is an example of how a user can customize this? What are the different parameters? This name is very confusing to me and doesn't provide must information on what it does.

pulsar/reconsume_later_util.go
cckellogg: Why is the definition a string? Is this documented somewhere or do other clients do this? This seems error phone and complicated.

pulsar/reconsume_later_util.go
```diff
@@ -0,0 +1,145 @@
...
+	d.delayLevelTable = make(map[int]int64)
+	timeUnitTable := make(map[string]int64)
+	timeUnitTable["s"] = 1000
```
cckellogg, line 70: There are constants and helpers in the language that do all of these conversions.

cckellogg (commented):
Will you provide some more context and or details on how this works? 

What is leveling? 

Why are times encoded in a string and then parsed?

We should think about a good clean api/design that is easy to use and understand.

cckellogg (changes requested):
I think this should be discussed more within the community with a design doc or PIP. This is changing behavior and adding new interfaces/features. Also, It seems like this can already be accomplished with some custom application code using the current `ReconsumeLater`?

wolfstudy:
OK, i will send a PIP for this change!

Issue #443, opened 2021-01-18
Add context param in interceptor


godchen0212:
**Is your feature request related to a problem? Please describe.**
I want to add opentracing tracking for message delivery. Executing opentracing related code in the interceptor is an elegant way which won't cause a bad influence on my logic code. However, the interceptor has no context parameter, which makes it difficult to pass in the relevant context.

**Describe the solution you'd like**
I want to add context param in ProducerInterceptor's BeforeSend method.
BeforeSend(producer Producer, message *ProducerMessage)
↓ 
BeforeSend(ctx context.Context, producer Producer, message *ProducerMessage)

davendu:
+1 for Golang client supporting context in conceptor. Besides tracing purpose, we internally use context to trace process of a message, and change properties based on some value. Currently we have to wrap the producer to modify Send & SendAsync. It can be better implemented if interceptor supports context.

#505 is closed since tracing is already supported in #518, but I think adding context is still necessary for Golang client. In other languages like Java and C++, execution context is implicitly shared (for example, using TLS variable). However, Golang requires explicit share of state, thus having `context.Context` type. For now, there is no way to share state (other than some [hack](https://github.com/jtolio/gls)) between caller and interceptor.

PR #613, opened 2021-09-08
[Issue 443] add context param in producer interceptor

GPrabhudas:
This PR address [issue-443](#443 )

Ability to add properties to message before sending using context param. 

Modified the producer interceptor methods to accept `context` parameter.

### DIFF
--- a/pulsar/producer_interceptor.go
+++ b/pulsar/producer_interceptor.go
@@ -25,7 +25,7 @@ type ProducerInterceptor interface {
 
 	// OnSendAcknowledgement This method is called when the message sent to the broker has been acknowledged,
 	// or when sending the message fails.
-	OnSendAcknowledgement(ctx context.Context, producer Producer, message *ProducerMessage, msgID MessageID)
+	OnSendAcknowledgement(producer Producer, message *ProducerMessage, msgID MessageID)
 }
 
 type ProducerInterceptors []ProducerInterceptor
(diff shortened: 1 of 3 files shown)


cckellogg (commented):
This is a breaking api change so we may need to figure out if that is ok or figure out if backwards compatibility is needed.

wolfstudy:
The Logic LGTM +1, as @cckellogg said, maybe we can add a new API for this change, e.g:

```
BeforeSendWithContext(ctx context.Context, producer Producer, message *ProducerMessage)
```

cckellogg:
> ```
> BeforeSendWithContext(ctx context.Context, producer Producer, message *ProducerMessage)
> ```
Adding a new function to the interface will still be a breaking change. Applications will have have to implement the new method and recompile. I think that's what we need to figure out. Do we need to support backward compatibility or is making a breaking api change ok for the next release.

wolfstudy:
> > ```
> > BeforeSendWithContext(ctx context.Context, producer Producer, message *ProducerMessage)
> > ```
> 
> Adding a new function to the interface will still be a breaking change. Applications will have have to implement the new method and recompile. I think that's what we need to figure out. Do we need to support backward compatibility or is making a breaking api change ok for the next release.

Yes, my mistake. It seems that adding context does have actual usage scenarios and can further improve the processing capabilities of the interceptor itself.

If we want to add this feature without breaking backward compatibility, can we try to directly expose a new interface to the user in ProducerOptions?

For exampels, in `ProducerOptions`:

```
// A chain of interceptors with context, These interceptors will be called at some points defined in ProducerInterceptorWithContext interface
InterceptorsWithContext ProducerInterceptorsWithContext
```

and introduce a new interface `ProducerInterceptorsWithContext`:

```
type ProducerInterceptorsWithContext interface {
	BeforeSendWithContext(ctx context.Context,producer Producer, message *ProducerMessage)

	OnSendAcknowledgement(producer Producer, message *ProducerMessage, msgID MessageID)
}
```

But it is not a very elegant way to achieve

liangyuanpeng (approved):
IMO,I agree with  making a breaking api change ok for the next release..

PR #617, opened 2021-09-13
Use the time wheel structure instead of traversing the map && Add ackTimeout function

EAHITechnology:


EAHITechnology:
Add a time wheel to optimize the delay logic.At the same time add ack timeout

wolfstudy (changes requested):
Thanks @EAHITechnology work for this.

`AckTimeout` is deliberately abandoned in the Go SDK, we can use Nack to replace the logic of AckTimeout

EAHITechnology:
> Thanks @EAHITechnology work for this.
> 
> `AckTimeout` is deliberately abandoned in the Go SDK, we can use Nack to replace the logic of AckTimeout

learn. Is the time wheel logic feasible?

EAHITechnology:
> Thanks @EAHITechnology work for this.
> 
> `AckTimeout` is deliberately abandoned in the Go SDK, we can use Nack to replace the logic of AckTimeout

Despite this, I still think acktimeout is very important. Like our project, nack is not the expected behavior when there is no way to restart

PR #618, opened 2021-09-14
Optimize negative_acks_tracker by time_wheel

EAHITechnology:
In the case that the task trigger time is relatively late, it is possible that the time complexity of traversing a map is O(n), but nothing is actually done, which leads to a waste of CPU time. The time wheel data structure can effectively reduce the order of magnitude of scanning. At the same time, this part of the logic can also be hidden behind the business logic in the internal.go

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -132,10 +132,6 @@ type ConsumerOptions struct {
 	// processed. Default is 1min. (See `Consumer.Nack()`)
 	NackRedeliveryDelay time.Duration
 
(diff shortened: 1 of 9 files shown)


cckellogg:
Curious how does this improve or enhance the current implementation? Why would users want to use this over what is currently there? What is the added overhead?

wolfstudy:
Thanks @EAHITechnology work for this.

**I am a little confused here, why do we need to use the time wheel to handle nack tasks?**

PR #627, opened 2021-09-29
Fix reconnection logic when topic is deleted

### LINKED ISSUE #623: When topic is deleted forced, don't to tigger reconnect logic
#### Expected behavior

When topic is deleted forced, under normal circumstances, we should force the producer to be deleted. What the client feels is that the producer has been closed, instead of triggering the logic of reconnection.

#### Actual behavior

When topic is deleted forced, the request triggers the logic of reconnection.

#### Steps to reproduce

1. Start standalone (e.g: 2.8.0) and set `allowAutoTopicCreation=false` in `standalone.conf`
2. Run the demo of producer:

```
func main() {
		URL: "pulsar://localhost:6650",
	})

	if err != nil {
		log.Fatal(err)
	}

	defer client.Close()

		Topic: "topic-1",
	})
	if err != nil {
		log.Fatal(err)
	}

	defer producer.Close()

	ctx := context.Background()

	for i := 0; i < 1000000; i++ {
		if msgId, err := producer.Send(ctx, &pulsar.ProducerMessage{
		}); err != nil {
			log.Fatal(err)
		} else {
			time.Sleep(1 * time.Second)
			log.Println("Published message: ", msgId)
		}
	}
}
```
3. Run the cmd to force delete `topic-1`

```
 bin/pulsar-admin topics delete topic-1 -f
```

4. Observing the behavior of this producer, we will find that we have triggered the reconnection logic.

image

The same code logic, we can see the following error message in java, this is the behavior we expect:

image

#### System configuration
**Pulsar version**: x.y

wolfstudy:

Fixes #623

### Motivation

As #623 said, when the topic is deleted forced, we don't should trying to reconnect, instead of giving up reconnection.

### Modifications

- Fix prodcuer reconnetion logic
- Fix consumer reconnection logic

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -893,13 +893,12 @@ func (pc *partitionConsumer) reconnectToBroker() {
 			// Successfully reconnected
 			pc.log.Info("Reconnected consumer to broker")
 			return
-		} else {
-			errMsg := err.Error()
-			if strings.Contains(errMsg, errTopicNotFount) {
-				// when topic is deleted, we should give up reconnection.
-				pc.log.Warn("Topic Not Found.")
-				break
-			}
+		}
+		errMsg := err.Error()
+		if strings.Contains(errMsg, errTopicNotFount) {
+			// when topic is deleted, we should give up reconnection.
+			pc.log.Warn("Topic Not Found.")
+			break
 		}
 
 		if maxRetry > 0 {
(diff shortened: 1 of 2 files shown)


wolfstudy:
image
This is the effect after the repair. When receiving the TopicNotFound error, we should jump out of the reconnection cycle, wait for sendTimeout, and automatically close the current producer process

zymap (approved):

Issue #658, opened 2021-11-02
Support nack backoff policy for Go SDK


wolfstudy:
In [PIP-106](https://github.com/apache/pulsar/issues/12379), we introduce the nack backoff policy and the java impl refer to https://github.com/apache/pulsar/pull/12566.

In Go SDK, we need to catch up the feature. The initial idea is to expose the new option in `ConsumerOptions`, as follows:

```
NackBackoffPolicy *NackBackoffPolicy
```

PR #660, opened 2021-11-03
Support nack backoff policy for SDK

wolfstudy:

Master Issue: #658

### Motivation

Support nack backoff policy

### Modifications

- Add `EnableDefaultNackBackoffPolicy` and `NackBackoffPolicy` options in ConsumerOptions.
- Modify the original implementation of the `Nack(Message)` interface
- Add test case
- Add defaultNackBackoffPolicy

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -158,6 +158,13 @@ type ConsumerOptions struct {
 
 	// Decryption decryption related fields to decrypt the encrypted message
 	Decryption *MessageDecryptionInfo
+
+	// NackBackoffPolicy is a redelivery backoff mechanism which we can achieve redelivery with different
+	// delays according to the number of times the message is retried.
+	//
+	// > Notice: the NackBackoffPolicy will not work with `consumer.NackID(MessageID)`
+	// > because we are not able to get the redeliveryCount from the message ID.
+	NackBackoffPolicy NackBackoffPolicy
 }
 
 // Consumer is an interface that abstracts behavior of Pulsar's consumer
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -159,6 +159,10 @@ type ConsumerOptions struct {
 	// Decryption decryption related fields to decrypt the encrypted message
 	Decryption *MessageDecryptionInfo
 
+	// If enabled, the default implementation of NackBackoffPolicy will be used to calculate the delay time of
+	// nack backoff, Default: false.
+	EnableDefaultNackBackoffPolicy bool
+
 	// NackBackoffPolicy is a redelivery backoff mechanism which we can achieve redelivery with different
 	// delays according to the number of times the message is retried.
 	//
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -35,6 +35,7 @@ const defaultNackRedeliveryDelay = 1 * time.Minute
 type acker interface {
 	AckID(id trackingMessageID)
 	NackID(id trackingMessageID)
+	NackMsg(msg Message)
 }
 
 type consumer struct {
@@ -85,6 +86,10 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		if options.Schema.GetSchemaInfo().Type == NONE {
 			options.Schema = NewBytesSchema(nil)
 		}
+	}
+
+	if options.NackBackoffPolicy == nil {
+		options.NackBackoffPolicy = new(defaultNackBackoffPolicy)
 	}
 
 	// did the user pass in a message channel?
@@ -326,6 +331,7 @@ func (c *consumer) internalTopicSubscribeToPartitions() error {
 				partitionIdx:               idx,
 				receiverQueueSize:          receiverQueueSize,
 				nackRedeliveryDelay:        nackRedeliveryDelay,
+				nackBackoffPolicy:          c.options.NackBackoffPolicy,
 				metadata:                   metadata,
 				replicateSubscriptionState: c.options.ReplicateSubscriptionState,
 				startMessageID:             trackingMessageID{},
@@ -489,7 +495,16 @@ func (c *consumer) ReconsumeLater(msg Message, delay time.Duration) {
 }
 
 func (c *consumer) Nack(msg Message) {
-	c.NackID(msg.ID())
+	mid, ok := c.messageID(msg.ID())
+	if !ok {
+		return
+	}
+
+	if mid.consumer != nil {
+		mid.Nack()
+		return
+	}
+	c.consumers[mid.partitionIdx].NackMsg(msg)
 }
 
 func (c *consumer) NackID(msgID MessageID) {
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -88,7 +88,7 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		}
 	}
 
-	if options.NackBackoffPolicy == nil && options.EnableDefaultNackBackoffPolicy == true {
+	if options.NackBackoffPolicy == nil && options.EnableDefaultNackBackoffPolicy {
 		options.NackBackoffPolicy = new(defaultNackBackoffPolicy)
 	}
 
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -88,7 +88,7 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		}
 	}
 
-	if options.NackBackoffPolicy == nil {
+	if options.NackBackoffPolicy == nil && options.EnableDefaultNackBackoffPolicy == true {
 		options.NackBackoffPolicy = new(defaultNackBackoffPolicy)
 	}
 
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -495,16 +495,21 @@ func (c *consumer) ReconsumeLater(msg Message, delay time.Duration) {
 }
 
 func (c *consumer) Nack(msg Message) {
-	mid, ok := c.messageID(msg.ID())
-	if !ok {
+	if c.options.EnableDefaultNackBackoffPolicy || c.options.NackBackoffPolicy != nil {
+		mid, ok := c.messageID(msg.ID())
+		if !ok {
+			return
+		}
+
+		if mid.consumer != nil {
+			mid.Nack()
+			return
+		}
+		c.consumers[mid.partitionIdx].NackMsg(msg)
 		return
 	}
 
-	if mid.consumer != nil {
-		mid.Nack()
-		return
-	}
-	c.consumers[mid.partitionIdx].NackMsg(msg)
+	c.NackID(msg.ID())
 }
 
 func (c *consumer) NackID(msgID MessageID) {
--- a/pulsar/negative_acks_tracker.go
+++ b/pulsar/negative_acks_tracker.go
@@ -66,7 +66,6 @@ func newNegativeAcksTracker(rc redeliveryConsumer, delay time.Duration,
 			negativeAcks:       make(map[messageID]time.Time),
 			rc:                 rc,
 			tick:               time.NewTicker(delay / 3),
-			tickForNackBackoff: nil,
 			nackBackoff:        nil,
 			delay:              delay,
 			log:                logger,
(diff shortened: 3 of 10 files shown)


wolfstudy:
```
==================
WARNING: DATA RACE
Read at 0x00c000158578 by goroutine 447:
  github.com/apache/pulsar-client-go/pulsar.(*negativeAcksTracker).track()
      /pulsar-client-go/pulsar/negative_acks_tracker.go:142 +0x7dc

Previous write at 0x00c000158578 by goroutine 291:
  github.com/apache/pulsar-client-go/pulsar.(*negativeAcksTracker).AddMessage()
      /pulsar-client-go/pulsar/negative_acks_tracker.go:103 +0x15e
  github.com/apache/pulsar-client-go/pulsar.TestNackBackoffTracker()
      /pulsar-client-go/pulsar/negative_acks_tracker_test.go:157 +0x124
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202

Goroutine 447 (running) created at:
  github.com/apache/pulsar-client-go/pulsar.(*negativeAcksTracker).AddMessage()
      /pulsar-client-go/pulsar/negative_acks_tracker.go:108 +0x4a4
  github.com/apache/pulsar-client-go/pulsar.TestNackBackoffTracker()
      /pulsar-client-go/pulsar/negative_acks_tracker_test.go:156 +0xeb
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202

Goroutine 291 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1168 +0x5bb
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1439 +0xa6
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1437 +0x612
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1345 +0x3b3
  main.main()
      _testmain.go:467 +0x356
==================
==================
WARNING: DATA RACE
Read at 0x00c000a1ff40 by goroutine 447:
  github.com/apache/pulsar-client-go/pulsar.(*negativeAcksTracker).track()
      /pulsar-client-go/pulsar/negative_acks_tracker.go:142 +0x7fb

Previous write at 0x00c000a1ff40 by goroutine 291:
  time.NewTicker()
      /usr/local/go/src/time/tick.go:30 +0x9a
  github.com/apache/pulsar-client-go/pulsar.(*negativeAcksTracker).AddMessage()
      /pulsar-client-go/pulsar/negative_acks_tracker.go:103 +0x134
  github.com/apache/pulsar-client-go/pulsar.TestNackBackoffTracker()
      /pulsar-client-go/pulsar/negative_acks_tracker_test.go:157 +0x124
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202

Goroutine 447 (running) created at:
  github.com/apache/pulsar-client-go/pulsar.(*negativeAcksTracker).AddMessage()
      /pulsar-client-go/pulsar/negative_acks_tracker.go:108 +0x4a4
  github.com/apache/pulsar-client-go/pulsar.TestNackBackoffTracker()
      /pulsar-client-go/pulsar/negative_acks_tracker_test.go:156 +0xeb
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202

Goroutine 291 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1168 +0x5bb
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1439 +0xa6
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202
  testing.runTests()
      /usr/local/go/src/testing/testing.go:1437 +0x612
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1345 +0x3b3
  main.main()
      _testmain.go:467 +0x356
==================
--- FAIL: TestNackBackoffTracker (0.40s)
```

pulsar/negative_backoff_policy.go
```diff
@@ -0,0 +1,46 @@
...
+	// The redeliveryCount indicates the number of times the message was redelivered.
+	// We can get the redeliveryCount from the CommandMessage.
+	Next(redeliveryCount uint32) int64
+}
+
+// defaultNackBackoffPolicy is default impl for NackBackoffPolicy.
+type defaultNackBackoffPolicy struct{}
+
+func (nbp *defaultNackBackoffPolicy) Next(redeliveryCount uint32) int64 {
+	minNackTimeMs := int64(1000 * 10) // 10sec
```
cckellogg, line 38: Let's use the time.Duration constants 
```
10 * time.Seconds
10 * time.Minutes
```

pulsar/negative_backoff_policy.go
```diff
@@ -0,0 +1,46 @@
...
+	maxNackTimeMs := 1000 * 60 * 10   // 10min
+
+	if redeliveryCount < 0 {
+		return minNackTimeMs
+	}
+
+	return int64(math.Min(math.Abs(float64(minNackTimeMs<<redeliveryCount)), float64(maxNackTimeMs)))
```
cckellogg, line 45: Can you add some comments to what this logic is doing. For me it's difficult to look and just understand it.

pulsar/negative_acks_tracker.go
```diff
@@ -105,15 +158,13 @@ func (t *negativeAcksTracker) track() {
...
 			}
-
 		}
 	}
 }
 
 func (t *negativeAcksTracker) Close() {
 	// allow Close() to be invoked multiple times by consumer_partition to avoid panic
 	t.doneOnce.Do(func() {
-		t.tick.Stop()
```
cckellogg, line 116: How is the ticker getting cleanup now?

pulsar/negative_acks_tracker.go
```diff
@@ -35,22 +35,41 @@ type negativeAcksTracker struct {
...
-		negativeAcks: make(map[messageID]time.Time),
-		rc:           rc,
-		tick:         time.NewTicker(delay / 3),
-		delay:        delay,
-		log:          logger,
-	}
+func newNegativeAcksTracker(rc redeliveryConsumer, delay time.Duration,
+	nackBackoffPolicy NackBackoffPolicy, logger log.Logger) *negativeAcksTracker {
+
+	t := new(negativeAcksTracker)
```
cckellogg, line 47: this can just be `var t *negativeAcksTracker`

pulsar/negative_acks_tracker.go
```diff
@@ -35,22 +35,41 @@ type negativeAcksTracker struct {
...
+
+	// When using NackBackoffPolicy, the delay time needs to be calculated based on the RedeliveryCount field in
+	// the CommandMessage, so for the original default Nack() logic, we still keep the negativeAcksTracker created
+	// when we open a gorutine to execute the logic of `t.track()`. But for the NackBackoffPolicy method, we need
+	// to execute the logic of `t.track()` when AddMessage().
+	if nackBackoffPolicy != nil {
```
cckellogg, line 53: I'm a little confused on why we need an if statement. Shouldn't the default Implementation of the `NackBackoffPolicy` be what the current behavior is? The benefit of the interface is to simply the code and delegate the to the implementation.

```
bp := nackBackoffPolicy
if bp == nil {
  bp = newDefaultBackoffPolicy(delay)
}
t = &negativeAcksTracker{
			doneCh:       make(chan interface{}),
			negativeAcks: make(map[messageID]time.Time),
			nackBackoff:  bp,
			rc:           rc,
			log:          logger,
		}
```

pulsar/negative_acks_tracker.go
```diff
@@ -76,14 +95,48 @@ func (t *negativeAcksTracker) Add(msgID messageID) {
 	t.negativeAcks[batchMsgID] = targetTime
 }
 
-func (t *negativeAcksTracker) track() {
+func (t *negativeAcksTracker) AddMessage(msg Message) {
```
cckellogg, line 98: Why is there a new method here?

Also, it looks like state is changing here without a lock. If multiple go routines call this at once multiple tracking routines could be started right? 

Can the tracking go routine just be started at creation time?

pulsar/consumer.go
```diff
@@ -158,6 +158,17 @@ type ConsumerOptions struct {
 
 	// Decryption decryption related fields to decrypt the encrypted message
 	Decryption *MessageDecryptionInfo
+
+	// If enabled, the default implementation of NackBackoffPolicy will be used to calculate the delay time of
+	// nack backoff, Default: false.
+	EnableDefaultNackBackoffPolicy bool
```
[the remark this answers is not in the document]
wolfstudy: If there is no `EnableDefaultNackBackoffPolicy`, it will invade the existing code logic. When the NackBackoffPolicy policy is empty, suppose we use the default NackBackoffPolicy, then when the user uses the Nack(Message) interface, the new implementation will be used.

pulsar/negative_backoff_policy.go
wolfstudy: Sure, will add comment for this change

pulsar/negative_acks_tracker.go
wolfstudy: They are the same effect

pulsar/negative_backoff_policy.go
wolfstudy: Because the << operation is required to cooperate with redeliveryCount, the unit conversion is still required, which will be converted to time.Duration in subsequent use.

pulsar/negative_backoff_policy.go
wolfstudy: Here we will first get the redeliveryCount object from the CommandMessage, and then start the << operation from minNackTimeMs to calculate the current length of time that nack needs to be executed, and then compare it with maxNackTimeMs, and take their maximum value as the nack duration. nack will increase from minNackTimeMs to maxNackTimeMs according to the above rule

wolfstudy: Here we will first get the redeliveryCount object from the CommandMessage, and then start the << operation from minNackTimeMs to calculate the current length of time that nack needs to be executed, and then compare it with maxNackTimeMs, and take their maximum value as the nack duration. nack will increase from minNackTimeMs to maxNackTimeMs according to the above rule

pulsar/negative_acks_tracker.go
wolfstudy: Yes, agree with your point of view. The problem here is because, for nackbackoff, we can't directly get the corresponding nackDelayTime, we need to get the redeliveryCount through the CommandMessage and then calculate the nackDelayTime, then we can determine the time.NewTicker based on the nackDelayTime. It is precisely because of such a relationship that the if statement is added

pulsar/negative_acks_tracker.go
wolfstudy: Because we need to get redeliveryCount through the Message interface

pulsar/negative_acks_tracker.go
wolfstudy: In the current implementation situation, if we use the t.ticker in the struct, there will be a data race, so now we use the temporary variables of the ticker, and there is no good way to see how to close the temporarily created ticker.

pulsar/consumer.go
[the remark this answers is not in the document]
cckellogg: To me a cleaner API is to just have `NackBackoffPolicy` and expose the basic/default policy. If the is policy is not set than it uses the current behavior. This way there is only 1 configuration knob to worry about.

```
// current behavior
ConsumerOpts{}

// custom behavior
ConsumerOpts{
  NackBackoffPolicy: pulsar.NewExpNackBackoffPolicy(),
}
```

pulsar/consumer_impl.go
```diff
@@ -489,6 +495,20 @@ func (c *consumer) ReconsumeLater(msg Message, delay time.Duration) {
 }
 
 func (c *consumer) Nack(msg Message) {
+	if c.options.EnableDefaultNackBackoffPolicy || c.options.NackBackoffPolicy != nil {
```
cckellogg, line 498: Nit create a single variable at creation for checking
c.useBackoffPolicy or something like that

pulsar/negative_backoff_policy.go
cckellogg: Nit let's make these constants 
```
minNackTimeMs := int64(1000 * 30) // 30sec
maxNackTimeMs := 1000 * 60 * 10   // 10min
```
and please add some comments about how this `int64(math.Min(math.Abs(float64(minNackTimeMs<<redeliveryCount)), float64(maxNackTimeMs)))` works

freeznet (approved):

zymap (approved):

PR #663, opened 2021-11-08
[Issue 662] Fix race in connection.go waitUntilReady()

### LINKED ISSUE #662: newPartitionProducer() occasionally hangs, waiting for a ready connection
#### Expected behavior

When creating a partitioned producer, producers for all partitions should be created.

#### Actual behavior

I am writing to a topic with six partitions. Very occasionally, I experience a process which is stuck because the producer for one of the partitions is not creating.

I recently had a process in this state, and I attached _delve_ to it in order to see exactly where it was stuck. Looking first at the main goroutine, then we can see that it is stuck in _internalCreatePartitionsProducers()_:

```
(dlv) bt
 0  0x00000000004425b6 in runtime.gopark
    at /usr/local/go/src/runtime/proc.go:367
 1  0x0000000000411bec in runtime.chanrecv
    at /usr/local/go/src/runtime/chan.go:576
 2  0x0000000000411658 in runtime.chanrecv2
    at /usr/local/go/src/runtime/chan.go:444
 3  0x000000000095e9ec in github.com/apache/pulsar-client-go/pulsar.(*producer).internalCreatePartitionsProducers
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/producer_impl.go:230
 4  0x000000000095dfaf in github.com/apache/pulsar-client-go/pulsar.newProducer
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/producer_impl.go:127
 5  0x0000000000945ba9 in github.com/apache/pulsar-client-go/pulsar.(*client).CreateProducer
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/client_impl.go:154
[...]
```

Line 230 of _producer_impl.go_ is waiting for a channel read:

```
pe, ok := <-c
```

The goroutine on the other end of this channel is launched at line 219 of _producer_impl.go_, and the issue lies in the call to _newPartitionProducer()_ on line 220. The full stack trace for that other goroutine is:

```
(dlv) bt
 0  0x00000000004425b6 in runtime.gopark
    at /usr/local/go/src/runtime/proc.go:367
 1  0x000000000046c3dd in runtime.goparkunlock
    at /usr/local/go/src/runtime/proc.go:372
 2  0x000000000046c3dd in sync.runtime_notifyListWait
    at /usr/local/go/src/runtime/sema.go:513
 3  0x0000000000477fac in sync.(*Cond).Wait
    at /usr/local/go/src/sync/cond.go:56
 4  0x00000000008e33cd in github.com/apache/pulsar-client-go/pulsar/internal.(*connection).waitUntilReady
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/internal/connection.go:339
 5  0x00000000008e8668 in github.com/apache/pulsar-client-go/pulsar/internal.(*connectionPool).GetConnection
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/internal/connection_pool.go:108
 6  0x00000000008f37ed in github.com/apache/pulsar-client-go/pulsar/internal.(*rpcClient).Request
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/internal/rpc_client.go:120
 7  0x0000000000960d63 in github.com/apache/pulsar-client-go/pulsar.(*partitionProducer).grabCnx
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/producer_partition.go:228
 8  0x00000000009602fa in github.com/apache/pulsar-client-go/pulsar.newPartitionProducer
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/producer_partition.go:161
 9  0x000000000095edc9 in github.com/apache/pulsar-client-go/pulsar.(*producer).internalCreatePartitionsProducers.func1
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/producer_impl.go:220
10  0x000000000095ed55 in github.com/apache/pulsar-client-go/pulsar.(*producer).internalCreatePartitionsProducers·dwrap·38
    at /go/pkg/mod/github.com/apache/pulsar-client-go@v0.7.0-candidate-1.0.20211104025721-d80a722ac1ab/pulsar/producer_impl.go:226
11  0x000000000046ff01 in runtime.goexit
    at /usr/local/go/src/runtime/asm_amd64.s:1581
```

For some reason, the call to _waitUntilReady()_ in _connection.go_ never completes.

The other five partitions seem to have created fine: I see creation log messages for all five, and in _delve_ I can see five running copies of _failTimeoutMessages_ and _runEventsLoop_ goroutines.

#### Steps to reproduce

This is not an easily bug to reproduce. It seems to occur around once per 10,000 invocations of the code in question, which means it happens around once per week for me.

It may be relevant that there are many new connections being made to the broker, at the time the bug occurs -- could a packet loss somewhere cause this, perhaps?

I do not know how to debug this further. However, if there is debug code I can insert, please let me know and I will do so.

#### System configuration

Pulsar version: 2.8.1
Client version: 0.7.0-candidate-1.0.20211104025721-d80a722ac1ab

The connection here is direct to the brokers, not via a proxy.

bschofield:
Fixes #662

### Motivation

As described in #662, there appears to be a potential race condition in connection.go function _waitUntilReady()_: the `cond.Broadcast()` can occur before the `cond.Wait()`.

[EDIT:] To be explicit, this is the issue:

[goroutine A] calls c.getState() and sees that it is not set to connectionReady
[goroutine B] changes the state to connectionReady
[goroutine B] sends a cond.Broadcast(), which goes nowhere because no goroutine is waiting.
[goroutine A] calls cond.Wait(), which never completes

### Modifications

Function _waitUntilReady()_ was previously holding the global `c.Lock()` on the connection. From my reading of the code, this mutex is intended to protect the `cnx` variable. I think that the use of `c.Lock()` in _waitUntilReady()_ was probably just a typo.

Instead, I think the author probably intended to grab the lock associated with the `sync.Cond`, i.e. `c.cond.L.Lock()`. This looks like the correct thing to do when using a `sync.Cond`. I think there should be a corresponding lock around the `cond.Broadcast()` also. See https://stackoverflow.com/questions/36857167/how-to-correctly-use-sync-cond/42772799#42772799 for more details.

### Verifying this change

I am unsure if this change is covered by existing tests. It fixes a rare race condition, so I think it would be quite difficult to test for.

I have deployed this change on my own production system, and it doesn't obviously break anything. I will report back if I see any issues that might be related to it.

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -327,8 +327,8 @@ func (c *connection) doHandshake() bool {
 }
 
 func (c *connection) waitUntilReady() error {
-	c.Lock()
-	defer c.Unlock()
+	c.cond.L.Lock()
+	defer c.cond.L.Unlock()
 
 	for c.getState() != connectionReady {
 		c.log.Debugf("Wait until connection is ready state=%s", c.getState().String())
@@ -894,6 +894,9 @@ func (c *connection) Close() {
 }
 
 func (c *connection) changeState(state connectionState) {
+	c.cond.L.Lock()
+	defer c.cond.L.Unlock()
+
 	c.setState(state)
 	c.cond.Broadcast()
 }
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -327,8 +327,9 @@ func (c *connection) doHandshake() bool {
 }
 
 func (c *connection) waitUntilReady() error {
-	// The lock is held here to prevent changeState() from modifying state and broadcasting
-	// in the time between we check the state and the time that we call cond.Wait().
+	// If we are going to call cond.Wait() at all, then we must call it _before_ we call cond.Broadcast().
+	// The lock is held here to prevent changeState() from calling cond.Broadcast() in the time between
+	// the state check and call to cond.Wait().
 	c.Lock()
 	defer c.Unlock()
 
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -327,8 +327,10 @@ func (c *connection) doHandshake() bool {
 }
 
 func (c *connection) waitUntilReady() error {
-	c.cond.L.Lock()
-	defer c.cond.L.Unlock()
+	// The lock is held here to prevent changeState() from modifying state and broadcasting
+	// in the time between we check the state and the time that we call cond.Wait().
+	c.Lock()
+	defer c.Unlock()
 
 	for c.getState() != connectionReady {
 		c.log.Debugf("Wait until connection is ready state=%s", c.getState().String())
@@ -894,8 +896,10 @@ func (c *connection) Close() {
 }
 
 func (c *connection) changeState(state connectionState) {
-	c.cond.L.Lock()
-	defer c.cond.L.Unlock()
+	// The lock is held here because we need setState() and cond.Broadcast() to be
+	// an atomic operation from the point of view of waitUntilReady().
+	c.Lock()
+	defer c.Unlock()
 
 	c.setState(state)
 	c.cond.Broadcast()


bschofield:
@cckellogg @BewareMyPower I think this relates to your discussion in https://github.com/apache/pulsar-client-go/pull/631/files#r723350028, would appreciate it if you could take a look.

BewareMyPower:
https://github.com/apache/pulsar-client-go/blob/fe3b7c4e445b3de42974ca692574229ad9099a45/pulsar/internal/connection.go#L212

And see https://pkg.go.dev/sync@go1.17.1#NewCond

> NewCond returns a new Cond with Locker l.

So I think `c.Lock()` is the same as `c.cond.L.Lock()`.

Back to your issue, how could you ensure `c.cond.Wait()` must be called before `c.cond.BroadCast()` is called? And I think it's not necessary. Please see my code snippet.

```go
package main

import (
	"fmt"
	"sync"
	"time"
)

func main() {
	m := sync.Mutex{}
	c := sync.NewCond(&m)
	flag := false

	defer m.Unlock()
	m.Lock()

	go func() {
		// No locks here
		fmt.Println("Before Broadcast")
		flag = true
		c.Broadcast()
		fmt.Println("After Broadcast")
	}()

	// Sleep for 2 seconds to make sure c.Broadcast() is called before c.Wait()
	time.Sleep(time.Duration(2) * time.Second)
	fmt.Println("Before Wait")
	for !flag {
		c.Wait()
	}
	fmt.Printf("After Wait, flag is %v\n", flag)
}
```

I called `time.Sleep` to make sure `c.Wait()` is called after `c.Broadcast()`. And the output is

```
Before Broadcast
After Broadcast
Before Wait
After Wait, flag is true
```

BewareMyPower:
Sorry I gave a wrong example. The `c.Wait()` is not called because `flag` is already true. It's right if my example was modified from

```go
	for !flag {
		c.Wait()
	}
```

to

```go
	for {
		c.Wait()
		if flag {
			break
		}
	}
```

BewareMyPower:
I agree that we should call `c.Broadcast()` after `c.Wait()`, but the code here is not like the example in https://stackoverflow.com/questions/36857167/how-to-correctly-use-sync-cond/42772799#42772799.

```go
    m := &sync.Mutex{}
    c := sync.NewCond(m)
    m.Lock()
    go func() {
        m.Lock() // Wait for c.Wait()
        c.Broadcast()
        m.Unlock()
    }()
    c.Wait() // Unlocks m, waits, then locks m again
    m.Unlock()
```

Because in that example, the `m.Lock()` has already been called before the goroutine is executed. So `c.Broadcast()` won't be called before `c.Wait()` releases the mutex internally. However, for the code here, how could you ensure that when `changeState` is called, the lock has already been acquired in `waitUntilReady`?

BewareMyPower:
The current `waitUntilReady` implementation is

```go
func (c *connection) waitUntilReady() error {
	c.Lock()
	defer c.Unlock()
	c.cond.L.Lock()
	defer c.cond.L.Unlock()

	for c.getState() != connectionReady {
		c.log.Debugf("Wait until connection is ready state=%s", c.getState().String())
                 // 1. `c.getState()` is called again
			return errors.New("connection error")
		}
                // 2. Even if the 2nd call of `c.getState()` returns connectionReady, `c.cond.Wait()` will still be called
		c.cond.Wait()
	}
	return nil
}
```

See my comments for my concern. `getState()` and `setState()` are both atomic. But if we the combination of two `getState()` calls is not atomic.

Therefore, I think a better solution is

```go
func (c *connection) waitUntilReady() error {
	for {
		switch state := c.getState(); state {
		case connectionReady:
			return nil
		case connectionClosed:
			return errors.New("connection error")
		default:
			c.log.Debugf("Wait until connection is ready state=%s", state.String())
			// Here we use c.cond.L.Lock() because we might not use c.Lock as the cond's internal Locker in future
			c.cond.L.Lock()
			defer c.cond.L.Unlock()
			c.cond.Wait()
		}
	}
}
```

pulsar/internal/connection.go
```diff
@@ -327,8 +327,8 @@ func (c *connection) doHandshake() bool {
 }
 
 func (c *connection) waitUntilReady() error {
-	c.Lock()
-	defer c.Unlock()
+	c.cond.L.Lock()
```
cckellogg, line 330: Curious what's the difference between doing `c.Lock()` vs `c.cond.L.Lock()`? Same for the changeState function? From looking at the code they should share the same lock?

bschofield:
Thanks for taking a look at this, @BewareMyPower and @cckellogg!

> So I think c.Lock() is the same as c.cond.L.Lock().

I'm pretty sure that's not the case. If we take a look at the definition of the connection type:

```go

type connection struct {
	sync.Mutex
	cond              *sync.Cond
[...]
```

...then we can see that `connection` struct is built out of a `sync.Mutex` using struct composition. That means that calls to `c.Lock()` are calling the `sync.Mutex` which is part of the `connection` struct itself.

If we next take a look at the definition of `sync.Cond` we see:

```go
type Cond struct {
	noCopy noCopy

	// L is held while observing or changing the condition
	L Locker
[...]
```

(where `Locker` is go's interface type for a `sync.Mutex`). So these are two separate locks: they are not the same.

Another way we can convince ourselves of this is by noting that it is forbidden to copy a `sync.Mutex`, you have to take pointers to it. Since the struct composition on `connection` does not use a pointer type, there would be no way to alias the mutex inside `cond` to the mutex on `connection`. They must be different things.

pulsar/internal/connection.go
bschofield: They're two different mutexes. See https://github.com/apache/pulsar-client-go/pull/663#issuecomment-963931326 for an explanation.

bschofield:
@BewareMyPower:
> However, for the code here, how could you ensure that when `changeState` is called, the lock has already been acquired in `waitUntilReady`?

That's the point of using the lock: It doesn't matter whether changeState() is called first, or waitUntilReady() is called first. Using the lock prevents the race condition that can occur in the gap between the test in `for c.getState() != connectionReady` and the call to `c.cond.Wait()` occurring.

If `changeState` gets called before `waitUntilReady`, then the broadcast happens first and all is well.

if `waitUntilReady` gets called before `changeState`, then `changeState` will block on the `cond.Wait()`, whilst still holding the lock. However, `cond.Wait()` call has a clever internal release of the lock which I think should still allow the `changeState` to happen:

```go
func (c *Cond) Wait() {
	c.checker.check()
	t := runtime_notifyListAdd(&c.notify)
	c.L.Unlock()
	runtime_notifyListWait(&c.notify, t)
	c.L.Lock()
}
```

(note that the `c.L` here is what we are calling `c.cond.L`).

bschofield:
@BewareMyPower: I tried this modification of your code, which I think is exactly the one you suggested:

```go
package main

import (
	"fmt"
	"sync"
	"time"
)

func main() {
	m := sync.Mutex{}
	c := sync.NewCond(&m)
	flag := false

	defer m.Unlock()
	m.Lock()

	go func() {
		// No locks here
		fmt.Println("Before Broadcast")
		flag = true
		c.Broadcast()
		fmt.Println("After Broadcast")
	}()

	// Sleep for 2 seconds to make sure c.Broadcast() is called before c.Wait()
	time.Sleep(time.Duration(2) * time.Second)
	fmt.Println("Before Wait")
	for {
		c.Wait()
		if flag {
			break
		}
	}
	fmt.Printf("After Wait, flag is %v\n", flag)
}
```

When I run that, this happens:
```
Before Broadcast
After Broadcast
Before Wait
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [sync.Cond.Wait]:
sync.runtime_notifyListWait(0xc0000b8050, 0x0)
	/home/ben/sdk/go1.17.1/src/runtime/sema.go:513 +0x13d
sync.(*Cond).Wait(0x4b4360)
	/home/ben/sdk/go1.17.1/src/sync/cond.go:56 +0x8c
main.main()
	/home/ben/tmp/foo.go:29 +0x191
```
So, I think that test case (which you very kindly provided) confirms that we agree that it really is necessary to call `cond.Broadcast()`  after `cond.Wait()`?

pulsar/internal/connection.go
bschofield: I agree that `waitUntilReady()` and `changeState()` should share the same lock. I think the correct lock to use is `c.cond.L`.

bschofield:
@BewareMyPower:
> Therefore, I think a better solution is [...]

I think that fix also looks fine (and maybe better), however you should avoid calling `defer` inside the for loop because a new defer call will be pushed onto the stack for each loop invocation. So it should be:

```go
func (c *connection) waitUntilReady() error {
	for {
		switch state := c.getState(); state {
		case connectionReady:
			return nil
		case connectionClosed:
			return errors.New("connection error")
		default:
			c.log.Debugf("Wait until connection is ready state=%s", state.String())
			// Here we use c.cond.L.Lock() because we are concerned about race conditions between the switch statement above, and the `cond.Wait()` below
			c.cond.L.Lock()
			c.cond.Wait()
			c.cond.L.Unlock()
		}
	}
}
```

I also edited your comment to make it clear why this lock is being used. I definitely agree that a comment is required, this is a really complicated situation and not made clearer by the slightly strange way that `sync.Cond` works.

bschofield:
@BewareMyPower:
> Therefore, I think a better solution is [...]

With that solution, the bug that I am experiencing in production could still occur:

[goroutine A] calls `c.getState()` and sees that it is not set to `connectionReady`
[goroutine B] changes the state to `connectionReady`
[goroutine B] sends a `cond.Broadcast()`, which goes nowhere because no goroutine is waiting.
[goroutine A] calls `cond.Wait()`, which never completes

To avoid this, it is necessary to hold the lock before calling `c.getState()`.

BewareMyPower:
> we agree that it really is necessary to call cond.Broadcast() after cond.Wait()?

Yeah, I agree.

> If changeState gets called before waitUntilReady, then the broadcast happens first and all is well.
>
> if waitUntilReady gets called before changeState, then waitUntilReady will block on the cond.Wait(), whilst still holding the lock. 

It makes sense. We should make the combination of `setState` and `c.Broadcast()` an atomic operation.

> however you should avoid calling defer inside the for loop because a new defer call will be pushed onto the stack for each loop invocation.

Thanks for pointing it out. I'm not so familiar with Go, just being used to C++ object's lifetime rule, which is a little different.

----

The only question I'm still confused is, could you explain following code and its output?

```go
package main

import (
	"fmt"
	"sync"
)

type connection struct {
	sync.Mutex
	cond *sync.Cond
}

func NewConnection() *connection {
	cnx := &connection{}
	cnx.cond = sync.NewCond(cnx)
	return cnx
}

func main() {
	cnx := NewConnection()
	fmt.Printf("%v\n%v\n", cnx, cnx.cond.L)

	cnx.cond.L.Lock()
	cnx.Lock()
	cnx.Unlock()
	cnx.cond.Wait()
}
```

```
&{{0 0} 0xc000104040}
&{{0 0} 0xc000104040}
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [semacquire]:
sync.runtime_SemacquireMutex(0xc000010244, 0x0, 0x1)
	/usr/local/go/src/runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc000010240)
	/usr/local/go/src/sync/mutex.go:138 +0x105
sync.(*Mutex).Lock(...)
	/usr/local/go/src/sync/mutex.go:81
main.main()
	main.go:24 +0x175
```

It looks like `cnx.Lock()` and `cnx.cond.L.Lock()` are the same.

BewareMyPower:
> @BewareMyPower:
> 
> > Therefore, I think a better solution is [...]
> 
> With that solution, the bug that I am experiencing in production could still occur:
> 
> [goroutine A] calls `c.getState()` and sees that it is not set to `connectionReady` [goroutine B] changes the state to `connectionReady` [goroutine B] sends a `cond.Broadcast()`, which goes nowhere because no goroutine is waiting. [goroutine A] calls `cond.Wait()`, which never completes
> 
> To avoid this, it is necessary to hold the lock before calling `c.getState()`.

It's right. I also realized this problem just now. Here we should lock the whole `changeState`. I'm just wondering whether there should be another lock for this. But `c.Lock()` might be used in more places in future. (If `c.Lock()` and `c.cond.L.Lock()` are the same)

bschofield:
> The only question I'm still confused is, could you explain following code and its output?

Ag! I'm so sorry @BewareMyPower, you're 100% correct: `c.Lock()` and `c.cond.L.Lock()` absolutely are the same lock. I had totally missed that `sync.NewCond` took a `Locker` as an argument. Thanks for clarifying that, and my apologies.

>  I'm just wondering whether there should be another lock for this.

I think that's up to you guys -- it depends what the ultimate intent of the lock is. Is it to protect against all state changes in the connection, or just some of them.

pulsar/internal/connection.go
bschofield: Nope, I was wrong! See https://github.com/apache/pulsar-client-go/pull/663#issuecomment-963966453 which proves that `c.Lock()` and `c.cond.L.Lock()` are the same thing.

BewareMyPower:
> I think that's up to you guys -- it depends what the ultimate intent of the lock is. Is it to protect against all state changes in the connection, or just some of them.

Yeah, it won't block this PR. Just say it because I've seen the similar case in C++ client that a common mutex is used everywhere.

For this PR, I think you can change `c.cond.L.` back to `c.`, the it will LGTM.

bschofield:
OK, I made that change and added a comment so hopefully nobody removes the lock in future. Thanks again for looking at this, I really appreciate it!

BewareMyPower (approved):

PR #669, opened 2021-11-19
[ClientOptions] Allow callers to specify custom metric name qualifiers

dferstay:
### Motivation

This change exposes the full metric name qualification supported in the
underlying prometheus client library by allowing metric names to be
qualified by both a "namespace" and "subsystem".  See the documentation for
the prometheus metric options structure:
https://pkg.go.dev/github.com/prometheus/client_golang@v1.11.0/prometheus#Opts

This is helpful to disambiguate metrics when there is more than one
component in a distributed system that makes use of pulsar-client-go.

### Modifications

A new, optional field has been added to the `ClientOptions` struct to allow for metrics namespaces

### Verifying this change

This change is trivial plumbing without any test coverage.

The public API: Yes, a new, optional field has been added to the `ClientOptions` struct

### Documentation

Does this pull request introduce a new feature? Yes

If yes, how is the feature documented? GoDocs

### DIFF
--- a/pulsar/internal/metrics.go
+++ b/pulsar/internal/metrics.go
@@ -88,7 +88,8 @@ type LeveledMetrics struct {
 	ReadersClosed       prometheus.Counter
 }
 
-func NewMetricsProvider(metricsCardinality int, userDefinedLabels map[string]string) *Metrics {
+func NewMetricsProvider(metricsCardinality int, userDefinedMetricsNamespace, userDefinedMetricsSubsystem string,
+	userDefinedLabels map[string]string) *Metrics {
 	constLabels := map[string]string{
 		"client": "go",
 	}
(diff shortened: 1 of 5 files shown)


pulsar/internal/metrics.go
```diff
@@ -114,183 +115,243 @@ func NewMetricsProvider(metricsCardinality int, userDefinedLabels map[string]str
 	metrics := &Metrics{
 		metricsLevel: metricsCardinality,
 		messagesPublished: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace:   userDefinedMetricsNamespace,
```
cckellogg, line 118: If the namespace and subsystem are empty strings will that have any effect on the metric name? 

Also, for my knowledge does setting the subsystem and namespace change the metric name? For example, this `pulsar_client_messages_published` => `namespace_subsystem_pulsar_client_messages_published`?

Curious, what are some advantages of this approach over others ones like custom labels or providing a custom prefix string to apply to the metric names?

dferstay: @cckellogg ,

> If the namespace and subsystem are empty strings will that have any effect on the metric name?

No it will not.  In pulsar/internal/metrics.go we currently create prometheus Opts structs with zero-value (`""`) Namespace and Subsystem fields:  https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/metrics.go#L116

> Also, for my knowledge does setting the subsystem and namespace change the metric name? For example, this `pulsar_client_messages_published` => `namespace_subsystem_pulsar_client_messages_published`?

Yes, setting the namespace and/or subsystem changes the fully-qualified name of the metric; your example is correct.

> Curious, what are some advantages of this approach over others ones like custom labels or providing a custom prefix string to apply to the metric names?

Providing a custom prefix string would have an equivalent effect.
Custom labels could be used for disambiguation as well, though consumption through PromQL would have to specify the appropriate label values.

> Are the namespace and subsystem going to be deprecated in later versions?

Oof, good catch!

dferstay:
@cckellogg , thanks for the review; I'm going to close this PR due to the deprecation of namespace and subsystem in the next version of the prometheus go client API.  If you think it is worthwhile exposing a custom metrics prefix string field in `ClientOptions` to achieve a similar effect let me know and I'll open a new PR.  Otherwise, users can lean on the existing custom metric labels to disambiguate metrics collected from different services using pulsar-client-go.

PR #670, opened 2021-11-23
Add new bug-resistant build constraints

reugn:
### Motivation
According to the [Bug-resistant build constraints proposal](https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md), //+build will be replaced by //go:build. A transition period from //+build to //go:build syntax will last from Go version 1.16 through version 1.18.

### Verifying this change

### DIFF
--- a/pulsar/internal/compression/zstd.go
+++ b/pulsar/internal/compression/zstd.go
@@ -15,6 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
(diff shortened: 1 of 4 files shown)


cckellogg (approved):
Maybe this can be a separate PR. Should we drop these classes `http_client_go_1.11.go` and `http_client_go_1.12.go`? The README now says go 1.13+ is required.

PR #677, opened 2021-12-01
Close the old connection to make sure the broker drops the producer on its side

### LINKED ISSUE #676: [BUG] When reconnecting to the server, the old connection is not closed
#### Expected behavior

When reconnecting to the server, the old connection should be closed.

#### Actual behavior

When reconnecting to the server, the old connection is not closed.

The following log progresses according to the timeline:

1. First, there is an active channel and producer that is writing data to the broker
image

2. In the case that the previous connection has not been disconnected, the client sends an add producer operation for the same producerName
image

3. This addition operation comes from the same ip of the client, a different port connection, and the producerName is equal to the previous one
image

4. Remove the producer in ServerCnx from the pulsar server, https://github.com/apache/pulsar/pull/11804, this PR modified the equals method of the Producer, resulting in the inability to remove the producer from the topic's prodcuers map.
image

5. Then the old connection starts to be disconnected, and the producer status of the server on the connection is cleared
image

image

6. The new channel connection keeps retrying addProducer, and the producer is added successfully after closing the old connection.

#### System configuration
**Pulsar version**: x.y

pulsar-client-go: 0.7.0
pulsar broker: 2.8.1

wenbingshen:
Master Issue: #676 

### Motivation
When reconnecting to the server, the old connection should be closed.

### Modifications

Close the old connection to make sure the broker drops the producer on its side

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -858,13 +858,14 @@ func (c *connection) Close() {
 		c.Lock()
 		cnx := c.cnx
 		c.Unlock()
-		c.changeState(connectionClosed)
+		c.changeState(connectionClosing)
 
 		if cnx != nil {
 			_ = cnx.Close()
 		}
 
 		close(c.closeCh)
+                c.changeState(connectionClosed)
 
 		listeners := make(map[uint64]ConnectionListener)
 		c.listenersLock.Lock()
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -865,7 +865,7 @@ func (c *connection) Close() {
 		}
 
 		close(c.closeCh)
-                c.changeState(connectionClosed)
+		c.changeState(connectionClosed)
 
 		listeners := make(map[uint64]ConnectionListener)
 		c.listenersLock.Lock()


wenbingshen:
@cckellogg PTAL, thanks

pulsar/internal/connection.go
```diff
@@ -858,13 +858,14 @@ func (c *connection) Close() {
 		c.Lock()
 		cnx := c.cnx
 		c.Unlock()
-		c.changeState(connectionClosed)
+		c.changeState(connectionClosing)
```
cckellogg, line 861: How does changing the state here to `connectionClosing` and then changing it to `connectionClosed` later fix the issue?

wenbingshen: This is not to say that the problem is solved, but first change the status to `connectionClosing` , the logic is more rigorous, I modified the description of the PR, it does not mean that the issue is fixed, but it is related to investigating this issue.

wenbingshen: If there is no intermediate `connectionClosing` state, other threads may start to create new connections in `GetConnection` before the current `cnx` is really closed. This is a mistake in itself. State transitions should be stricter. You can see that the network connection state transition of the java client is very rigorous.

If we can ensure that the state transition is more rigorous, set the state to `connectionClosing` before `cnx`  closed, which can prevent other threads to create a new connection. This may be able to solve the issue #676 that the old connection was not closed, and the new connection has initiated a request, even if this is not the root cause of the problem.

PR #678, opened 2021-12-02
[Issue 513] Correct apparent logic error in batchContainer's hasSpace() method

bschofield:
Fixes issue https://github.com/apache/pulsar-client-go/issues/513.

This issue was previously partially addressed by @zzzming in PR https://github.com/apache/pulsar-client-go/pull/528.

### Motivation

Following PR https://github.com/apache/pulsar-client-go/pull/528, the logic for the `hasSpace()` function, which is used to determine whether a batch container has enough space for a new message, reads:

```
func (bc *keyBasedBatchContainer) hasSpace(payload []byte) bool {
	return bc.numMessages+1 < bc.maxMessages || (bc.buffer.ReadableBytes()+msgSize) < uint32(bc.maxBatchSize)
}
```

Currently, then, the batch container will be considered to have space (be non-full) until **both** the max-messages and the max-size limits are hit.

I think that is incorrect. Nearby in the code, the batch container is considered to be full if **either** of the conditions are hit:

```
// IsFull check if the size in the current batch exceeds the maximum size allowed by the batch
func (bc *batchContainer) IsFull() bool {
	return bc.numMessages >= bc.maxMessages || bc.buffer.ReadableBytes() > uint32(bc.maxBatchSize)
}
```

Because the logical rule is `!(A || B) = (!A) && (!B)`, the `||` in `hasSpace()` should be an `&&`.

### Modifications

Change the `||` in `hasSpace()` to an `&&`, so that the container will be considered full if either limit is hit.

### Verifying this change

~This change is a trivial rework / code cleanup without any test coverage.~

Added `TestMaxBatchSize()`.

### Documentation

### DIFF
--- a/pulsar/internal/batch_builder.go
+++ b/pulsar/internal/batch_builder.go
@@ -155,7 +155,7 @@ func (bc *batchContainer) IsFull() bool {
 	return bc.numMessages >= bc.maxMessages || bc.buffer.ReadableBytes() > uint32(bc.maxBatchSize)
 }
 
-// hasSpace should return true if and only if the batch container can accommodate another message with the length of payload.
+// hasSpace should return true if and only if the batch container can accommodate another message of length payload.
 func (bc *batchContainer) hasSpace(payload []byte) bool {
 	msgSize := uint32(len(payload))
 	return bc.numMessages+1 < bc.maxMessages && (bc.buffer.ReadableBytes()+msgSize) < uint32(bc.maxBatchSize)
--- a/pulsar/internal/batch_builder.go
+++ b/pulsar/internal/batch_builder.go
@@ -150,15 +150,15 @@ func NewBatchBuilder(
 	return &bc, nil
 }
 
-// IsFull check if the size in the current batch exceeds the maximum size allowed by the batch
+// IsFull checks if the size in the current batch meets or exceeds the maximum size allowed by the batch
 func (bc *batchContainer) IsFull() bool {
-	return bc.numMessages >= bc.maxMessages || bc.buffer.ReadableBytes() > uint32(bc.maxBatchSize)
+	return bc.numMessages >= bc.maxMessages || bc.buffer.ReadableBytes() >= uint32(bc.maxBatchSize)
 }
 
 // hasSpace should return true if and only if the batch container can accommodate another message of length payload.
 func (bc *batchContainer) hasSpace(payload []byte) bool {
 	msgSize := uint32(len(payload))
-	return bc.numMessages+1 < bc.maxMessages && (bc.buffer.ReadableBytes()+msgSize) <= uint32(bc.maxBatchSize)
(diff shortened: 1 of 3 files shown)


bschofield:
I see this is now causing a max-message-size test to fail with an `AddToBatchFailed` error. Looking at the code, I think that's actually a legit failure mode when the message payload is too large, but would be great if someone like @wolfstudy had the time to take a look at some point.

bschofield:
The tests now seem to be failing on `TestNacksTracker()`. Is that a consequence of this change, or an unrelated issue?

cckellogg:
> The tests now seem to be failing on `TestNacksTracker()`. Is that a consequence of this change, or an unrelated issue?

It's probably unrelated since the tests passed for the over go versions. The concerning part is there may be a flaky `TestNacksTracker()` test now?

sharashka-io:
> The concerning part is there may be a flaky `TestNacksTracker()` test now?

I would guess that it's probably this part of `newNackMockedConsumer()`:

```
		// since the client ticks at an interval of delay / 3
		// wait another interval to ensure we get all messages
		time.Sleep(testNackDelay + 101*time.Millisecond)
```

Since the sleep above happens inside a new goroutine, and (*negativeAcksTracker).track() also runs inside a new goroutine, it's probably possible in rare cases for the goroutine scheduling to interfere with the intended sequence of actions.

bschofield:
The concerning part is there may be a flaky TestNacksTracker() test now?

I would guess that it's probably this part of newNackMockedConsumer():

```
		// since the client ticks at an interval of delay / 3
		// wait another interval to ensure we get all messages
		time.Sleep(testNackDelay + 101*time.Millisecond)
```

Since the sleep above happens inside a new goroutine, and (*negativeAcksTracker).track() also runs inside a new goroutine, it's probably possible in rare cases for the goroutine scheduling to interfere with the intended sequence of actions.

bschofield:
I added `TestMaxBatchSize()` to `producer_test.go`, based on `TestMaxBatchSize()`.

From the results of that test, it looks like the `<` in the size test should actually be `<=`, to accommodate the case where the new payload fits exactly in the container.

cckellogg (approved):

zzzming:
We should use this 
```
bc.numMessages+1 <= bc.maxMessages 
```

Because this will accommodate an individual message when batch is disabled. Currently this configuration fails at [unable to add message to batch]
```
pulsar.ProducerOptions{
		Topic:               topicName,
		DisableBatching:     true,
		BatchingMaxMessages: 1,
		BatchingMaxSize:     1,
	})
```

bschofield:
Made the additional tweak as suggested by @zzzming, also cleaned up `IsFull()` (which is not actually used anywhere as far as I can see).

wolfstudy (approved):
LGTM +1

PR #691, opened 2021-12-24
Use a separate gorutine to handle the logic of reconnect

### LINKED ISSUE #687: [Discuss] Split eventsCh to use a separate Channel to handle `runEventsLoop`
Currently, when sending and consuming messages, we use eventsCh to receive command requests for message sending and receiving:

```
func (p *partitionProducer) runEventsLoop() {
	for {
		select {
		case i := <-p.eventsChan:
			switch v := i.(type) {
			case *sendRequest:
				p.internalSend(v)
			case *flushRequest:
				p.internalFlush(v)
			case *closeProducer:
				p.internalClose(v)
				return
			}
		case <-p.connectClosedCh:
			p.reconnectToBroker()
		case <-p.batchFlushTicker.C:
			if p.batchBuilder.IsMultiBatches() {
				p.internalFlushCurrentBatches()
			} else {
				p.internalFlushCurrentBatch()
			}
		}
	}
}
```

This looks OK under normal circumstances, but in extreme cases they may affect each other. For example, for send command, we have a parameter of maxPendingMessages locally, which is used as the size of the eventsCh cache(default: 1000):

```
eventsChan:       make(chan interface{}, maxPendingMessages),
```

Assuming that the local maxPendingMessages reaches the default threshold at this time, then the channel will enter a blocking state (in fact, this is completely possible), at this time related flush or close requests will also be blocked by this eventsCh, because they use the same channel(`eventsCh`).

Once runEventsLoop is blocked, it will cause the following series of problems:

1. send message TimoutError

image

When a certain block action is triggered, this phenomenon will continue to exist until the Go SDK is restarted.

2. the `receiveCommand()` stuck and the gorutine is `gopark`
image

[pprof.gateway.goroutine.008.pb.gz](https://github.com/apache/pulsar-client-go/files/7767543/pprof.gateway.goroutine.008.pb.gz)

In this case, when we try to `bin/pulsar-admin topics unload` this topic, it doesn't work, because in fact it is the block inside the Go SDK that caused the **sending timeout**. At this point, assuming that the Go SDK is restarted, this service can be restored immediately.

So, Here I want to discuss whether we can split eventsCh, and each request corresponds to an independent Go Channel. This has two advantages:

1. There will be no mutual influence between the multiple commands discussed above.
2. The threshold limit for maxPendingMessages is more precise. Because currently multiple requests may be sent to eventsCh, it receives more than just send command.

wolfstudy:

Fixes #687

### Motivation

As #687 said, use a separate gorutine to handle the logic of reconnect. Avoid blocking the reconnection logic due to the gorutine block of runEventsLoop

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -77,8 +77,10 @@ type partitionProducer struct {
 	batchFlushTicker         *time.Ticker
 
 	// Channel where app is posting messages to be published
-	eventsChan      chan interface{}
 	connectClosedCh chan connectionClosed
+	sendRequestCh   chan sendRequest
+	flushRequestCh  chan flushRequest
+	closeProducerCh chan closeProducer
 
 	publishSemaphore internal.Semaphore
 	pendingQueue     internal.BlockingQueue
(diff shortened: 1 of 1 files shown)


zymap (approved):

pulsar/producer_partition.go
```diff
@@ -369,6 +369,16 @@ func (p *partitionProducer) reconnectToBroker() {
 }
 
 func (p *partitionProducer) runEventsLoop() {
+
+	go func() {
+		for {
+			for range p.connectClosedCh {
```
cckellogg, line 375: I think this is going to leak a go routine since the `p.connectClosedCh` is never closed?

PR #692, opened 2021-12-24
Set default go version to 1.13 in CI related files

reugn:
### Motivation
Set the default go version to 1.13 in the CI related files as per the minimum version requirements.


wolfstudy (approved):
LGTM +1

PR #699, opened 2022-01-05
client option to set DefaultMaxMessageSize

zzzming:
### Motivation

When the broker sets maxMessageSize more than 5MB, the go client can still get error `message size exceeds MaxMessageSize`. Upon some conditions, the broker is not communicate the max message size in CommandConnected.MaxMessageSize. This PR will allow the client to set a non-default (5MB) message size if the MaxMessageSize is not present from the control channel.

### Verifying this change

  - The public API: (yes)
  - The schema:  no 
  - The wire protocol: (no)

It exposes a new configuration parameter to allow the client set the mexMessageSize
### Documentation

### DIFF
--- a/pulsar/internal/connection_pool.go
+++ b/pulsar/internal/connection_pool.go
@@ -46,6 +46,7 @@ type connectionPool struct {
 	maxConnectionsPerHost int32
 	roundRobinCnt         int32
 	metrics               *Metrics
+	defaultMaxMessageSize int
 
 	log log.Logger
 }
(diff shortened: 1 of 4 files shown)


pulsar/internal/connection_pool.go
```diff
@@ -46,6 +46,7 @@ type connectionPool struct {
 	maxConnectionsPerHost int32
 	roundRobinCnt         int32
 	metrics               *Metrics
+	defaultMaxMessageSize int
```
dferstay, line 49: Suggest making this an `int32` to be consistent with the type of `connection.defaultMaxMessageSize`.

wolfstudy:
Hello @zzzming , in `connection` struct, we have defined the maxMessageSize field and obtained the current maxMessageSize value from the broker during the handshake. Does this function not take effect?

PR #700, opened 2022-01-06
Revert "Use a separate gorutine to handle the logic of reconnect"

wolfstudy:
Reverts apache/pulsar-client-go#691

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -77,8 +77,8 @@ type partitionProducer struct {
 	batchFlushTicker         *time.Ticker
 
 	// Channel where app is posting messages to be published
-	connectClosedCh chan connectionClosed
 	eventsChan      chan interface{}
+	connectClosedCh chan connectionClosed
 
 	publishSemaphore internal.Semaphore
 	pendingQueue     internal.BlockingQueue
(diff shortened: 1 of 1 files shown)


cckellogg (approved):

bschofield:
Just to be clear, this commit doesn't actually revert #691 but instead adds a close channel, right? If so it might be good to update the title.

I do have a separate concern with this approach, which might or might not be justified. It looks to me like operations which were previously serialized by virtue of the single goroutine in `runEventsLoop()` can now occur concurrently, because they are now running in separate goroutines.

For example: suppose the user calls `p.Flush()` on the producer, which causes a `&flushRequest{}` to be enqueued to `p.eventsChan`. The primary goroutine in `runEventsLoop()` will then call `p.internalFlush()` which will try to write to `p.cnx`. Suppose that at the same time, the broker closes the connection. and the new goroutine calls `p.reconnectToBroker()`, which tries to change `p.cnx` whilst `p.internalFlush()` is still using it.

Does it not matter that `p.reconnectToBroker()` can now be happening at the same time as `p.internalSend()` / `p.internalFlush()`, `p.internalClose()`, when previously the use of a single goroutine meant that these functions could not be executing simultaneously? Is there some other lock or feature of the code that means this is OK?

PR #701, opened 2022-01-11
Fix data race while accessing connection in partitionProducer

wolfstudy:
Signed-off-by: xiaolongran <rxl@apache.org>

### Motivation

In https://github.com/apache/pulsar-client-go/pull/700, we use a separate go rutine to handle the logic of reconnect, so here you may encounter the same data race problem as https://github.com/apache/pulsar-client-go/pull/535

### Modifications

Now, the conn field is read and written atomically; avoiding race conditions.

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -66,7 +66,6 @@ type partitionProducer struct {
 	client *client
 	topic  string
 	log    log.Logger
-	//cnx    internal.Connection
 
 	conn atomic.Value
 
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -963,4 +963,3 @@ func (p *partitionProducer) _getConn() internal.Connection {
 	//            invariant is broken
 	return p.conn.Load().(internal.Connection)
 }
-
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -66,7 +66,9 @@ type partitionProducer struct {
 	client *client
 	topic  string
 	log    log.Logger
-	cnx    internal.Connection
+	//cnx    internal.Connection
+
+	conn atomic.Value
 
 	options                  *ProducerOptions
 	producerName             string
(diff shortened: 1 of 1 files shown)


wolfstudy:
cc @dferstay PTAL thanks

pulsar/producer_partition.go
```diff
@@ -800,7 +801,7 @@ func (p *partitionProducer) ReceivedSendReceipt(response *pb.CommandSendReceipt)
 		// the state discrepancy.
 		p.log.Warnf("Received ack for %v on sequenceId %v - expected: %v, closing connection", response.GetMessageId(),
 			response.GetSequenceId(), pi.sequenceID)
-		p.cnx.Close()
+		p._getConn().Close()
```
freeznet, line 804: maybe clean up the cached connection when the cnx is closed? otherwise `_getConn()` will return a closed connection.

wolfstudy: Looks like https://github.com/apache/pulsar-client-go/pull/703 is working on this

leizhiyuan (approved):
LGTM

pulsar/producer_partition.go
wolfstudy: @freeznet Looking at the implementation of the current Close method, when triggering the logic of Close, the state of the current connection will first be set to connectionClosed. When GetConnection() is executed, if the state of the connection is detected to be closed, the connection will be removed from the cache.

```
func (c *connection) Close() {
	c.closeOnce.Do(func() {
		c.Lock()
		cnx := c.cnx
		c.Unlock()
		c.changeState(connectionClosed)
```


```
if conn.closed() {
			delete(p.connections, key)
			p.log.Infof("Removed connection from pool key=%s logical_addr=%+v physical_addr=%+v",
				key, conn.logicalAddr, conn.physicalAddr)
			conn = nil // set to nil so we create a new one
		}
```

congbobo184 (approved):

freeznet (approved):

PR #705, opened 2022-01-12
reader subscribe all partitioned topics

GPrabhudas:
Initial changes to address reader topic not found issue: https://github.com/apache/pulsar-client-go/issues/553

Changes: 
- create multi-topic reader

### DIFF
--- a/pulsar/client_impl.go
+++ b/pulsar/client_impl.go
@@ -179,14 +179,15 @@ func (c *client) CreateReader(options ReaderOptions) (Reader, error) {
 		}
 		c.handlers.Add(reader)
 		return reader, nil
-	} else {
-		multiTopicReader, err := newMultiTopicReader(c, options)
-		if err != nil {
-			return nil, err
-		}
-		c.handlers.Add(multiTopicReader)
-		return multiTopicReader, nil
 	}
+
+	// create multi topic reader
+	multiTopicReader, err := newMultiTopicReader(c, options)
+	if err != nil {
+		return nil, err
+	}
+	c.handlers.Add(multiTopicReader)
+	return multiTopicReader, nil
 }
 
 func (c *client) TopicPartitions(topic string) ([]string, error) {
--- a/pulsar/client_impl.go
+++ b/pulsar/client_impl.go
@@ -168,12 +168,25 @@ func (c *client) Subscribe(options ConsumerOptions) (Consumer, error) {
 }
 
 func (c *client) CreateReader(options ReaderOptions) (Reader, error) {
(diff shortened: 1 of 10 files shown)


wolfstudy:
cc @cckellogg PTAL

pulsar/consumer.go
```diff
@@ -248,4 +257,13 @@ type Consumer interface {
 
 	// Name returns the name of consumer.
 	Name() string
+
+	// lastDequeuedMsg used for setting last dequeued msg id by internal partition consumers
```
cckellogg, line 261: I don't think we should be adding private methods to the public interface. I think these should be removed. Also, it seems these are only relevant to the reader implementation so I think these helper functions should live within the reader files.

pulsar/consumer.go
```diff
@@ -182,6 +182,15 @@ type ConsumerOptions struct {
 	// > Notice: the NackBackoffPolicy will not work with `consumer.NackID(MessageID)`
 	// > because we are not able to get the redeliveryCount from the message ID.
 	NackBackoffPolicy NackBackoffPolicy
+
+	// startMessageID internally used by multitopic-reader
+	startMessageID MessageID
```
cckellogg, line 187: The ConsumerOptions is a public interface so I don't think we should add package private fields here. Let's remove these.

pulsar/consumer_partition.go
```diff
@@ -1265,6 +1265,10 @@ func (pc *partitionConsumer) _getConn() internal.Connection {
 	return pc.conn.Load().(internal.Connection)
 }
 
+func (pc *partitionConsumer) messagesInQueue() int {
+	return len(pc.queueCh)
```
cckellogg, line 1269: If there are multi go routines consuming and producing to this channel this value might not be useful after being read.

pulsar/client_impl.go
```diff
@@ -168,12 +168,26 @@ func (c *client) Subscribe(options ConsumerOptions) (Consumer, error) {
...
 func (c *client) CreateReader(options ReaderOptions) (Reader, error) {
-	reader, err := newReader(c, options)
+	topics, err := c.TopicPartitions(options.Topic)
 	if err != nil {
 		return nil, err
 	}
-	c.handlers.Add(reader)
-	return reader, nil
+	if len(topics) <= 1 {
+		reader, err := newReader(c, options)
```
cckellogg, line 176: What's the different between a reader and a multi reader? It seems like they should be the same but just consume a different number of partitions?

cckellogg (changes requested):

PR #706, opened 2022-01-13
fix batchIdx error for non-batch message

wolfstudy:

### Motivation

When producer send message or consumer receive message in non-batch, return the batchIdx is `0`

```
msgId: pulsar.trackingMessageID{messageID:pulsar.messageID{ledgerID:11653378, entryID:3807, batchIdx:0, partitionIdx:2}
```

In fact, when we disable batch in ProducerOptions, the batchIdx is `-1`

```
msgId: pulsar.trackingMessageID{messageID:pulsar.messageID{ledgerID:11653378, entryID:3807, batchIdx:0, partitionIdx:2}
```

### Modifications

- Get batchIdx from broker

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -596,7 +596,7 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 		msgID := newTrackingMessageID(
 			int64(pbMsgID.GetLedgerId()),
 			int64(pbMsgID.GetEntryId()),
-			int32(i),
+			pbMsgID.GetBatchIndex(),
 			pc.partitionIdx,
 			ackTracker)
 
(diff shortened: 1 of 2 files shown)


pulsar/consumer_partition.go
```diff
@@ -596,7 +596,7 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 		msgID := newTrackingMessageID(
 			int64(pbMsgID.GetLedgerId()),
 			int64(pbMsgID.GetEntryId()),
-			int32(i),
+			pbMsgID.GetBatchIndex(),
```
cckellogg, line 599: Could there be compatibility issues if users upgrade? Is this what the java client does too?

wolfstudy: Yes, this may break compatibility. The Java SDK side obtains the Batch Idx information from the Broker instead of maintaining the index information locally.

Therefore, if the user uses the Java SDK on the producer side and the Go SDK on the consumer side, there is no way to match the attributes of the message ID.

This effect is relatively small and only affects non-partitioned topic types. In most cases, users may prefer to use partitioned-topic

Issue #707, opened 2022-01-13
Race condition detected for producer/consumer when encryption is enabled


GPrabhudas:
#### Expected behaviour

Creation of producer/consumer on partitioned topic should not run into race condition.

#### Actual behaviour
For partitioned topic, creation of producer and consumer run into race condition when trying to set DefaultMessageCrypto if it is not provided in the options.

#### Steps to reproduce
add a test case https://github.com/GPrabhudas/pulsar-client-go/commit/039bbd0e6fec7452ecb90fd28a4165781631d68c

Test case fails with below warnings
```
....pulsar-client-go/pulsar/testing.go:1093: race detected during execution of test
--- FAIL: TestPartitionTopicsConsumerPubSubEncryption (1.06s)
=== CONT
    ....pulsar-client-go/pulsar/testing.go:1093: race detected during execution of test
FAIL
FAIL    github.com/apache/pulsar-client-go/pulsar       1.699s

WARNING: DATA RACE
Read at 0x00c0004ec130 by goroutine 294:
  github.com/apache/pulsar-client-go/pulsar.newPartitionConsumer()
      ....pulsar-client-go/pulsar/consumer_partition.go:189 +0x1490
  github.com/apache/pulsar-client-go/pulsar.(*consumer).internalTopicSubscribeToPartitions.func1()
      ....pulsar-client-go/pulsar/consumer_impl.go:348 +0x899

Previous write at 0x00c0004ec130 by goroutine 50:
  github.com/apache/pulsar-client-go/pulsar.newPartitionConsumer()
      ....pulsar-client-go/pulsar/consumer_partition.go:194 +0x172d
  github.com/apache/pulsar-client-go/pulsar.(*consumer).internalTopicSubscribeToPartitions.func1()
      ....pulsar-client-go/pulsar/consumer_impl.go:348 +0x899

Goroutine 294 (running) created at:
  github.com/apache/pulsar-client-go/pulsar.(*consumer).internalTopicSubscribeToPartitions()
      ....pulsar-client-go/pulsar/consumer_impl.go:317 +0x6b4
  github.com/apache/pulsar-client-go/pulsar.newInternalConsumer()
      ....pulsar-client-go/pulsar/consumer_impl.go:208 +0x548
  github.com/apache/pulsar-client-go/pulsar.newConsumer()
      ....pulsar-client-go/pulsar/consumer_impl.go:159 +0x764
[...]

PR #708, opened 2022-01-13
Encryption failure test case fix

GPrabhudas:
address issue: https://github.com/apache/pulsar-client-go/issues/707

### DIFF
--- a/pulsar/producer_impl.go
+++ b/pulsar/producer_impl.go
@@ -19,11 +19,13 @@ package pulsar
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
 	"unsafe"
 
+	"github.com/apache/pulsar-client-go/pulsar/crypto"
 	"github.com/apache/pulsar-client-go/pulsar/internal"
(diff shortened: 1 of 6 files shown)


pulsar/producer_impl.go
```diff
@@ -124,6 +126,25 @@ func newProducer(client *client, options *ProducerOptions) (*producer, error) {
 		}
 	}
 
+	encryption := options.Encryption
+	// add default message crypto if not provided
+	if encryption != nil && len(encryption.Keys) > 0 {
+		if encryption.KeyReader == nil {
```
wolfstudy, line 132: It seems that there is no similar check in `newConsumer` and `newReader`, do they not need it?

wolfstudy (commented):
@GPrabhudas Nice work, just a little comment, please check.

pulsar/producer_impl.go
GPrabhudas: Similar check is not needed on consumer/reader  side.

wolfstudy (approved):
LGTM +1

PR #720, opened 2022-01-28
[optimize] Stop batchFlushTicker when Disable batching

Shoothzj:
### Motivation

Disable batchFlushTicker when Disable batching, reduce cpu cost

### Modifications

Stop the batchFlushTicker when Disable batching

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -126,6 +126,9 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		metrics:          metrics,
 		epoch:            0,
 	}
+	if p.options.DisableBatching {
+		p.batchFlushTicker.Stop()
+	}
 	p.setProducerState(producerInit)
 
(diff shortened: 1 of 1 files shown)


Shoothzj:
This change is limit samllest change to the code.
Otherwise we can add a `if/else` in `func (p *partitionProducer) runEventsLoop()`

@wolfstudy

freeznet (approved):

pulsar/producer_partition.go
```diff
@@ -126,6 +126,9 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		metrics:          metrics,
 		epoch:            0,
 	}
+	if p.options.DisableBatching {
+		p.batchFlushTicker.Stop()
```
cckellogg, line 130: Is it ok to call Stop() multiple times? There is a call to stop in the internal close.

Shoothzj: It's ok

wolfstudy (approved):

PR #723, opened 2022-02-14
Dlq producer on topic with schema

GPrabhudas:
- create dlq producer with schema from consumer 
- address issue #726 
- follow java client to create dlq producer [here](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1984)


wolfstudy (approved):
LGTM +1

Issue #729, opened 2022-02-17
there was a ticker leak when create producer failed


NaraLuwan:

NaraLuwan:
The CPU is full when creating a large number of producers fails.

pprof heap 
![image](https://user-images.githubusercontent.com/15198796/154391384-c2119bed-b1dd-45b0-94eb-093d12e751d0.png)

pprof profile
![image](https://user-images.githubusercontent.com/15198796/154391561-73701e4d-8870-4377-b7a3-ab8f94007d90.png)

Shoothzj:
closed by #730

PR #730, opened 2022-02-17
[Issue 729]stop ticker when create producer failed

NaraLuwan:
Master Issue: [#729 ]

### Motivation

I received an alarm indicating that the server load was too high, used the pprof tool to locate the problem and determined that the cause was a ticker leak while a large number of producer creating failed.
Here are some screenshots to show it:
- pprof heap, there are a number of SDK-started ticker, accounted for 90%
image
- pprof cpu, cpu consumption is also here
image

### Modifications

stop ticker when create producer failed

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -145,6 +145,7 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 	}
 	err := p.grabCnx()
 	if err != nil {
+		p.batchFlushTicker.Stop()
 		logger.WithError(err).Error("Failed to create producer at newPartitionProducer")
 		return nil, err
 	}


Shoothzj (approved):

pulsar/producer_partition.go
```diff
@@ -145,6 +145,7 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 	}
 	err := p.grabCnx()
 	if err != nil {
+		p.batchFlushTicker.Stop()
```
wolfstudy, line 148: The `grabCnx()` is called in two places in total, except here in the`reconnectToBroker()` logic, it looks like we need the same fix.

wolfstudy (commented):
Thanks @NaraLuwan work for this, just a little comments, please check

pulsar/producer_partition.go
NaraLuwan: but `newTicker()` is called only with `newPartitionProducer()`, i don't think we need to fix it.

PR #732, opened 2022-02-21
[Issue 731] Prevent sent pendingItem from being resend again after the reconnection of producer

### LINKED ISSUE #731: Resend after reconnection cause reproduce of the same msg
https://github.com/apache/pulsar-client-go/blob/965045aa0d2c077826f64d0b88a5d5e55d9d50c4/pulsar/producer_partition.go#L272

After the reconnection of the producer, the msg in the pendingQueue will be resend.
But not every msg in pendingQueue hasn't send to broker, the resend logic here seems to deal with the problem that producer meet a network error when producing a msg which cause the msg doesn't really send to broker.

While, the code here
https://github.com/apache/pulsar-client-go/blob/master/pulsar/producer_partition.go#L793
will also trigger the reconnection logic again and again, as well as the resend logic, which cause the consumer to consume the same msg with difference msgId

WinChua:
Fixes #731

### Motivation

As the #731 has show, some msg maybe sent after the reconnection of the producer, no mater the msg has been sent or not, which will cause that the same msg will be sent and consumed mulit time at the application view.

### Modifications

A sent flag was added to [pendingItem](https://github.com/WinChua/pulsar-client-go/blob/42af6818c03cf6803289cdc5706eea09d82ec95f/pulsar/producer_partition.go#L520) to indicated that the corresponding buffer has been sent successfully.
And in the resend logic the sent item will not be sent again.

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -71,7 +71,7 @@ type ConnectionListener interface {
 type Connection interface {
 	SendRequest(requestID uint64, req *pb.BaseCommand, callback func(*pb.BaseCommand, error))
 	SendRequestNoWait(req *pb.BaseCommand) error
-	WriteData(data Buffer)
+	WriteData(data Buffer, onSuccess func())
 	RegisterListener(id uint64, listener ConnectionListener)
 	UnregisterListener(id uint64)
 	AddConsumeHandler(id uint64, handler ConsumerHandler)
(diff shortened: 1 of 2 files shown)


wolfstudy (changes requested):
Thanks @WinChua work for this

It seems that this is not a good way to deal with it. If the message is sent successfully, we will remove the current message from the pendingQueue. The messages in the pendingQueue are all the messages that the Broker did not return the sendReceipt command. In theory, the messages here are not sent successfully.

PR #719, opened 2022-01-26
[issue 679][oauth2] Fix macos compiler warnings

pgier:
This upgrades the dependency from the oauth2 module to [99designs/keyring](https://github.com/99designs/keyring/) to remove the compiler warnings on macos.

Fixes #679 
Fixes #626
Fixes #594

See also https://github.com/99designs/keyring/pull/102

### Motivation

*Explain here the context, and why you're making that change. What is the problem you're trying to solve.*

Building on recent versions of OSX/MacOS outputs the following compiler warnings.

```
cgo-gcc-prolog:81:11: warning: 'SecKeychainCreate' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecKeychain.h:301:10: note: 'SecKeychainCreate' has been explicitly marked deprecated here
cgo-gcc-prolog:139:11: warning: 'SecKeychainLock' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecKeychain.h:357:10: note: 'SecKeychainLock' has been explicitly marked deprecated here
cgo-gcc-prolog:159:11: warning: 'SecKeychainOpen' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecKeychain.h:287:10: note: 'SecKeychainOpen' has been explicitly marked deprecated here
cgo-gcc-prolog:183:11: warning: 'SecKeychainUnlock' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecKeychain.h:347:10: note: 'SecKeychainUnlock' has been explicitly marked deprecated here
cgo-gcc-prolog:203:11: warning: 'SecTrustedApplicationCreateFromPath' is deprecated: first deprecated in macOS 10.15 - No longer supported [-Wdeprecated-declarations]
[...]

PR #737, opened 2022-02-25
[issue 679][oauth2] Fix macos compiler warnings, remove store impleme…

llqgit:
…nt keyring which not used at all

Master Issue: #719 

### Motivation

Fix MacOS version warning from the unused OAuth2 store implement named keyring.go

### Modifications

To fix the warning that on MacOS once for all by remove the `keyring.go` which no references from other places, rather than upgrade the keyring package.

### Verifying this change

    -  removed `github.com/99designs/go-keychain`

### Documentation


llqgit:
Any change to confirm PR ?

Shoothzj:
@wolfstudy can you confirm this PR? thanks

wolfstudy:
cc @zymap Please take a look thanks

zymap (changes requested):
It's a persistent implementation, we shouldn't remove it. It's an option of the store implementation, user can choose by themselves.

llqgit:
@zymap I got it and understand your point. 
And I still have a question about it.

If it should be kept for the users which choose this implementation, now what's your opinion about the users not use this extra implementation and still shown the warning below?

It is really annoying when you develop and run the program in every time.

image

zymap:
I saw the author has fixed this issue(https://github.com/keybase/go-keychain/pull/76). Could you please try to upgrade the dependency to check if it works?

llqgit:
@zymap I just run:
```
go mod init my_project
go mod tidy
```
This is the official way to init, it should be all right.
And the dependences like this:

image

image

If I run my program, the errors shown like this:
image

You said the new version of `github.com/keybase/go-keychain` fix the problem. 
So, if I just upgrade this package by `go get github.com/keybase/go-keychain` CMD.

The version would be:
image

It should be good. But I got some new errors which looks worse:

image

By the way. My mac OS version

image


From my perspective, the warning  problem was not fixed yet or not effective.

ciiiii:
@llqgit I think they remove the support to avoid the warning. [keybase/go-keychain#76](https://github.com/keybase/go-keychain/pull/76)

shenqianjin:
> @llqgit I think they remove the support to avoid the warning. [keybase/go-keychain#76](https://github.com/keybase/go-keychain/pull/76)

so, what's our next plan? any update on this?

It's ugly for local run on Mac.

PR #745, opened 2022-03-11
Exposing broker metadata

Shoothzj:
### Motivation
Adapt [PIP 90](https://github.com/apache/pulsar/wiki/PIP-90%3A-Expose-broker-entry-metadata-to-the-client) in go client
allow go client retrieve brokermetadata

### Modifications
- modify `CommandConnect` featureflags, turn `SupportsBrokerEntryMetadata` to true
- add `index()` `brokerPublishTime` methods on `message` interface
- add `Skip` method on bytebuffer

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -498,12 +498,17 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 	pbMsgID := response.GetMessageId()
 
 	reader := internal.NewMessageReader(headersAndPayload)
+	brokerMetadata, err := reader.ReadBrokerMetadata()
+	if err != nil {
+		// todo optimize use more appropriate error codes
+		pc.discardCorruptedMessage(pbMsgID, pb.CommandAck_BatchDeSerializeError)
+		return err
+	}
 	msgMeta, err := reader.ReadMessageMetadata()
 	if err != nil {
 		pc.discardCorruptedMessage(pbMsgID, pb.CommandAck_ChecksumMismatch)
 		return err
 	}
-
 	decryptedPayload, err := pc.decryptor.Decrypt(headersAndPayload.ReadableSlice(), pbMsgID, msgMeta)
 	// error decrypting the payload
 	if err != nil {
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -17,8 +17,8 @@
 # under the License.
 #
 
-ARG GO_VERSION=golang:1.13
-FROM apachepulsar/pulsar:latest as pulsar
+ARG GO_VERSION=golang:1.15
+FROM apachepulsar/pulsar:2.8.1 as pulsar
```
wolfstudy, line 21: Please do not make this modification, we may need to check why the latest docker image cannot pass Action CI, and then create a corresponding issue to fix this problem.

pulsar/consumer_partition.go
```diff
@@ -597,7 +602,18 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 			pc.AckID(msgID)
 			continue
 		}
-
+		var messageIndex *uint64
+		var brokerPublishTime *time.Time
+		if brokerMetadata != nil {
```
wolfstudy, line 607: It looks like we've checked above that the brokerMetadata object is nil by doing the following:

```
brokerMetadata, err := reader.ReadBrokerMetadata()
	if err != nil { // here
		// todo optimize use more appropriate error codes
		pc.discardCorruptedMessage(pbMsgID, pb.CommandAck_BatchDeSerializeError)
		return err
	}
```

Do we still need to double check here? Will someone concurrently modify the value of the brokerMetadata object?

pulsar/consumer_partition.go
```diff
@@ -597,7 +602,18 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 			pc.AckID(msgID)
 			continue
 		}
-
+		var messageIndex *uint64
+		var brokerPublishTime *time.Time
```
wolfstudy, line 606: Maybe instead of using pointers we can use the following definition:

```
		var messageIndex uint64
		var brokerPublishTime time.Time
```

pulsar/internal/commands.go
```diff
@@ -119,6 +121,20 @@ func (r *MessageReader) ReadMessageMetadata() (*pb.MessageMetadata, error) {
 	return &meta, nil
 }
 
+func (r *MessageReader) ReadBrokerMetadata() (*pb.BrokerEntryMetadata, error) {
+	magicNumber := binary.BigEndian.Uint16(r.buffer.Get(r.buffer.ReaderIndex(), 2))
+	if magicNumber != magicBrokerEntryMetadata {
+		return nil, nil
+	}
```
wolfstudy, line 128: Why is the error object returned as nil here? If we don't need this error field, maybe we can cancel it in function definition? Or we can use `errors.new("xxx")` to customize the error message

wolfstudy (changes requested):
Thanks @Shoothzj work for this , just a little comments, please check

Dockerfile
Shoothzj: I already created the issue, https://github.com/apache/pulsar-client-go/issues/748. But, without the modification, let the publish of pulsar broker blocking all the pr ?

pulsar/consumer_partition.go
Shoothzj: @wolfstudy It can be null if broker not enable `brokerMetadata` feature

pulsar/internal/commands.go
Shoothzj: It's not an error, it means broker not enable `brokermetadata` feature

pulsar/consumer_partition.go
Shoothzj: `brokermetadata` can be null if broker not enable this feature

wolfstudy (approved):
LGTM +1

PR #753, opened 2022-03-25
[Issue 664] fix ReconsumeLater panic

### LINKED ISSUE #664: ReconsumeLater cause panic
#### Expected behavior

should be `ReconsumeLater`

#### Actual behavior

Panic when call ReconsumeLater

With version 0.6.0
https://github.com/apache/pulsar-client-go/blob/fe3b7c4e445b3de42974ca692574229ad9099a45/pulsar/consumer_impl.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L481

#### Steps to reproduce

```
		URL: addr,
	})
	if err != nil {
		panic(err)
	}

		Topic:            "test",
		SubscriptionName: "xyz",
	})
	if err != nil {
		panic(err)
	}

        for {
	        select {
	        case cm := <-consumer.Chan():
		        fmt.Println(string(cm.Payload()))
		        cm.ReconsumeLater(cm.Message,time.Second)
	        }
        }
```

#### System configuration
**Pulsar version**: 0.6.0

crossoverJie:
Fixes #664 

### Motivation

Removing possible panic.

### Modifications

Add nil check and more friendly tips.

### Verifying this change


crossoverJie:
@Shoothzj would you mind enabling the workflows and reviewing them?

shileiyu:
Calling ReconsumeLater without a DLQ policy doesn't make sense in most cases.  that sounds some messages could be stuck in an infinite loop forever.

crossoverJie:
@shileiyu Yes, I agree. 

However, there may be careless developers like me who cause the program to run with panic which can even affect the business, this may cut some losses.

shileiyu:
I’m thinking how could we have a better interface to help developers actively avoiding the pitfall.

shileiyu:
@crossoverJie do you think having a default DLQ policy would be a better solution to this issue?

crossoverJie:
@shileiyu It's a good idea.
If `RetryEnable==true` in the Java SDK, a default policy is created.
image

pulsar/consumer_test.go
```diff
@@ -3017,3 +3017,59 @@ func TestEncryptDecryptRedeliveryOnFailure(t *testing.T) {
...
+	//consumer.Ack(msg)
+
+	//for {
+	//	select {
+	//	case cm := <-consumer.Chan():
+	//		fmt.Println(string(cm.Payload()))
+	//		//cm.ReconsumeLater(cm.Message, time.Second)
+	//		cm.Ack(cm.Message)
+	//	}
+	//}
```
wolfstudy, line 3073: please remove this comment

pulsar/consumer_test.go
```diff
@@ -3017,3 +3017,59 @@ func TestEncryptDecryptRedeliveryOnFailure(t *testing.T) {
...
+		log.Fatal(err)
+	}
+
+	// create consumer
+	consumer, err := client.Subscribe(ConsumerOptions{
+		Topic:            topic,
+		SubscriptionName: "my-sub",
+		Type:             Exclusive,
+		//DLQ:              &DLQPolicy{MaxDeliveries: 10, DeadLetterTopic: "dlq-topic"},
+		//RetryEnable:      true,
```
wolfstudy, line 3052: If we don't need this logic, please remove

wolfstudy:
> @shileiyu It's a good idea. If `RetryEnable==true` in the Java SDK, a default policy is created. image

In Go SDK, we have also made the same settings. When retry is enabled, a default DLQ policy will be created. The essential reason for this panic is that we have not enabled retry, but are trying to use ReconsumeLater

```
if options.DLQ == nil {
			options.DLQ = &DLQPolicy{
				MaxDeliveries:    MaxReconsumeTimes,
				DeadLetterTopic:  dlqTopic,
				RetryLetterTopic: retryTopic,
			}
		}
```

pulsar/consumer_test.go
crossoverJie: done

crossoverJie: done

wolfstudy (approved):
LGTM +1

PR #761, opened 2022-04-20
Fix producer unable register when cnx closed

wolfstudy:

### Motivation

At present, it is assumed that there are multiple topics multiplexing the same TCP connection, then any topic may trigger the reconnection operation of this TCP connection. In the actual production environment, we see that a small number of producers will report the following errors:

```
time="2022-04-19T21:43:28+08:00" level=warning msg="Connection closed unable register listener id=41" local_addr="9.xcxx:43051" remote_addr="pulsar://1.xxxx:6650"
```

The specific code logic is as follows:

https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/connection.go#L851-L855

We can see that when we try to register the producer, we will first check the status of the current connection. If the connection is closed, then we will not continue to register the producer object with the connection, but at the same time we will not do any processing and cause some topics to continue to hold the old connection before, causing the sending to time out.

### Modifications

- return an error when unable to register listener

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -53,7 +53,8 @@ type TLSOptions struct {
 }
 
 var (
-	errConnectionClosed = errors.New("connection closed")
+	errConnectionClosed       = errors.New("connection closed")
+	errUnableRegisterListener = errors.New("Unable register listener when con closed")
 )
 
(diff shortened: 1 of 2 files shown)


zymap (approved):

lhotari:
I created a related issue #766 . @wolfstudy @zymap please take a look

PR #765, opened 2022-04-25
Add -c/--max-connections parameter to pulsar-perf-go and set it to 100 by default

lhotari:
### Motivation

- The Java based pulsar-perf uses 100 connections by default

### Modifications

- add support for setting `MaxConnectionsPerBroker` in pulsar-perf-go


zzzming:
+1

merlimat:
The Java default to 100 is not a very good one (I put it a very long time ago).. there usually not an advantage to have these many connections per host. I'd stick to 1 by default (and change Java too! :D )

lhotari:
> The Java default to 100 is not a very good one (I put it a very long time ago).. there usually not an advantage to have these many connections per host. I'd stick to 1 by default (and change Java too! :D )

@merlimat I changed it to 1 here. PTAL

michaeljmarshall (approved):

lhotari:
> The Java default to 100 is not a very good one (I put it a very long time ago).. there usually not an advantage to have these many connections per host. I'd stick to 1 by default (and change Java too! :D )

@merlimat  #15387 is the change for the Java pulsar-perf.

PR #767, opened 2022-04-26
Revert "Fix stuck when reconnect broker (#703)"

### LINKED ISSUE #766: pulsar-client-go marks the connection closed when a "producer close" or "consumer close" message is received.
#### Expected behavior

The connection in pulsar-client-go should not be marked closed when a "producer close" or "consumer close" message is received.

Code in the broker where close consumer / close producer messages are sent:
https://github.com/apache/pulsar/blob/91ec9d58e81fad367f04a70d37a76b53d7388089/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/ServerCnx.java#L2528-L2548

These messages are sent when the broker unloads a topic / namespace bundle. This is used in Pulsar's load balancing.

#### Actual behavior

pulsar-client-go marks the connection closed when a "producer close" or "consumer close" message is received:

https://github.com/apache/pulsar-client-go/blob/e3f625ae8da938f5d147bdddd3a2cadced69c07b/pulsar/internal/connection.go#L822-L849

This leads to problems since the connection in the connection pool cannot be used any more and the connection will get closed and removed from the pool. All consumers and producers sharing the same connection will get closed as a result of this.

This happens when the connection is retrieved from the connection pool:

https://github.com/apache/pulsar-client-go/blob/e3f625ae8da938f5d147bdddd3a2cadced69c07b/pulsar/internal/connection_pool.go#L81-L89

#### Steps to reproduce

while a pulsar-client-go client is connected, unload the topic / namespace:
```
pulsar-admin namespaces unload public/default
```

lhotari:
Fixes #766 

### Motivation

The change made in #703 is invalid. See #766 .

One of the motivation of #703 was to implement similar behavior as there is in Java SDK. The change didn't do that.
In the Java SDK, the concepts (classes) are fairly misleading, and something that might seem to be a connection close isn't one. The ClientHandler is really a reference to a connection in the Java SDK. The ClientCnx (which holds the TCP/IP connection) isn't closed in the Java SDK in ClientHandler.connectionClosed:
https://github.com/apache/pulsar/blob/04aa9e8e51869d1621a7e25402a656084eebfc09/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java#L122-L141

### Modifications

This reverts #703 / commit 1a8432cfd3aa231f8eb3c97171a47eab98a8f20a.

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -823,8 +823,6 @@ func (c *connection) handleCloseConsumer(closeConsumer *pb.CommandCloseConsumer)
 	consumerID := closeConsumer.GetConsumerId()
 	c.log.Infof("Broker notification of Closed consumer: %d", consumerID)
 
-	c.changeState(connectionClosed)
-
 	if consumer, ok := c.consumerHandler(consumerID); ok {
 		consumer.ConnectionClosed()
 		c.DeleteConsumeHandler(consumerID)
@@ -836,8 +834,6 @@ func (c *connection) handleCloseConsumer(closeConsumer *pb.CommandCloseConsumer)
 func (c *connection) handleCloseProducer(closeProducer *pb.CommandCloseProducer) {
 	c.log.Infof("Broker notification of Closed producer: %d", closeProducer.GetProducerId())
 	producerID := closeProducer.GetProducerId()
-
-	c.changeState(connectionClosed)
 
 	producer, ok := c.deletePendingProducers(producerID)
 	// did we find a producer?
--- a/pulsar/internal/connection_pool.go
+++ b/pulsar/internal/connection_pool.go
@@ -80,7 +80,7 @@ func (p *connectionPool) GetConnection(logicalAddr *url.URL, physicalAddr *url.U
 
 		// remove stale/failed connection
 		if conn.closed() {
-			p.log.Infof("Removed connection from pool key=%s logical_addr=%+v physical_addr=%+v",
+			p.log.Debugf("Removed connection from pool key=%s logical_addr=%+v physical_addr=%+v",
 				key, conn.logicalAddr, conn.physicalAddr)
 			delete(p.connections, key)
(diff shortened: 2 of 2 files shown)


michaeljmarshall (approved):

liangyuanpeng (approved):

wolfstudy:
Hello @lhotari 

The processing of #703 is indeed a bit rough. For scenarios where multiple topics reuse the same connection, frequent reconnection does have a big impact.

The #703 originally wanted to solve the problem that the #693 could not reconnect during the reconnection process. In the Go SDK, there is currently a no good way to release the reference of the current connection object like the Java SDK, and then return the real TCP connection to the connection pool. Therefore, the method of closing the connection is temporarily used in the repair of #703.

If there is a more elegant way to handle this, please feel free to communicate.

wolfstudy:
The #703 fix does fix the #693 's problem, but it's probably not an elegant way to do it.

wolfstudy:
Closing the TCP connection here is indeed an inelegant operation, I merge the change first, and then use state management to fix the problem

lhotari:
@billowqiu @zwb1234567 @jiaqingchen Would you mind explaining why you reverted this change in your fork with https://github.com/billowqiu/pulsar-client-go/pull/15 ? It would be useful if you'd share the details so that we could improve the official Apache Pulsar Go client. 
/cc @wolfstudy @zzzming @pgier

billowqiu:
> @billowqiu @zwb1234567 @jiaqingchen Would you mind explaining why you reverted this change in your fork with [billowqiu#15](https://github.com/billowqiu/pulsar-client-go/pull/15) ? It would be useful if you'd share the details so that we could improve the official Apache Pulsar Go client. /cc @wolfstudy @zzzming @pgier

In some cases, not closing the connection causes the producer to enter an abnormal state in which all send requests are fail with "request timed out"

PR #769, opened 2022-04-28
[bugfix] fix stuck when ProducerBlockedQuotaExceededException

### LINKED ISSUE #768: [bugfix] fix stuck when ProducerBlockedQuotaExceededException
#### Expected behavior

fail fast when user call Send

#### Actual behavior

stuck in Send method
image

image

#### Steps to reproduce
1：set a 10MB backlog in pulsar topic
2：Infinite loop call partitionProducer.Send to produce msg to that topic
3：see the sdk output : [Failed to create producer at send PRODUCER request]" error="server error: ProducerBlockedQuotaExceededException: Cannot create producer on topic with backlog quota exceeded" producerID=1

#### System configuration
**Pulsar version**: x.y

billowqiu:

Fixes #768

### Motivation

- fix stuck when ProducerBlockedQuotaExceededException

### Modifications

- store the error when partitionProducer.grabCnx
- fail fast when user call partitionProducer.Send
- add some info log

### Verifying this change

- see the issue #768


pulsar/consumer_impl.go
```diff
@@ -423,6 +423,11 @@ func (c *consumer) Unsubscribe() error {
 }
 
 func (c *consumer) Receive(ctx context.Context) (message Message, err error) {
+	start := time.Now()
+	defer func() {
+		now := time.Now().UnixNano()
+		c.metrics.RecvTime.Observe(float64(now-start.UnixNano()) / 1.0e9)
```
zzzming, line 429: Golang has a more standard way to calculate since time in seconds or any time unit. Here is an example.

time.Since(startTime).Seconds()

zzzming (changes requested):
My concern is there are too many debug level information logged as info, that can be unnecessary.

liangyuanpeng:
> My concern is there are too many debug level information logged as info, that can be unnecessary.

 +1,  We don't need too more info message

PR #772, opened 2022-05-06
Add consumer state check when request commands

wolfstudy:

### Motivation

At present, for different Command requests, when we request the Broker, some commands check the state of the current consumer, and some do not check the state of the current consumer.

If the consumer's state is closing or closed, in fact, we don't need to continue to request the broker, but should directly return the current request.

The current change checks all requests to the broker's command and adds a check on the consumer state.

### Modifications

- add error info for `RequestOnCnxNoWait` func
- add consumer state check when request commands

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -279,6 +279,10 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
 }
 
 func (pc *partitionConsumer) getLastMessageID() (trackingMessageID, error) {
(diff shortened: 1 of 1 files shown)


pulsar/consumer_partition.go
```diff
@@ -477,6 +490,10 @@ func (pc *partitionConsumer) clearMessageChannels() {
 }
 
 func (pc *partitionConsumer) internalAck(req *ackRequest) {
+	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
```
billowqiu, line 493: 这里判断已经卡住了，要在外面Ack接口地方判断

zymap (approved):

Issue #456, opened 2021-01-28
Support Large Message Size (PIP-37 / "chunking")


flowchartsman:
Currently the go client does not seem to support [PIP-37](https://github.com/apache/pulsar/wiki/PIP-37:-Large-message-size-handling-in-Pulsar) which allows for messages to be sent that are larger than the maximum message size by breaking them up on the producer side and re-assembling them in the consumer. This would be a handy feature to have parity with the [Java client](https://github.com/apache/pulsar/pull/4400)

geniusjoe:
@RobertIndie Since chunk feature supported in v0.10 #805 , can we close this issue as completed?

PR #777, opened 2022-05-25
Introduce doneCh for ack error

wolfstudy:

### Motivation

In #456, we support the error information that may occur when the Go SDK side processes the Ack Command, but there is a bug in this implementation. We do not wait for the error information to return directly, so the error obtained in the Ack interface will always be nil.

### Modifications

Introduce `doneCh` to wait for the information of ack error to return

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -325,6 +325,7 @@ func (pc *partitionConsumer) AckID(msgID trackingMessageID) error {
 	}
 
 	ackReq := new(ackRequest)
+	ackReq.doneCh = make(chan struct{})
 	if !msgID.Undefined() && msgID.ack() {
 		pc.metrics.AcksCounter.Inc()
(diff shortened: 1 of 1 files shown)


liangyuanpeng (commented):
I'm not sure if deleting A was the right choice.

wolfstudy:
@freeznet PTAL thanks

freeznet (approved):

xiaofan-luan:
hi @wolfstudy , what would happen if currently we don't wait for ACK but broker side failed to ack?
If we get the error response how do we suppose to handle it?

merlimat (approved):

wolfstudy:
> hi @wolfstudy , what would happen if currently we don't wait for ACK but broker side failed to ack? If we get the error response how do we suppose to handle it?

Here we can look at two situations:

1. If the location of the ack error is an error in the Go SDK, then we can now catch the error and try to call the ack interface again
2. If the ack error information is returned by the Broker, it means that the Broker has an error in processing the ack, and we can also try to ack again, but in general, we may need to further confirm why the Broker side failed and whether it conforms to the predetermined logic.

PR #783, opened 2022-05-31
add shared subscription priority level

fourhu:
<--
### Contribution Checklist
  
  - Name the pull request in the form "[Issue XYZ][component] Title of the pull request", where *XYZ* should be replaced by the actual issue number.
    Skip *Issue XYZ* if there is no associated github issue for this pull request.
    Skip *component* if you are unsure about which is the best component. E.g. `[docs] Fix typo in produce method`.

  - Fill out the template below to describe the changes contributed by the pull request. That will give reviewers the context they need to do the review.
  
  - Each pull request should address only one issue, not mix up code from multiple issues.
  
  - Each commit in the pull request has a meaningful commit message

  - Once all items of the checklist are addressed, remove the above text and this checklist, leaving only the filled out template below.

**(The sections below can be removed for hotfixes of typos)**
-->

Fixes #<xyz>

Master Issue: [#134](https://github.com/apache/pulsar/issues/134)

### Motivation
support priority level for subscription of shared subscription type

### Modifications

consumer options support priority level for subscription of shared subscription type

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: no
  - The schema: no
  - The default values of configurations: yes
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -74,7 +74,7 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 	}
 
 	if options.PriorityLevel <= 0 {
-		options.PriorityLevel = 0
+		return nil, newError(ConsumerOptionsInvalid, "consumer option priority level needs to be >= 0")
 	}
 
 	if options.Type != Shared {
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -47,6 +47,7 @@ type consumer struct {
 	options                   ConsumerOptions
 	consumers                 []*partitionConsumer
 	consumerName              string
+	priorityLevel             int32
 	disableForceTopicCreation bool
 
 	// channel used to deliver message to clients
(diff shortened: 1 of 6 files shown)


pulsar/consumer_impl.go
```diff
@@ -72,6 +73,14 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		return nil, newError(SubscriptionNotFound, "subscription name is required for consumer")
 	}
 
+	if options.PriorityLevel <= 0 {
+		options.PriorityLevel = 0
```
liangyuanpeng, line 77: It should be return error. Ref: https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBuilderImpl.java#L254

pulsar/consumer_impl.go
```diff
@@ -72,6 +73,14 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
...
+	}
+
+	if options.Type != Shared {
```
liangyuanpeng, line 80: Does we need this?

liangyuanpeng (commented):
Thanks for your work and added some comment.

pulsar/consumer_impl.go
fourhu: ok, i will add then error later

pulsar/consumer_impl.go
fourhu: Is PriorityLevel effectived when SubscriptionType is Shared?

merlimat (commented):
We should have a unit test to check the priorities are applied correctly

fourhu:
> or



> 

ok, I will add unit test later.

PR #794, opened 2022-06-23
schema creation and validation functions without panicing

### LINKED ISSUE #793: NewAvroSchema's error lead to panic
#### Expected behavior

NewAvroSchema or NewJSONSchema should not panic a client process. If there is an error, it should return error and allow user to deal with that in the client code.

#### Actual behavior

User should handle any codec error if the NewAvroSchema returns an error reference, although a user can still write a panic recovery procedure but that is not a standard way to do error handling.

This is a recoverable error. So the library should not panic.

zzzming:
Fixes #793 

https://github.com/apache/pulsar-client-go/issues/793

### Motivation

This is an API improvement to avoid panic for any user recoverable error when a schema is created. 

User should handle any codec error if the NewAvroSchema returns an error reference, although a user can still write a panic recovery procedure but that is not a standard way to do error handling.

### Modifications

Introduce three additional functions to create JSON, Proto, and Avro schema with error returned without panic. The existing functions are kept for backward compatibility.

### Verifying this change

  - The public API: (yes)
  - The schema: (no)
  - The wire protocol: (no)

New APIs are introduced to return any errors when schema codec creation fails. The legacy API are kept.

### Documentation

### DIFF
--- a/pulsar/schema.go
+++ b/pulsar/schema.go
@@ -91,18 +91,29 @@ type JSONSchema struct {
 	SchemaInfo
 }
 
+// NewJSONSchema creates a new JSONSchema
+// Note: the function will panic if creation of codec fails
 func NewJSONSchema(jsonAvroSchemaDef string, properties map[string]string) *JSONSchema {
+	js, err := NewJSONSchemaWithValidation(jsonAvroSchemaDef, properties)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


eolivelli (approved):
Lgtm

michaeljmarshall (approved):
LGTM

merlimat (approved):

PR #799, opened 2022-06-30
[issue #752] replace go-rate rate limiter with a buffered channel implementation

### LINKED ISSUE #752: Update license txt files to reflect updated dependency (go-rate)
With this PR https://github.com/apache/pulsar-client-go/issues/725. we upgraded https://github.com/beefsack/go-rate to a version that's MIT licensed. This issue suggests updating the same at two places:

(1) distribution/LICENSE.bin.txt
https://github.com/apache/pulsar-client-go/blob/4f50a678d9030828933e03c1a79ae310c38893ad/distribution/LICENSE.bin.txt#L221-L222

(2) distribution/license/LICENSE-go-rate.txt
https://github.com/apache/pulsar-client-go/blob/4f50a678d9030828933e03c1a79ae310c38893ad/distribution/license/LICENSE-go-rate.txt#L1-L7

The new version of go-rate uses MIT instead of GPL-3

zzzming:
Fixes #752 

### Motivation

The go rate limiter library, beefsack/go-rate, is an MIT license. We need update the license. The PR removes this dependency.

### Modifications

In Go, it is rather simple to implement a rate limiter directly. A buffered channel is used to store a request timestamp as message. The channel receiver check each timestamp (in the message) to sleep on the remaining time duration. 

The per second rate is the size of channel buffer. If the buffer is full, the main producing go routine blocks on producing Pulsar message, until the sleep expires and move to on the next message in the channel thus free up a slot in the channel.

It resembles leaky bucket rate limiting implementation. Because of leveraging go channel, it uses less than 10 lines of code and remove the dependency entirely. Since the implementation only uses one buffered channel, avoid using expensive mutex lock/unlocking and a list by the original library implementation.

### Verifying this change

Since the change only affects the perf utility, I built the tool and verify the producing rate is accurate.

      Removes beefsak/go-rate dependency
  - The public API: ( no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/perf/perf-producer.go
+++ b/perf/perf-producer.go
@@ -103,12 +103,9 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
 	rateLimitCh := make(chan time.Time, produceArgs.Rate)
 	go func(rateLimit int, interval time.Duration) {
 		for {
-			select {
-			case last := <-rateLimitCh:
-				if rateLimit > 0 {
-					remaining := interval - time.Since(last)
-					time.Sleep(remaining)
-				}
+			last := <-rateLimitCh
+			if rateLimit > 0 { // 0 is defined as no limit enforced
+				time.Sleep(interval - time.Since(last))
 			}
 		}
 	}(produceArgs.Rate, time.Second)
--- a/perf/perf-producer.go
+++ b/perf/perf-producer.go
@@ -103,9 +103,9 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
 	rateLimitCh := make(chan time.Time, produceArgs.Rate)
 	go func(rateLimit int, interval time.Duration) {
 		for {
-			last := <-rateLimitCh
+			oldest := <-rateLimitCh
 			if rateLimit > 0 { // 0 is defined as no limit enforced
-				time.Sleep(interval - time.Since(last))
+				time.Sleep(interval - time.Since(oldest))
 			}
 		}
 	}(produceArgs.Rate, time.Second)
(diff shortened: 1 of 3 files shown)


perf/perf-producer.go
```diff
@@ -101,25 +100,26 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
 	payload := make([]byte, produceArgs.MessageSize)
 
 	ch := make(chan float64)
-
-	go func(stopCh <-chan struct{}) {
-		var rateLimiter *rate.RateLimiter
-		if produceArgs.Rate > 0 {
-			rateLimiter = rate.New(produceArgs.Rate, time.Second)
+	rateLimitCh := make(chan time.Time, produceArgs.Rate)
```
pgier, line 103: Since the buffer size here is proportional to the rate, should there be a limit on the max size?  Is there a way to do this with a fixed buffer size?

perf/perf-producer.go
```diff
@@ -101,25 +100,26 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
...
+	go func(rateLimit int, interval time.Duration) {
+		for {
+			last := <-rateLimitCh
```
pgier, line 106: `last` makes me think this is coming from the back of the queue, maybe change `last` to `oldest`?

perf/perf-producer.go
```diff
@@ -101,25 +100,26 @@ func produce(produceArgs *ProduceArgs, stop <-chan struct{}) {
...
 				return
 			default:
 			}
 
-			if rateLimiter != nil {
-				rateLimiter.Wait()
-			}
-
 			start := time.Now()
+			rateLimitCh <- start
```
pgier, line 122: I think this should be skipped completely if there is no limit set.

pgier (approved):
I like the idea of removing the dependency, just left a few in-line comments.

PR #717, opened 2022-01-25
[WIP] [Issue 456] feat: support chunked msg

ShuaoZhang:
Fixes #456 

### Motivation
Allow go clients to send and receive large messages in chunks.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (yes / no)
  - The public API: (yes / no)
  - The schema: (yes / no / don't know)
  - The default values of configurations: (yes / no)
  - The wire protocol: (yes / no)

### Documentation

  - Does this pull request introduce a new feature? (yes / no)
  - If yes, how is the feature documented? (not applicable / docs / GoDocs / not documented)
  - If a feature is not applicable for documentation, explain why?
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

Issue #447, opened 2021-01-22
pulsar-client-go reports message too large before compressing it.


flowchartsman:
I'm getting "message too large" messages on some large JSON documents that, when compressed, are nowhere near the limit.

in `producer_partiton.go`:
```go
	// if msg is too large
	if len(payload) > int(p.cnx.GetMaxMessageSize()) {
		p.publishSemaphore.Release()
		request.callback(nil, request.msg, errMessageTooLarge)
		p.log.WithError(errMessageTooLarge).
			WithField("size", len(payload)).
			WithField("properties", msg.Properties).
			Error()
		p.metrics.PublishErrorsMsgTooLarge.Inc()
		return
	}
```

Correct me if I'm wrong, but this appears to check the payload size before any compression.

Whereas in `ProducerImpl.java`:
```java
int compressedSize = compressedPayload.readableBytes();
            if (compressedSize > ClientCnx.getMaxMessageSize() && !this.conf.isChunkingEnabled()) {
                compressedPayload.release();
```

merlimat:
That's right, the Java and C++ clients are more permissive in that the message go through if it's below the limit after compression. We should also do the same thing in Go. 

Keep in mind, though, that it won't change the "guaranteed" max size, since there is guarantee that the compression will actually reduce the size.

flowchartsman:
> Keep in mind, though, that it won't change the "guaranteed" max size, since there is guarantee that the compression will actually reduce the size.

I am not sure what this means. Of course there is never a guarantee with compression...

merlimat:
Yes, if the max limit is 5MB, it will only guaranteed to accept up to 5MB. Above that, it's just "best effort" to get that to fit in with compression.

flowchartsman:
Okay, so wait, I’m confused. I guess it might help to discuss the intended purpose of maxmessagesize.  If it’s just a knob to adjust cluster performance and storage allocation and such, that’s one thing, but if it is intended as a to guarantee that, for example, a consumer can keep a fixed buffer for handing the payload, then this would seem to be correct behavior, am I right? Are the Java and C++ clients actually *incorrect* in allowing message compression before payload size checking?

freeznet:
[...]

PR #805, opened 2022-07-10
[Issue 456] Support chunking for big messages.

Gleiphir2769:
### Contribution Checklist

Master Issue: [#456](https://github.com/apache/pulsar-client-go/issues/456)

### Motivation

Make pulsar go client support chunking to produce/consume big messages. The earlier implementation ([#717](https://github.com/apache/pulsar-client-go/pull/717)) didn't take into account many details, so I decided to reimplement it.

### Modifications

- Add `internalSingleSend` to send message without batch because batch message will not be received by chunk.
- Moved `BlockIfQueueFull` check from `internalSendAsync` to `internalSend` (`canAddQueue`) to ensure the normal block in chunking.
- Make producer send big messages by chunking.
- Add `chunkedMsgCtxMap` to store chunked messages meta and data.
- Make consumer can obtain chunks and consume the big message.

### Verifying this change

- Add `TestProducerChunking` to verify send big message by chunking.
- Add `message_chunking_test` to verify message chunking.

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - If yes, how is the feature documented? not yet


zzzming (commented):
Based on Java implementation, ConsumerImpl.java, Chunking also requires changes at the consumer side to be able to assembly chunks into the original message. Are you going to add support on the consumer side?

Gleiphir2769:
> Based on Java implementation, ConsumerImpl.java, Chunking also requires changes at the consumer side to be able to assembly chunks into the original message. Are you going to add support on the consumer side?

The consumer side implement is already planned. It will be commited in another PR.

Gleiphir2769:
> Based on Java implementation, ConsumerImpl.java, Chunking also requires changes at the consumer side to be able to assembly chunks into the original message. Are you going to add support on the consumer side?

Hi, the consumer side is implemented here. Looking forward to your review. @zzzming

pulsar/producer_partition.go
```diff
@@ -827,6 +998,10 @@ func (p *partitionProducer) Send(ctx context.Context, msg *ProducerMessage) (Mes
 
 	// wait for send request to finish
 	<-doneCh
+
+	// handle internal error
+	p.internalErrHandle(err)
```
RobertIndie, line 1003: Why do we need to introduce the internal error handler here? Can we handle this error inside the sending operation?

pulsar/producer_partition.go
```diff
@@ -463,12 +465,12 @@ func (p *partitionProducer) Name() string {
 }
 
 func (p *partitionProducer) internalSend(request *sendRequest) {
-	p.log.Debug("Received send request: ", *request)
+	p.log.Debug("Received send request: ", *request.msg)
```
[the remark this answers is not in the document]
Gleiphir2769: Because this call causes a warning about “copies the lock value” . It's just a debug log, so this change can be removed.

pulsar/producer_partition.go
```diff
@@ -843,33 +1018,24 @@ func (p *partitionProducer) internalSendAsync(ctx context.Context, msg *Producer
...
 		ctx:              ctx,
 		msg:              msg,
 		callback:         callback,
 		flushImmediately: flushImmediately,
 		publishTime:      time.Now(),
+		blockCh:          bc,
 	}
 	p.options.Interceptors.BeforeSend(p, msg)
 
-	if p.options.DisableBlockIfQueueFull {
```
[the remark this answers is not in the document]
Gleiphir2769: Yes, you're right. I will change it.

pulsar/producer_partition.go
Gleiphir2769: If we handle this error inside the sending operation, we can only do it in following.
https://github.com/apache/pulsar-client-go/blob/2d5f6fcfa22af4184793a22ca353731a02b54ca9/pulsar/producer_partition.go#L627-L636
I don't think it's obvious, do you have a better idea?

pulsar/producer_partition.go
```diff
@@ -519,29 +526,180 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
...
 
-	sendAsBatch := !p.options.DisableBatching &&
-		msg.ReplicationClusters == nil &&
-		deliverAt.UnixNano() < 0
+	return
+}
 
-	smm := &pb.SingleMessageMetadata{
-		PayloadSize: proto.Int(len(payload)),
+func (p *partitionProducer) genSingleMetaMessage(msg *ProducerMessage,
```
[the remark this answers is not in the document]
Gleiphir2769: Yes, the chunking or single message do not need smm (SingleMessageMetadata), it only used for batching messages.

pulsar/producer_partition.go
```diff
@@ -246,42 +256,14 @@ func (p *partitionProducer) grabCnx() error {
 
 	p.producerName = res.Response.ProducerSuccess.GetProducerName()
 
-	var encryptor internalcrypto.Encryptor
+	//var encryptor internalcrypto.Encryptor
```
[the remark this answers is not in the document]
Gleiphir2769: OK, I will remove it.

pulsar/producer.go
```diff
@@ -172,6 +172,15 @@ type ProducerOptions struct {
 
 	// Encryption specifies the fields required to encrypt a message
 	Encryption *ProducerEncryptionInfo
+
+	// EnableChunking controls whether automatic chunking of messages is enabled for the producer. By default, chunking
+	// is disabled.
+	// Chunking can not be enabled when batching is not closed.
```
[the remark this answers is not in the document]
Gleiphir2769: OK.

pulsar/producer_partition.go
```diff
@@ -519,29 +526,180 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
...
+		}
+		if request.flushImmediately {
+
+			p.internalFlushCurrentBatch()
+
+		}
+	}
+}
+
+func (p *partitionProducer) genMetadata(msg *ProducerMessage, uncompressedSize int) (mm *pb.MessageMetadata) {
```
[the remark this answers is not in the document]
Gleiphir2769: You mean testing `genMetadata`? This is just a simple generator function for `MessageMetadata`, what aspects should I test if I want to test it?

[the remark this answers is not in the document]
Gleiphir2769: Related issue: https://github.com/apache/pulsar/pull/16196/files#
The payload chunk size may be calculated not correctly by two reasons in the Java client. 
1. The MessageMetadata will be updated after computing the payload chunk size, i.e. the actual metadata size would be greater.
2. OpSendMsg#getMessageHeaderAndPayloadSize doesn't exclude all bytes other than the metadata and payload, e.g. the 4 bytes checksum field.

Firstly, in this PR, the order in which `messageMetadata` are se has been concerned to make sure that `totalchunks` can be calculated correctly. Secondly, Since the go client places the data directly into the write buffer instead of creating a data buffer for `OpSendMsg` in advance, Go client doesn't have the problem mentioned in reason 2.

By the way, the unit test for `chunkSize` has been added to message_chunking_test.go.

pulsar/producer_partition.go
```diff
@@ -717,6 +886,7 @@ func (p *partitionProducer) failTimeoutMessages() {
 			for _, i := range pi.sendRequests {
 				sr := i.(*sendRequest)
 				if sr.msg != nil {
+					// todo: it's not correct. the size should be schemaed uncompressed payload size
```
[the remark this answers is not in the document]
Gleiphir2769: OK, it's done.

pulsar/producer_partition.go
```diff
@@ -911,6 +1077,7 @@ func (p *partitionProducer) ReceivedSendReceipt(response *pb.CommandSendReceipt)
 				p.metrics.PublishLatency.Observe(float64(now-sr.publishTime.UnixNano()) / 1.0e9)
 				p.metrics.MessagesPublished.Inc()
 				p.metrics.MessagesPending.Dec()
+				// todo: it's not correct. the size should be schemaed uncompressed payload size
```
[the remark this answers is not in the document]
Gleiphir2769: OK, it's done.

pulsar/producer_partition.go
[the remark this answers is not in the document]
Gleiphir2769: > Why changed here?

Now if I don't change it, the code can not pass golangci-lint: `Error: copylocks: call of p.log.Debug copies lock value: github.com/apache/pulsar-client-go/pulsar.sendRequest contains sync.Once contains sync.Mutex (govet)`.

pulsar/consumer.go
```diff
@@ -189,6 +189,17 @@ type ConsumerOptions struct {
 	// error information of the Ack method only contains errors that may occur in the Go SDK's own processing.
 	// Default: false
 	AckWithResponse bool
+
+	// MaxPendingChunkedMessage sets the maximum pending chunked messages. (default: 100)
+	MaxPendingChunkedMessage int
+
+	// ExpireTimeOfIncompleteChunk sets the expiry time of discarding incomplete chunked message.
+	// The timing accuracy is 100ms level and the default value is 60 * 1000 millis
```
nodece, line 197: Looks like you should a description for the minimum value.

Gleiphir2769: OK, I will add it.

pulsar/internal/timewheel.go
```diff
@@ -0,0 +1,191 @@
...
+package internal
+
+import (
+	"container/list"
+	"log"
+	"time"
+)
+
+// TimeWheel can execute job after waiting given duration
+type TimeWheel struct {
```
nodece, line 27: This code should come from https://github.com/ouqiang/timewheel/blob/master/timewheel.go

nodece, line 27: This code should come from https://github.com/ouqiang/timewheel/blob/master/timewheel.go

pulsar/consumer_impl.go
```diff
@@ -453,16 +465,16 @@ func (c *consumer) Ack(msg Message) error {
...
 // AckID the consumption of a single message, identified by its MessageID
 func (c *consumer) AckID(msgID MessageID) error {
-	mid, ok := c.messageID(msgID)
-	if !ok {
-		return errors.New("failed to convert trackingMessageID")
-	}
-
-	if mid.consumer != nil {
-		return mid.Ack()
+	partition := int(msgID.PartitionIdx())
```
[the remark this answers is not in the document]
Gleiphir2769: > Why not make a method to check the partition or the msgID?

Good suggestion, I will make the method.

> You can improve messageID() method.

`messageID()` converts the `MessageID` to `trackingMessageID`. It will cause the type lost of `chunkMessageID`.

pulsar/internal/timewheel.go
Gleiphir2769: I just realized that this code is from a existed repo. I will remove it.

Gleiphir2769: Same as above

pulsar/consumer_impl.go
```diff
@@ -533,17 +545,15 @@ func (c *consumer) Nack(msg Message) {
...
 
 func (c *consumer) NackID(msgID MessageID) {
-	mid, ok := c.messageID(msgID)
-	if !ok {
-		return
-	}
-
-	if mid.consumer != nil {
-		mid.Nack()
+	partition := int(msgID.PartitionIdx())
```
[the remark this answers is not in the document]
Gleiphir2769: Same as above

pulsar/consumer_impl.go
```diff
@@ -579,12 +589,13 @@ func (c *consumer) Seek(msgID MessageID) error {
 		return newError(SeekFailed, "for partition topic, seek command should perform on the individual partitions")
 	}
 
-	mid, ok := c.messageID(msgID)
-	if !ok {
-		return nil
+	if msgID.PartitionIdx() < 0 || int(msgID.PartitionIdx()) >= len(c.consumers) {
```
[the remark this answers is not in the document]
Gleiphir2769: Same as above

pulsar/consumer_partition.go
```diff
@@ -1444,3 +1564,210 @@ func convertToMessageID(id *pb.MessageIdData) trackingMessageID {
...
+	defer c.mu.Unlock()
+	c.chunkedMsgIDs[chunkID] = msgID
+	c.chunkedMsgBuffer.Write(partPayload.ReadableSlice())
+	c.lastChunkedMsgID = chunkID
+}
+
+func (c *chunkedMsgCtx) firstChunkID() messageID {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.chunkedMsgIDs[0]
```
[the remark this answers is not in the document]
Gleiphir2769: Good suggestion. Original code may cause index out of range error.

pulsar/producer_partition.go
```diff
@@ -519,29 +523,193 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
...
+	var totalChunks int
+	// max chunk payload size
+	var payloadChunkSize int
+	if sendAsBatch || !p.options.EnableChunking {
+		totalChunks = 1
+		payloadChunkSize = int(p._getConn().GetMaxMessageSize())
+	} else {
+		payloadChunkSize = int(p._getConn().GetMaxMessageSize()) - mm.Size()
+		if payloadChunkSize <= 0 {
+			request.callback(nil, msg, errMetaTooLarge)
```
RobertIndie, line 587: Seems to need to release publishSemaphore here.

pulsar/producer.go
```diff
@@ -172,6 +172,15 @@ type ProducerOptions struct {
...
 	Encryption *ProducerEncryptionInfo
+
+	// EnableChunking controls whether automatic chunking of messages is enabled for the producer. By default, chunking
+	// is disabled.
+	// Chunking can not be enabled when batching is enabled.
+	EnableChunking bool
+
+	// MaxChunkSize is the max size of single chunk payload.
+	// It will actually only take effect if it is smaller than broker.MaxMessageSize
+	MaxChunkSize uint
```
RobertIndie, line 183: It's better to keep it consistent with the java client: `ChunkMaxMessageSize`.

pulsar/producer.go
```diff
@@ -172,6 +172,15 @@ type ProducerOptions struct {
...
 	// Encryption specifies the fields required to encrypt a message
 	Encryption *ProducerEncryptionInfo
+
+	// EnableChunking controls whether automatic chunking of messages is enabled for the producer. By default, chunking
+	// is disabled.
+	// Chunking can not be enabled when batching is enabled.
+	EnableChunking bool
+
+	// MaxChunkSize is the max size of single chunk payload.
+	// It will actually only take effect if it is smaller than broker.MaxMessageSize
```
RobertIndie, line 182:
```suggestion
	// It will actually only take effect if it is smaller than the maxMessageSize from the broker.
```

pulsar/producer_partition.go
```diff
@@ -519,29 +523,193 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
...
+		}
+		totalChunks = int(math.Max(1, math.Ceil(float64(compressedSize)/float64(payloadChunkSize))))
 	}
 
-	sendAsBatch := !p.options.DisableBatching &&
-		msg.ReplicationClusters == nil &&
-		deliverAt.UnixNano() < 0
+	// correct limit queue when chunked
+	for i := 0; i < totalChunks-1; i++ {
+		if !p.canAddToQueue(request) {
```
RobertIndie, line 604: Seems the DisableBlockIfQueueFull will not work here.

pulsar/producer_partition.go
```diff
@@ -560,49 +728,60 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
...
+	if err := internal.SingleSend(
+		buffer,
+		p.producerID,
+		sid,
+		mm,
+		payloadBuf,
+		p.encryptor,
+		maxMessageSize,
+	); err != nil {
+		request.callback(nil, request.msg, err)
```
RobertIndie, line 768: Seems need to release the publishSemaphore here.

pulsar/producer_partition.go
```diff
@@ -745,28 +933,34 @@ func (p *partitionProducer) failTimeoutMessages() {
 }
 
 func (p *partitionProducer) internalFlushCurrentBatches() {
-	batchesData, sequenceIDs, callbacks, errors := p.batchBuilder.FlushBatches()
+	batchesData, sequenceIDs, callbacks, errs := p.batchBuilder.FlushBatches()
```
RobertIndie, line 936: Why rename it?

pulsar/producer_partition.go
Gleiphir2769: You're right. And it seems that there are some places in the original code that did not release publishSemaphore.
https://github.com/apache/pulsar-client-go/blob/edd5c71651b79bd35358a51ae3925905ed9f17e1/pulsar/producer_partition.go#L484-L490

pulsar/producer_partition.go
Gleiphir2769: `Errors.Is()` is used to judge internal error. The original parameter name `errors` is conflict with package `errors`

pulsar/producer.go
Gleiphir2769: I think you r right. I have fix it in the new commit.

pulsar/producer.go
Gleiphir2769: Good suggestion. I have fix it.

pulsar/producer_partition.go
Gleiphir2769: I think BlockIfQueueFull shoud not be enabled when `chunking`. When `chunking` is enabled, the sending process will acquire more than one permits (the number equals to chunks). It may never be able to release the permits it has acquired. I create a pr here https://github.com/apache/pulsar/pull/17447.

pulsar/producer_partition.go
Gleiphir2769: Same as above.

pulsar/producer_partition.go
Gleiphir2769: Enable `DisableBlockIfQueueFull` may not a good idea. I fix this problem by moving the `canAddToQueue` before the real send. This solution can ensure that the semaphore acquired for by chunking can be released in time. But it will also cause the sending of a single message (no batch) to waste more CPU resources (e.g. the single message must wait for permits or be discarded after compression).
What do you think?

pulsar/impl_message.go
```diff
@@ -372,3 +386,28 @@ func (t *ackTracker) completed() bool {
...
+		receivedTime: time.Now(),
+	}
+}
+
+func (id chunkMessageID) String() string {
+	return fmt.Sprintf("%s;%s", id.firstChunkID.String(), id.messageID.String())
+}
+
+func (id chunkMessageID) Serialize() []byte {
+	return id.firstChunkID.Serialize()
```
RobertIndie, line 412: Should use `MessageIdData` to serialize the chunk message id.

pulsar/message_chunking_test.go
```diff
@@ -0,0 +1,593 @@
...
+		expectMsg := expectMsgs[i]
+
+		assert.Equal(t, expectMsg, msg.Payload())
+		// ack message
+		err = consumer.Ack(msg)
+		assert.NoError(t, err)
+	}
+}
+
+func TestPublishChunkWithFailure(t *testing.T) {
```
RobertIndie, line 149:
```suggestion
func TestPublishLargeMessagesFailedWithoutChunking(t *testing.T) {
```

RobertIndie: I think we can remove this test because it has been covered in other tests.  Regarding the `testPublishWithFailure` in the java client,  it's hard to simulate `stopBroker` here. I think it's no need to add that test here.

pulsar/message_chunking_test.go
```diff
@@ -0,0 +1,593 @@
...
+	assert.NoError(t, err)
+
+	for i := 1; i < totalMessages; i++ {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+		msg, err := consumer.Receive(ctx)
+		cancel()
+		assert.NoError(t, err)
+		assert.NotNil(t, msg)
+		assert.Equal(t, msgIDs[i].Serialize(), msg.ID().Serialize())
+	}
```
RobertIndie, line 388: Seems that there is no startMessageIdInclusive option in the consumer option currently. But it exists in the reader option. In this test, we should create two consumer(with startMessageIdInclusive true and false), if the startMessageIdInclusive is true, then we can receive msgIDs[1] after the seek operation. Otherwise, we should receive msgIDs[2]. 

The default value of startMessageIdInclusive is false, so the default behavior of the consumer is to receive msgIDs[2] after the seek operation. We need to add this option to the consumer. But if it's too complex to add to this PR, we can add it to a separate PR.

pulsar/message_chunking_test.go
```diff
@@ -0,0 +1,593 @@
...
+				if strings.HasPrefix(msg, "topic-1") {
+					receivedTopic1++
+				} else if strings.HasPrefix(msg, "topic-2") {
+					receivedTopic2++
+				}
+				consumer.Ack(cm.Message)
+			} else {
+				t.Fail()
+			}
+		}
```
RobertIndie, line 552: It's better to add the timeout for this select.

pulsar/producer_partition.go
```diff
@@ -1050,3 +1270,41 @@ func (p *partitionProducer) _getConn() internal.Connection {
...
+			if sr.callback != nil {
+				sr.callback(nil, sr.msg, errSendQueueIsFull)
+			}
+			return false
+		}
+	} else if !p.publishSemaphore.Acquire(sr.ctx) {
+		sr.callback(nil, sr.msg, errContextExpired)
+		sr.blockCh <- struct{}{}
+		return false
+	} else if sr.totalChunks == 0 || sr.totalChunks == 1 || (sr.totalChunks > 1 && sr.chunkID == sr.totalChunks-1) {
```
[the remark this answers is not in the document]
Gleiphir2769: OK, I has removed it. The blockCh will be closed by a method safety.

totalChunks is initialized as 0. Do you think it's better to be initailized as 1?

pulsar/consumer_impl.go
```diff
@@ -91,6 +90,15 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		}
 	}
 
+	if options.MaxPendingChunkedMessage == 0 {
+		options.MaxPendingChunkedMessage = 100
+	}
+
+	// the minimum timer interval is 100ms
```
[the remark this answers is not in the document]
Gleiphir2769: Sorry, this minimum value is original for `timewheel`. Given that `timewheel` has been removed, this minimum value is unneccessary.
I will remove it.

pulsar/consumer_partition.go
```diff
@@ -1475,3 +1609,210 @@ func convertToMessageID(id *pb.MessageIdData) trackingMessageID {
...
+	}
+	u.remove(cmid)
+	return nil
+}
+
+func (u *unAckChunksTracker) nack(cmid chunkMessageID) {
+	ids := u.get(cmid)
+	for _, id := range ids {
+		u.pc.nackTracker.Add(id)
+		u.pc.metrics.NacksCounter.Inc()
```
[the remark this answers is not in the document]
Gleiphir2769: > I think it's better to call u.pc.NackID here. 

My thought at the time was to avoid unnecessary type judgments twice in `u.pc.NackID`.

> we need to maintain in two places.

But I agree with you that saving two type judgments is less beneficial than keeping the code concise.

I will fix it.

pulsar/consumer_partition.go
```diff
@@ -1475,3 +1609,210 @@ func convertToMessageID(id *pb.MessageIdData) trackingMessageID {
...
+	if c.closed {
+		return
+	}
+	if _, ok := c.chunkedMsgCtxs[uuid]; !ok {
+		c.chunkedMsgCtxs[uuid] = newChunkedMsgCtx(totalChunks, totalChunkMsgSize)
+		c.pendingQueue.PushBack(uuid)
+		go c.removeChunkIfExpire(uuid, true, c.pc.options.expireTimeOfIncompleteChunk)
+	}
+	if c.maxPending > 0 && c.pendingQueue.Len() > c.maxPending {
+		go c.removeChunkMessage(uuid, c.pc.options.autoAckIncompleteChunk)
```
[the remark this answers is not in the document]
Gleiphir2769: You r right. I will fix it.

pulsar/consumer_partition.go
```diff
@@ -1475,3 +1609,210 @@ func convertToMessageID(id *pb.MessageIdData) trackingMessageID {
...
+	e := c.pendingQueue.Front()
+	for ; e != nil; e = e.Next() {
+		if e.Value.(string) == uuid {
+			c.pendingQueue.Remove(e)
+			break
+		}
+	}
+}
+
+func (c *chunkedMsgCtxMap) removeChunkMessage(uuid string, autoAck bool) {
```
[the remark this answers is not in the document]
Gleiphir2769: Good suggestion, thx.

pulsar/consumer_partition.go
```diff
@@ -802,6 +886,55 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
...
+		}
+		pc.log.Warnf(fmt.Sprintf(
+			"Received unexpected chunk messageId %s, last-chunk-id %d, chunkId = %d, total-chunks %d",
+			msgID.String(), lastChunkedMsgID, chunkID, totalChunks))
+		pc.chunkedMsgCtxMap.remove(uuid)
+		atomic.AddInt32(&pc.availablePermits, 1)
+		return nil
+	}
+
+	ctx.refresh(chunkID, msgID, compressedPayload)
```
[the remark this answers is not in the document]
Gleiphir2769: Good suggestion, thx.

pulsar/message_chunking_test.go
Gleiphir2769: Yes, the biggest challenge is that it is difficult to `stopBroker` here. So `TestExpireIncompleteChunksTestExpireIncompleteChunks` is not the "real" `expire`. 

Do you think such a test has enough coverage?

pulsar/message_chunking_test.go
Gleiphir2769: Good suggestion, thx.

pulsar/impl_message.go
Gleiphir2769: Done, Thx.

pulsar/message_chunking_test.go
Gleiphir2769: I think startMessageIdInclusive should add to another PR. Because it doesn't matter of chunking implement.

Gleiphir2769:
This PR will fix this issue #447 too.

RobertIndie (approved):

pulsar/producer_partition.go
```diff
@@ -299,6 +278,26 @@ func (p *partitionProducer) grabCnx() error {
 	if err != nil {
 		return err
 	}
+
+	if p.options.DisableBatching {
```
nodece, line 282: The following code looks like more clear:
```
	if !p.options.DisableBatching {
		batcherBuilderType := p.options.BatcherBuilderType
		provider, err := GetBatcherBuilderProvider(batcherBuilderType)
		if err != nil {
			return err
		}
		maxMessageSize := uint32(p._getConn().GetMaxMessageSize())
		p.batchBuilder, err = provider(p.options.BatchingMaxMessages, p.options.BatchingMaxSize,
			maxMessageSize, p.producerName, p.producerID, pb.CompressionType(p.options.CompressionType),
			compression.Level(p.options.CompressionLevel),
			p,
			p.log,
			p.encryptor)
		if err != nil {
			return err
		}
	}
```

nodece (changes requested):
Your PR seems to have made improvements unrelated to chunking, if so, I suggest you make a new PR to improve.

pulsar/consumer_impl.go
```diff
@@ -456,20 +466,15 @@ func (c *consumer) Ack(msg Message) error {
...
 // AckID the consumption of a single message, identified by its MessageID
 func (c *consumer) AckID(msgID MessageID) error {
-	mid, ok := c.messageID(msgID)
-	if !ok {
-		return errors.New("failed to convert trackingMessageID")
-	}
-
-	if mid.consumer != nil {
-		return mid.Ack()
+	if err := c.checkMsgIDPartition(msgID); err != nil {
```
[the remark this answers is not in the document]
Gleiphir2769: `trackingMessageID` does not records chunking infomation. 

For example, `Ack()` a big message need to ack all the chunks of it. Using `trackingMessageID` can not figure out which chunk (messageId) need to be ack.

`trackingMessageID` is designed to tracking batch messages so it shoud not be the messageId type accepted by the method exposed by `partitionConsumer`. I think the better way would be to accept `MessageID ` as the messageId type in partitionConsumer methods. However, only the necessary interfaces have been modified (`Ack`, `NAck` and `Seed`) for the least changes

pulsar/consumer_impl.go
```diff
@@ -540,17 +545,11 @@ func (c *consumer) Nack(msg Message) {
...
 
 func (c *consumer) NackID(msgID MessageID) {
-	mid, ok := c.messageID(msgID)
-	if !ok {
-		return
-	}
-
-	if mid.consumer != nil {
-		mid.Nack()
+	if err := c.checkMsgIDPartition(msgID); err != nil {
```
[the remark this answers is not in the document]
Gleiphir2769: Same as above.

pulsar/consumer_impl.go
```diff
@@ -586,12 +585,11 @@ func (c *consumer) Seek(msgID MessageID) error {
 		return newError(SeekFailed, "for partition topic, seek command should perform on the individual partitions")
 	}
 
-	mid, ok := c.messageID(msgID)
-	if !ok {
-		return nil
+	if err := c.checkMsgIDPartition(msgID); err != nil {
```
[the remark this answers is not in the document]
Gleiphir2769: Same as above.

pulsar/internal/batch_builder.go
```diff
@@ -164,7 +167,11 @@ func (bc *batchContainer) hasSpace(payload []byte) bool {
 		return true
 	}
 	msgSize := uint32(len(payload))
-	return bc.numMessages+1 <= bc.maxMessages && bc.buffer.ReadableBytes()+msgSize <= uint32(bc.maxBatchSize)
+	expectedSize := bc.buffer.ReadableBytes() + msgSize
```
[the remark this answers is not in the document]
Gleiphir2769: L171 compared the `bc.buffer.ReadableBytes() + msgSize` with `bc.maxMessageSize` and the original code does not.

It used to make sure that one batch size does not exceed the `maxMessageSize`. It's a part of correctly calculation whether the message is too large.

By the way, the compare code is too long to be inline.

pulsar/producer_partition.go
Gleiphir2769: Good suggestion, thx.

pulsar/internal/batch_builder.go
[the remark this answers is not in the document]
nodece: Good catch, but using inline looks like more clear.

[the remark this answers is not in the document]
nodece: @Gleiphir2769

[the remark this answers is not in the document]
RobertIndie: @nodece This PR introduces the `maxMessageSize`. I think these changes should be related to this PR.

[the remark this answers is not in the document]
Gleiphir2769: > Good catch, but using inline looks like more clear, so like:

I have tried but this line is too long. It can not pass the golangci-lint

> Good catch, but using inline looks like more clear, so like:
> 
> ```
> return bc.numMessages+1 <= bc.maxMessages && expectedSize <= uint32(bc.maxBatchSize) && expectedSize <= bc.maxMessageSize
> ```

Done, thx.

pulsar/producer_partition.go
```diff
@@ -279,7 +279,7 @@ func (p *partitionProducer) grabCnx() error {
 		return err
 	}
 
-	if !p.options.DisableBatching {
+	if !p.options.DisableBatching && p.batchBuilder == nil {
```
nodece, line 282: Why add `p.batchBuilder == nil`, this code block is not reachable.

nodece (approved):

Gleiphir2769:
> Does the chunking work on shared subscription introduced in Pulsar 2.11

Hi @maraiskruger1980, This PR is finished when pulsar 2.11 has not been released. So it doesn't support shared subscription chunking.

I think I can take some time on it. Welcome to follow the progress.

maraiskruger1980:
That will be great if it can support shared subscription

Gleiphir2769:
> That will be great if it can support shared subscription

Hi @maraiskruger1980. After I checked, it does no limit in consumer when subscription is `shared`. Which means you can safely consume chunking messages in `shared` subscribtion if your pulsar version >= 2.11.

PR #808, opened 2022-07-14
[Issue 387] fix goroutine leak for closing consumers.

### LINKED ISSUE #387: Close consumer message channel when consumer closed.

lots of consumer code maybe like this:

```go
for msg := range consumer.Chan(){
   // consume msg
}
``` 
if not close channel when consumer close, maybe cause goroutine leak.

when close a consumer,close the consumer's message channel.

**Additional context**

crossoverJie:
Fixes #387  

### Motivation

Fix goroutine leak for closing consumers.

### Modifications

Close the `c.messageCh` at the same time as closing the consumer.

**Before fix:**

image
image


pgier (approved):
Is there any chance that c.messageCh can be nil?  Just wondering if we need a nil check around closing it since trying to close a nil channel can cause a panic.
Otherwise LGTM

merlimat (approved):

pgier:
Looks like there is a failure in one of the tests that we're trying to close an already closed channel.  I guess in some cases the channel gets closed and others it doesn't?

crossoverJie:
@pgier 
PTAL, already adjusted.

The `regexConsumer` and `multiTopicConsumer` share the same `messageCh` within the same consumer, so they only need to be closed once.

This is the reason why `chan` was closed repeatedly before.

pulsar/consumer_impl.go
```diff
@@ -563,6 +566,11 @@ func (c *consumer) Close() {
 		}
 		wg.Wait()
 		close(c.closeCh)
+		closed := closeChanSet[c.messageCh]
+		if !closed {
```
zzzming, line 570: closeChanSet is a global variable. The element in the set is never deleted. Would it be more reasonable to create a isClosed (bool) attribute in the `consumer` struct. It can be checked at line 569 and 570 as 
```
if !c.isClosed {
    close(c.messageCh)
}
```
I just do not understand why there is a need to use a global map to track the messageCh channel which will not be GCed. Isn't a leak?

zzzming (changes requested):

pulsar/consumer_impl.go
crossoverJie: That's what I thought before, but in the case of `multiTopicConsumer` it causes the chan to be closed repeatedly.
https://github.com/apache/pulsar-client-go/blob/6a8e7f39aac100a285a2c190186e38b73a5c9d34/pulsar/consumer_multitopic_test.go#L46
image


This is because although the consumers are multiple instances, they use the same `messageCh` and can only be closed once.

So I'd like to record the closed flag for `messageCh` via a global variable.


With regard to the `closeChanSet` leak, we can set it to `nil` in the `client.Close()` method.
https://github.com/apache/pulsar-client-go/blob/6a8e7f39aac100a285a2c190186e38b73a5c9d34/pulsar/client_impl.go#L212

image

Is there a better way?

Gleiphir2769 (commented):
Why not use `Sync.Once()`? Let `closeMsgChOnce` be a attribute in the consumer.
```
c.closeMsgChOnce.Do(func() {
	close(c.messageCh)
})
```
`multiTopicConsumer` uses `newInternalConsumer` to creat. You can add a parameter `closeMsgChOnce` in `newInternalConsumer`, then the consumers also share the `Sync.once()`.
https://github.com/apache/pulsar-client-go/blob/6a8e7f39aac100a285a2c190186e38b73a5c9d34/pulsar/consumer_impl.go#L208-L209

crossoverJie:
@Gleiphir2769 Good suggestion, thx.

Gleiphir2769 (commented):
Could you remove the redundant code comments? Otherwise, LGTM.

pulsar/consumer_impl.go
```diff
@@ -566,11 +566,16 @@ func (c *consumer) Close() {
 		}
 		wg.Wait()
 		close(c.closeCh)
-		closed := closeChanSet[c.messageCh]
-		if !closed {
+		c.closeMsgChOnce.Do(func() {
 			close(c.messageCh)
-			closeChanSet[c.messageCh] = true
-		}
+		})
```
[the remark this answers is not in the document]
crossoverJie: PTAL

Gleiphir2769 (approved):
LGTM

PR #809, opened 2022-07-17
[issue #807] dlq topic producer options

### LINKED ISSUE #807: How to change the BatchingMaxSize of DLQ
The default BatchingMaxSize is 128K, and messages that exceed this size cannot be put into DLQ。

zzzming:
Fixes #807 

### Motivation

To customize producer options for DLQ topics.

### Modifications

Add DLQProducerOptions to consumer's DLQ policy.

### Verifying this change

The existing DLQ test is enhanced to cover customized producer options.

A new producer options field, `DLQProuducerOptions`, is introduced for DLQ policy to govern the producer options.
```
ConsumerOptions{
		Topics:              topics,
		DLQ: &DLQPolicy{
			MaxDeliveries:   3,
			DeadLetterTopic: dlqTopic,
			DLQProducerOptions: ProducerOptions{
				BatchingMaxPublishDelay: 100 * time.Millisecond,
			},
		},
```
  - The public API: yes
  - The schema: no
  - The default values of configurations: yes (the existing lz4 compression type is the default.)
  - The wire protocol: no
 
### Documentation

### DIFF
--- a/pulsar/dlq_router.go
+++ b/pulsar/dlq_router.go
@@ -135,13 +135,13 @@ func (r *dlqRouter) getProducer(schema Schema) Producer {
 	// Retry to create producer indefinitely
 	backoff := &internal.Backoff{}
 	for {
-		opt := r.policy.DLQProducerOptions
+		opt := r.policy.ProducerOptions
 		opt.Topic = r.policy.DeadLetterTopic
 		opt.Schema = schema
 
 		// the origin code sets to LZ4 compression with no options
 		// so the new design allows compression type to be overwritten but still set lz4 by default
-		if r.policy.DLQProducerOptions.CompressionType == NoCompression {
+		if r.policy.ProducerOptions.CompressionType == NoCompression {
 			opt.CompressionType = LZ4
 		}
 		producer, err := r.client.CreateProducer(opt)
--- a/pulsar/dlq_router.go
+++ b/pulsar/dlq_router.go
@@ -135,11 +135,17 @@ func (r *dlqRouter) getProducer(schema Schema) Producer {
 	// Retry to create producer indefinitely
 	backoff := &internal.Backoff{}
 	for {
+		opt := r.policy.DLQProducerOptions
+		opt.Topic = r.policy.DeadLetterTopic
+
+		// the origin code sets to LZ4 compression with no options
+		// so the new design allows compression type to be overwritten but still set lz4 by default
+		if r.policy.DLQProducerOptions.CompressionType == NoCompression {
+			opt.CompressionType = LZ4
+		}
 		producer, err := r.client.CreateProducer(ProducerOptions{
-			Topic:                   r.policy.DeadLetterTopic,
-			CompressionType:         LZ4,
-			BatchingMaxPublishDelay: 100 * time.Millisecond,
-			Schema:                  schema,
+			Topic:  r.policy.DeadLetterTopic,
+			Schema: schema,
 		})
 
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pulsar/dlq_router.go
```diff
@@ -135,12 +135,16 @@ func (r *dlqRouter) getProducer(schema Schema) Producer {
...
-			BatchingMaxPublishDelay: 100 * time.Millisecond,
-			Schema:                  schema,
-		})
+		opt := r.policy.DLQProducerOptions
+		opt.Topic = r.policy.DeadLetterTopic
+		opt.Schema = schema
+
+		// the origin code sets to LZ4 compression with no options
+		// so the new design allows compression type to be overwritten but still set lz4 by default
+		if r.policy.DLQProducerOptions.CompressionType == NoCompression {
```
pgier, line 144: Are there any cases where a user would want to explicitly set to NoCompression?

pgier (commented):
LGTM, just had one in-line question.

pulsar/dlq_router.go
zzzming: The current implementation complies with the default LZ4 compression type in the previous implementation. So user cannot choose NoCompression. If NoCompression is required (I would prefer data has to be sent compressed as the original design),  we might have to introduce another option to maintain backward compatibility.

ne1llee:
DLQ and RLQ are created together, and the same data goes first to RLQ and then to DLQ，so the producer configuration of RLQ also needs to be modified?
https://github.com/apache/pulsar-client-go/blob/6a8e7f39aac100a285a2c190186e38b73a5c9d34/pulsar/retry_router.go#L128-L132

zzzming:
Added code to use the same ProducerOptions for both DLQ and RLQ.

> DLQ and RLQ are created together, and the same data goes first to RLQ and then to DLQ，so the producer configuration of RLQ also needs to be modified?
> 
> https://github.com/apache/pulsar-client-go/blob/6a8e7f39aac100a285a2c190186e38b73a5c9d34/pulsar/retry_router.go#L128-L132

ne1llee:
May I ask which version this pull will be megred in？

zzzming:
@wolfstudy @michaeljmarshall Can you help review this PR? @ne1llee is asking when to merge this?

pgier (approved):
LGTM

michaeljmarshall (approved):
LGTM

pulsar/dlq_router.go
```diff
@@ -135,12 +135,16 @@ func (r *dlqRouter) getProducer(schema Schema) Producer {
 	// Retry to create producer indefinitely
 	backoff := &internal.Backoff{}
 	for {
-		producer, err := r.client.CreateProducer(ProducerOptions{
-			Topic:                   r.policy.DeadLetterTopic,
-			CompressionType:         LZ4,
-			BatchingMaxPublishDelay: 100 * time.Millisecond,
```
michaeljmarshall, line 141: Is there a reason we disabled batching by default?

zzzming: Batching is enabled by default. The `DisableBatching` in the ProducerOptions is false by the default. So the producer for DLQ and RLQ is always batched.

michaeljmarshall: So just to be explicit, this effectively changes the `BatchingMaxPublishDelay` from 100 millis to the producer default, which is first 0 millis and then is overridden to 10 millis in the producer's constructor. Given that this would otherwise require merging producer options, I think this is fine as long as we document it in the release notes.

Issue #814, opened 2022-07-27
Metrics for producer and consumer reconnect failure


zzzming:
In a Pulsar cluster's kubernetes deployment or a deployment with Proxy/LB in the front, we need metrics counter to track the re-connection failure producers and consumers.

When brokers go offline but the proxy/LB is still functioning, TCP connection can still be established but the topic look up failed. `pulsar_client_connections_establishment_errors` counter is not incremented in this case. We have clients built-in failover to other clusters based on counter. Therefore new counters are required to track such failure cases.

This case is to track reconnect failure. When the brokers are offline upon initial connection, the establishment error counter is used. It is not covered by this issue.

Here is the complete client log from a test that demonstrates 1) TCP connection can be established 2) Topic look up failure upon reconnect but we have to differentiate non-existent topic case which the code already does

```$ go run examples/producer/producer.go 
INFO[0000] [Connecting to broker]                        remote_addr="pulsar+ssl://pulsar-aws-useast1.staging.streaming.datastax.com:6651"
INFO[0000] [TCP connection established]                  local_addr="192.168.1.124:57557" remote_addr="pulsar+ssl://pulsar-aws-useast1.staging.streaming.datastax.com:6651"
INFO[0000] [Connection is ready]                         local_addr="192.168.1.124:57557" remote_addr="pulsar+ssl://pulsar-aws-useast1.staging.streaming.datastax.com:6651"
INFO[0000] [Connecting to broker]                        remote_addr="pulsar+ssl://pulsar-aws-useast1.staging.streaming.datastax.com:6651"
INFO[0001] [TCP connection established]                  local_addr="192.168.1.124:57558" remote_addr="pulsar+ssl://pulsar-aws-useast1.staging.streaming.datastax.com:6651"
INFO[0001] [Connection is ready]                         local_addr="192.168.1.124:57558" remote_addr="pulsar+ssl://pulsar-aws-useast1.staging.streaming.datastax.com:6651"
INFO[0001] [Connected producer]                          cnx="192.168.1.124:57558 -> 34.236.188.146:6651" epoch=0 topic="persistent://mingawsuseast1/default/topic2"
INFO[0001] [Created producer]                            cnx="192.168.1.124:57558 -> 34.236.188.146:6651" producerID=1 producer_name=pulsar-aws-useast1-staging-137-34 topic="persistent://mingawsuseast1/default/topic2"

... sending messages
... and stop all the brokers to simulate broker side failure 
[...]

PR #815, opened 2022-07-27
[issue 814] consumer and producer reconnect failure metrics counter

zzzming:
Implement #814 

### Motivation
In a Pulsar cluster's kubernetes deployment or a deployment with Proxy/LB in the front, we need metrics counter to track the re-connection failure producers and consumers.

When brokers go offline but the proxy/LB is still functioning, TCP connection can still be established but the topic look up failed. pulsar_client_connections_establishment_errors counter is not incremented in this case.  Therefore new counters are required to track such failure cases.

### Modifications

Two new counter metrics `pulsar_client_producers_reconnect_failure` and `pulsar_client_consumers_reconnect_failure` will be incremented at the producer_partition and consumer_partition retry failure code block.

Two new counter metrics `pulsar_client_producers_reconnect_max_retry` and `pulsar_client_consumers_reconnect_max_retry` will be incremented at the producer_partition and consumer_partition when either the max retry or max back off is reached.

The existing code logic already covers the case when the topic does not exist. The counters will not be pegged if the topic does not exist. It simply exists from the retry loop at once.

### Verifying this change

This has been verified in the Pulsar cluster deployment with Proxy. We do not have such set up in CI because it's not possible to test with Pulsar standalone mode.

  - The public API: ( no)
  - The schema: (no)
  - The default values of configurations: ( no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -1155,6 +1155,9 @@ func (pc *partitionConsumer) reconnectToBroker() {
 		if maxRetry > 0 {
 			maxRetry--
 		}
+		if maxRetry == 0 || backoff.IsMaxBackoffReached() {
+			pc.metrics.ConsumersReconnectFailure.Inc()
+		}
 	}
 }
 
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -1155,9 +1155,7 @@ func (pc *partitionConsumer) reconnectToBroker() {
 		if maxRetry > 0 {
 			maxRetry--
 		}
-		if maxRetry == 0 || backoff.IsMaxBackoffReached() {
-			pc.metrics.ConsumersReconnectFailure.Inc()
-		}
+		pc.metrics.ConsumersReconnectFailure.Inc()
 	}
 }
 
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -1156,6 +1156,9 @@ func (pc *partitionConsumer) reconnectToBroker() {
 			maxRetry--
 		}
 		pc.metrics.ConsumersReconnectFailure.Inc()
+		if maxRetry == 0 || backoff.IsMaxBackoffReached() {
+			pc.metrics.ConsumersReconnectMaxRetry.Inc()
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pulsar/consumer_partition.go
```diff
@@ -1155,6 +1155,9 @@ func (pc *partitionConsumer) reconnectToBroker() {
 		if maxRetry > 0 {
 			maxRetry--
 		}
+		if maxRetry == 0 || backoff.IsMaxBackoffReached() {
+			pc.metrics.ConsumersReconnectFailure.Inc()
```
michaeljmarshall, line 1159: This will only increment when the max retries have been exhausted, which won't occur when unlimited retries are enabled. I think it'd be reasonable to increment this metric for each failed attempt to reconnect to the broker. 

> Because reconnecting to broker by producer/consumer creation has doubling back off retry, to reduce excessive retry failure noise, these two counters will only incremented by either of two conditions are met.

Is there a reason you view regular (non final) failure as noise? Given that we're willing to log that the consumer failed to connect, I think it's reasonable to increment the metric.

pulsar/producer_partition.go
```diff
@@ -419,6 +419,9 @@ func (p *partitionProducer) reconnectToBroker() {
 		if maxRetry > 0 {
 			maxRetry--
 		}
+		if maxRetry == 0 || backoff.IsMaxBackoffReached() {
+			p.metrics.ProducersReconnectFailure.Inc()
```
michaeljmarshall, line 423: Same comment here, but for the producer case.

michaeljmarshall (commented):
These metrics make sense, and their cardinality is reasonable.

pulsar/consumer_partition.go
zzzming: It will also be triggered when the max backoff is reached. There is a default max. So it will reached in a minute range.

What if there is temporary network failure, it recovers. Do you think if it is up to someone to consume the metrics to decide the course of action?

zzzming: @michaeljmarshall ^

pgier (approved):
LGTM

michaeljmarshall (approved):
LGTM

PR #820, opened 2022-08-01
Support dynamic set go sdk version in connection

wolfstudy:

### Motivation

Currently, when Go SDK creates a connection, the version information set is fixed to Pulsar Go 0.1. When we maintain multiple versions of Go SDK in the production environment, we want to obtain the specific version information used by users through operations such as tcpdump .

### Modifications

Support dynamic set go sdk version in connection

企业微信截图_77defb3a-bf37-4fc1-8640-7e2c076a4f78

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -1061,7 +1061,3 @@ func getClientVersion() string {
 	}
 	return time.Unix(time.Now().Unix(), 0).Format("2006-01-02 15:04:05")
 }
-
-func init() {
-	getClientVersion()
-}
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -23,7 +23,6 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"io/ioutil"
 	"net"
 	"net/url"
@@ -33,6 +32,7 @@ import (
 	"time"
 
 	"github.com/gogo/protobuf/proto"
+	"github.com/sirupsen/logrus"
 
 	"github.com/apache/pulsar-client-go/pulsar/internal/auth"
 	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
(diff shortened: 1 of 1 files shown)


Shoothzj (changes requested):
That's a useful feature. But why not determine version in build time like https://blog.alexellis.io/inject-build-time-vars-golang/ ?

xuthus5:
VERSION file:
```text
v1.1.0
```

logic file:
```go
package main

import (
	_ "embed"
	"fmt"
)

//go:embed VERSION
var version string

func main() {
	fmt.Println("my version: ", version)
}
```

echo: `my version: v1.1.0`

What do you think of this way ?

wolfstudy:
>

wolfstudy:
> That's a useful feature. But why not determine version in build time like https://blog.alexellis.io/inject-build-time-vars-golang/ ?

We cannot expect users to use runtime parameters to determine whether to add the current Go SDK version information when building their own code.

wolfstudy:
> ```go
> embed
> ```

Yeah, `go:embed ` is a goog way, sadly it was only introduced after Go 1.16, in this case we have to require the version of the Go SDK > Go 1.16

freeznet:
I like the `embed` way, and according to https://go.dev/doc/devel/release#policy (quoted below), the current live supported go version should be 1.19, 1.18 and 1.17, so I think it should be fine if we move the pulsar-client-go with 1.16 as the minimum version. Maybe we could call for the community vote. 

> Release Policy
> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including [critical security problems](https://go.dev/security), in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

Shoothzj:
@freeznet Agree, I think we can move to go 1.16

wolfstudy:
> 

OK, Here we launch a round of community voting to decide.

zzzming:
Have you thought about using `ldflags` to pass git commit or git tag which can be the actual version. This will tie into a build system automatically without additional files.
Here is an example.
https://stackoverflow.com/questions/11354518/application-auto-build-versioning/55090713#55090713

PR #825, opened 2022-08-10
NackBackoffPolicy.Next return time.Duration

h-hy:
image

NackBackoffPolicy.Next return int64 type before, but it used as time.Durtion.
image


wolfstudy (approved):
LGTM +1

PR #829, opened 2022-08-16
Enable seeking individual topic partitions

severinson:
Currently, seeking on a consumer with a KeyShared subscription fails. This PR removes an unnecessary check to seek the underlying partitionConsumer.

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -580,6 +580,16 @@ func (c *consumer) Seek(msgID MessageID) error {
 		return nil
 	}
 
+	if mid.partitionIdx < 0 {
+		return newError(SeekFailed, "partitionIdx is negative")
+	}
+	if mid.partitionIdx > int32(len(c.consumers)-1) {
+		return newError(
+			SeekFailed,
+			fmt.Sprintf("partitionIdx is %d, but there are %d partitions", mid.partitionIdx, len(c.consumers)),
+		)
+	}
+
 	return c.consumers[mid.partitionIdx].Seek(mid)
 }
 
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -575,10 +575,6 @@ func (c *consumer) Seek(msgID MessageID) error {
 	c.Lock()
 	defer c.Unlock()
 
-	if len(c.consumers) > 1 {
-		return newError(SeekFailed, "for partition topic, seek command should perform on the individual partitions")
-	}
-
 	mid, ok := c.messageID(msgID)
 	if !ok {
 		return nil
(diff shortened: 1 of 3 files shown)


codelipenghui (commented):
Please also help add a test to cover the new changes.

severinson:
@codelipenghui I've added tests.

pulsar/consumer_impl.go
```diff
@@ -575,15 +575,21 @@ func (c *consumer) Seek(msgID MessageID) error {
...
-	}
-
 	mid, ok := c.messageID(msgID)
 	if !ok {
 		return nil
 	}
 
+	if mid.partitionIdx < 0 {
+		return newError(SeekFailed, "partitionIdx is negative")
+	}
```
codelipenghui, line 585: Will this break the non-partitioned topic message acknowledgment?
The partition index -1 means the message from a non-partitioned topic.

@freeznet @wolfstudy Please help confirm

severinson: Is that really true? The pre-existing code would do the following:
```
return c.consumers[mid.partitionIdx].Seek(mid)
```

This suggests that the partition index is 0 for non-partitioned topics. Otherwise seek() would've always resulted in an index out of bounds panic.

wolfstudy: In Go SDK, the partition index less than 0 is not allowed.

```
// did we receive a valid partition index?
	if partition < 0 || partition >= len(c.consumers) {
		c.log.Warnf("invalid partition index %d expected a partition between [0-%d]",
			partition, len(c.consumers))
		return trackingMessageID{}, false
	}
```

And why do we need to modify the current logic, where the number of consumers is equivalent to the number of partitions, and each partition corresponds to a unique consumer.

The Go SDK currently does not support seek operations on partitioned topics, and this logic is still being implemented.

wolfstudy (changes requested):
Just a few changes, please check

wolfstudy:
If you want to directly perform the seek operation on a partitioned topic, you can refer to the behavior here. https://github.com/apache/pulsar-client-go/pull/782

codelipenghui:
@severinson After checking more details about this part. Before 2.8.0, the Java client only supported seek with the earliest/latest position. https://github.com/apache/pulsar/pull/10033 has introduced a new API https://github.com/apache/pulsar/pull/10033/files#diff-6010f94fead60a5a3ed8aa58a37fe96fdf2cd9c5c1573989cdf89abcb9f9c256R613 which able to provide seek message ID for each partition or topic(For multi-topic/regex consumer)

I think the go client should also follow the same way.

pulsar/consumer_impl.go
severinson: Thanks for the comments. I have a few questions.

> And why do we need to modify the current logic, where the number of consumers is equivalent to the number of partitions, and each partition corresponds to a unique consumer.
I don't understand what you mean. I'm not changing the current logic. I'm just passing through the message id to the partitionConsumer responsible for the partition.

> The Go SDK currently does not support seek operations on partitioned topics, and this logic is still being implemented.
All right. Do you know what work still remains to have the Go client support seeking by partition?

PR #835, opened 2022-08-25
[Issue 833] Fix the availablePermits leak that could cause consumer stuck.

### LINKED ISSUE #833: [BUG] The availablePermits of consumer may leak
#### Overview

When a message is corrupted for some reason when it is received, it will be discarded.

https://github.com/apache/pulsar-client-go/blob/a09460ed6865700f9e5020e9108297d950a00644/pulsar/consumer_partition.go#L605-L609

`partitionConsumer.discardCorruptedMessage()` causes the `availablePermits` leak. Because `availablePermits` increased after `messageReceived()` without error (after `pc.queueCh <- messages`), discarding will cause the `pc.availablePermits++` has no chance to be excuted. So consumer's `availablePermits` may be exhausted causing `receive()` blocking.
 
https://github.com/apache/pulsar-client-go/blob/a09460ed6865700f9e5020e9108297d950a00644/pulsar/consumer_partition.go#L933

In the Java client, the `availablePermits` will be increased when message has discarded.
image

#### Steps to reproduce

It can be reproduced by following.
```
func TestAvailablePermitsLeak(t *testing.T) {
		URL: serviceURL,
	})
	assert.Nil(t, err)
	client.Close()

	topic := fmt.Sprintf("my-topic-test-ap-leak-%v", time.Now().Nanosecond())

	// 1. Producer with valid key name
	p1, err := client.CreateProducer(ProducerOptions{
		Topic: topic,
		Encryption: &ProducerEncryptionInfo{
			KeyReader: crypto.NewFileKeyReader("crypto/testdata/pub_key_rsa.pem",
				"crypto/testdata/pri_key_rsa.pem"),
			Keys: []string{"client-rsa.pem"},
		},
		Schema:          NewStringSchema(nil),
		DisableBatching: true,
	})
	assert.Nil(t, err)
	assert.NotNil(t, p1)

	subscriptionName := "enc-failure-subcription"
	totalMessages := 1000

	// 2. KeyReader is not set by the consumer
	// Receive should fail since KeyReader is not setup
	// because default behaviour of consumer is fail receiving message if error in decryption
		Topic:            topic,
		SubscriptionName: subscriptionName,
	})
	assert.Nil(t, err)

	messageFormat := "my-message-%v"
	for i := 0; i < totalMessages; i++ {
		_, err := p1.Send(context.Background(), &ProducerMessage{
			Value: fmt.Sprintf(messageFormat, i),
		})
		assert.Nil(t, err)
	}

	// 2. Set another producer that send message without crypto.
	// The consumer can receive it correct.
	p2, err := client.CreateProducer(ProducerOptions{
		Topic:           topic,
		Schema:          NewStringSchema(nil),
		DisableBatching: true,
	})
	assert.Nil(t, err)
	assert.NotNil(t, p2)

	_, err = p2.Send(context.Background(), &ProducerMessage{
		Value: fmt.Sprintf(messageFormat, totalMessages),
	})
	assert.Nil(t, err)

	// 3. Discard action on decryption failure. Create a availablePermits leak scenario
	consumer.Close()

	consumer, err = client.Subscribe(ConsumerOptions{
		Topic:            topic,
		SubscriptionName: subscriptionName,
		Decryption: &MessageDecryptionInfo{
			ConsumerCryptoFailureAction: crypto.ConsumerCryptoFailureActionDiscard,
		},
		Schema: NewStringSchema(nil),
	})
	assert.Nil(t, err)
	assert.NotNil(t, consumer)

	// 4. If availablePermits does not leak, consumer can get the last message which is no crypto.
	// The ctx3 will not exceed deadline.
	ctx3, cancel3 := context.WithTimeout(context.Background(), 15*time.Second)
	_, err = consumer.Receive(ctx3)
	cancel3()
	assert.NotEqual(t, true, errors.Is(err, context.DeadlineExceeded),
		"This means the resource is exhausted. consumer.Receive() will block forever.")
}
```

#### System configuration
**Pulsar version**: It doesn't matter.

Gleiphir2769:
Fixes #833 

### Motivation

The availablePermits may leak when message is discarded. Please refer to #833 for details. 

### Modifications

- Wraps a struct availablePermits to atomic increase and reset. 
- Add the permits increasing in `discardCorruptedMessage`.

### Verifying this change

This change added tests and can be verified as follows: `pulsar/consumer_test.go:TestAvailablePermitsLeak()`

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -24,7 +24,6 @@ import (
 	"math"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/gogo/protobuf/proto"
@@ -83,6 +82,15 @@ const (
 	noMessageEntry = -1
 )
 
+type permitsReq int32
+
+const (
+	// reset the availablePermits of pc
+	permitsReset permitsReq = iota
+	// increase the availablePermits
+	permitsInc
+)
+
 type partitionConsumerOpts struct {
 	topic                      string
 	consumerName               string
@@ -128,7 +136,8 @@ type partitionConsumer struct {
 	messageCh chan ConsumerMessage
 
 	// the number of message slots available
-	availablePermits availablePermits
+	availablePermits   int32
+	availablePermitsCh chan permitsReq
 
 	// the size of the queue channel for buffering messages
 	queueSize       int32
@@ -199,18 +208,6 @@ func (s *schemaInfoCache) add(schemaVersionHash string, schema Schema) {
 	s.cache[schemaVersionHash] = schema
 }
 
-type availablePermits struct {
-	permits int32
-}
-
-func (a *availablePermits) inc() int32 {
-	return atomic.AddInt32(&a.permits, 1)
-}
-
-func (a *availablePermits) reset() {
-	atomic.StoreInt32(&a.permits, 0)
-}
-
 func newPartitionConsumer(parent Consumer, client *client, options *partitionConsumerOpts,
 	messageCh chan ConsumerMessage, dlq *dlqRouter,
 	metrics *internal.LeveledMetrics) (*partitionConsumer, error) {
@@ -236,6 +233,7 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 		dlq:                  dlq,
 		metrics:              metrics,
 		schemaInfoCache:      newSchemaInfoCache(client, options.topic),
+		availablePermitsCh:   make(chan permitsReq, 10),
 	}
 	pc.setConsumerState(consumerInit)
 	pc.log = client.log.SubLogger(log.Fields{
@@ -918,7 +916,7 @@ func (pc *partitionConsumer) dispatcher() {
 			messages = nil
 
 			// reset available permits
-			pc.availablePermits.reset()
+			pc.availablePermitsCh <- permitsReset
 			initialPermits := uint32(pc.queueSize)
 
 			pc.log.Debugf("dispatcher requesting initial permits=%d", initialPermits)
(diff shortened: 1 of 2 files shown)


zzzming (commented):
That's a good finding to fix the availablePermits leak.

Can we let dipatcher() to manage the availablePermits since most cases already covered by various signals in dispatcher()'s select cases. The benefits are 1) centralize one function to manage availablePermits 2) to avoid using Atomic or any synchronizing because avialablePermits can only updated one at a time. 
To achieve that, can we have failure cases from MessageReceived() to send a signal to a new channel that dispatcher() listens to that will call your new function increasePermitsAndRequestMoreIfNeed() 

There might be other cases leaking availablePermits. So in the future, those cases can send a sig to this new increasePermits channel. 

What do you think? I think the statement management (i.e. availablePermits) should be done by dispatcher() instead of leaking all over the places then it forces to use synchronization (i.e. atomic) and better use of channel.

Gleiphir2769:
> Can we let dipatcher() to manage the availablePermits since most cases already covered by various signals in dispatcher()'s select cases. The benefits are 1) centralize one function to manage availablePermits 2) to avoid using Atomic or any synchronizing because avialablePermits can only updated one at a time. To achieve that, can we have failure cases from MessageReceived() to send a signal to a new channel that dispatcher() listens to that will call your new function increasePermitsAndRequestMoreIfNeed()

Hi @zzzming. I agree with you that handle the availablePermits in dispatcher().

> 
> There might be other cases leaking availablePermits. So in the future, those cases can send a sig to this new increasePermits channel.
> 

You are right. Chunking also need to increase availablePermits when all the chunks haven't all arrived yet. So this is also a case where need to request more `availablePermits` after return early from `MessageReceived()`.

zzzming:
@Gleiphir2769 `availablePermits` is consumer counter to keep track how many messages the broker can send. Upon the initial connection of a consumer, the consumer sets the initial permits to the queue size and sends to the broker; because of that, the available permits reduces to 0. That is in the dispatcher() function.
For every messages it receives, the available permit is increased by one. This allows the consumer to request more messages once the number of permits over the threshold. This is where you have changed the code in this PR.

`requestPermits` is the number of messages the consumer communicates with the broker that it allows to send. 

These two are very close terminologies. Ultimately, we need one counter to keep track the number of availablePermits. The requestPermits is just an intermediary variable hold the availablePermits value and send to the broker.

zzzming (approved):

Issue #846, opened 2022-09-15
Suggest rewrite `regexConsumer` with inheritance


labuladong:
I'm working on the batch receive feature #840. But before that, I think simplifying the implementation of `regexConsumer` and `multiTopicConsumer` is necessary.

In Java client, [`PatternMultiTopicsConsumerImpl`](https://github.com/apache/pulsar/blob/15a00137012669b0f6fd869ca5dbdf50898242ac/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PatternMultiTopicsConsumerImpl.java) inherits from [`MultiTopicsConsumerImpl`](https://github.com/apache/pulsar/blob/2d6b534ef72824ffb5b2f76568ca8346bbc058c6/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiTopicsConsumerImpl.java). But Go client implements [`multiTopicConsumer`](https://github.com/apache/pulsar-client-go/blob/5108332c9dd4cb454c26804304bffb82eeffc713/pulsar/consumer_multitopic.go#L33) and [`regexConsumer`](https://github.com/apache/pulsar-client-go/blob/5108332c9dd4cb454c26804304bffb82eeffc713/pulsar/consumer_regex.go#L39) separately for now. 

I suggest rewriting [`regexConsumer`](https://github.com/apache/pulsar-client-go/blob/5108332c9dd4cb454c26804304bffb82eeffc713/pulsar/consumer_regex.go#L39) with inheritance just like Java client. If so, further feature development could be easier.

Feel free to correct me if my understanding is inaccurate. cc @wolfstudy @merlimat @nodece

iamseki:
Any plans to get back to work on the batch receive feature? This seems like a really cool feature but is not available in the Go client :cry:

PR #848, opened 2022-09-16
[client] rewrite regexConsumer and fix unit test

labuladong:
Master Issue: #846 

### Modifications

1. Reuse `multiTopicConsumer` in `regexConsumer`.
2. Fix test cases of `regexConsumer`. Due to the system topic `__change_events`, we can't compare the number of consumers to check the correctness:

image

Instead, I try to check the key in the `consumers` map.

image


labuladong:
cc @nodece @RobertIndie @wolfstudy

nodece (approved):
LGTM

pulsar/consumer_regex.go
```diff
@@ -360,13 +235,13 @@ func (c *regexConsumer) unsubscribe(topics []string) {
 	}
 }
 
-func (c *regexConsumer) topics() ([]string, error) {
-	topics, err := c.client.lookupService.GetTopicsOfNamespace(c.namespace, internal.Persistent)
+func topics(client *client, namespace string, pattern *regexp.Regexp) ([]string, error) {
+	topics, err := client.lookupService.GetTopicsOfNamespace(namespace, internal.Persistent)
```
liangyuanpeng, line 239: It should check the `client` for `nil`.

labuladong: This is a private function, we ensure the client is not nil. If the client is nil, I think panic is acceptable.

zzzming: Can we make this function as part of struct method? So that client would not be nil. 
func (c *regexConsumer) topics() ..{
  topics, err: =c.client

zzzming (approved):

pulsar/consumer_regex.go
```diff
@@ -37,73 +35,36 @@ const (
...
-	msgCh chan ConsumerMessage, dlq *dlqRouter, rlq *retryRouter) (Consumer, error) {
+func newRegexConsumer(client *client, opts ConsumerOptions, tn *internal.TopicName, pattern *regexp.Regexp,
+	msgCh chan ConsumerMessage, dlq *dlqRouter, rlq *retryRouter) (*regexConsumer, error) {
+	topics, err := topics(client, tn.Namespace, pattern)
+	if err != nil {
+		return nil, err
+	}
+	mtc, err := newMultiTopicConsumer(client, opts, topics, msgCh, dlq, rlq)
+	if err != nil {
+		return nil, err
```
labuladong, line 57: > Can we make this function as part of struct method?

Hi @zzzming. In the original implementation, `topics` is a method of `regexConsumer` struct. In the new  implementation, we need to get the topics before initialize `regexConsumer`. But I think make this `topics` function as method of `client` struct will be better, I will work on this.

pulsar/consumer_regex.go
```diff
@@ -37,73 +35,36 @@ const (
...
+	}
```
labuladong, line 58: > Can we make this function as part of struct method?

Hi @zzzming. In the original implementation, `topics` is a method of `regexConsumer` struct. In the new implementation, we need to get the topics before initializing `regexConsumer`. But I think make this `topics` function as a method of `client` struct will be better, I will work on this.

labuladong:
Conflict resolved.

nodece:
Ping @labuladong

labuladong:
Conflict resolved. [#805](https://github.com/apache/pulsar-client-go/pull/805/files#diff-541f3a2b60fd1d22abb2b35ce07048aecd7bdcf09e03d22cd5dd3c5d95c814bc) does the same change in `regexConsumer` and `multiTopicConsumer`, so this PR is still compatible.

labuladong:
Can this pr be merged? @nodece @Gleiphir2769 @RobertIndie

nodece (approved):

PR #853, opened 2022-09-28
Parameterize the reconnection option

### LINKED ISSUE #851: Support backoff parameters as optional parameters

The default backoff parameters are alternatively very aggressive (100ms minimum) and not aggressive enough (60 seconds maximum) for high throughput / short response time environments where we are trying to
* Avoid a "thundering herd" / "retry storm" (increase the minimum backoff)
* But need to fail fast (lower the maximum backoff) 

Make these settings parameterizable
https://github.com/apache/pulsar-client-go/blob/bd19458b32ff89206c135cc647336e690e99c32f/pulsar/internal/backoff.go#L36-L38

None I can think of 

**Additional context**

wolfstudy:

Fixes #851

### Motivation

As  #851 said, the reconnection strategy in the default reconnection logic is not universal. In different usage scenarios, we better allow users to customize the reconnection strategy.

### Modifications

Exposed `BackoffPolicy` in Producer and Consumer

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"context"
+	"github.com/apache/pulsar-client-go/pulsar/internal"
 	"time"
 )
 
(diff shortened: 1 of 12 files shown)


wolfstudy:
cc @frankjkelly PTAL

frankjkelly (approved):
LGTM - just confirming that the backoff will be on both the consumer and producer config 
(it was a little confusing to me as in the Java library it is solely on the client https://github.com/apache/pulsar/blob/31203c3952dfb82057ac1fbee4efc10682a5570b/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientBuilderImpl.java#L319-L323 )

(And also thanks so much for the quick turnaround)

frankjkelly:
> Yes, this is where the current Go SDK design is worth optimizing. In the Go SDK, we implemented the reconnect logic to the Producer side and the Consumer side respectively, and did not abstract a unified reconnect logic at the Client level. This leads to the Backoff strategy, we can only divide it into ProducerOptions and ConsumerOptions. In the follow-up, I think we should refactor here and abstract the logic of reconnect to the client side, which is more reasonable

Thanks - sounds good - this is fine for our use-case as is

codelipenghui (approved):

PR #856, opened 2022-09-30
read module version info from golang runtime

pgier:
### Motivation

Instead of maintaining a hard coded version string in the sources, we can get the version of the pulsar-client-go module used to build the client application.

### Modifications

Removed the hard coded version string and replaced it with code that runs at initialization to look up the version of the client golang module dependency that was used to build the app.

### Verifying this change


merlimat (approved):
👍 Probably we should assert in the tests that the version in not `unknown`

nodece (approved):

pgier:
@merlimat I agree it would be good to have a test like that, but it's a little difficult to set up because in order to have a module version, we would need to import this module from another module, but then we can't access the Version variable, because it's in the `internal` package.

PR #869, opened 2022-10-24
Fix PR template

RobertIndie:
### Modifications

Fix PR template


Gleiphir2769:
As the CI log shows, it seems that the `internal.buffer` has data race problem.  https://github.com/apache/pulsar-client-go/actions/runs/3361033527/jobs/5570931674

Maybe we should create a new issue to find out why the data race appeared.

PR #878, opened 2022-10-30
[Issue 877]Fix send stuck

### LINKED ISSUE #877: partitionProducer Send stuck
We use fork version in this line：https://github.com/billowqiu/pulsar-client-go/blob/1dc06ae37025b011ecc345312a65d797b436dedb/pulsar/producer_partition.go#L745

 but the upstream version have the same problem.

#### Expected behavior

partitionProducer.Send should return after success or timeout.

#### Actual behavior

stuck in doneCh.

image

image

#### Steps to reproduce

#### System configuration
**Pulsar version**: x.y

billowqiu:
Fixes #877

### Motivation

Use select to avoid channel stuck. 

### Modifications

- like consumer.Receive func,  use for...select idiom.

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -1035,9 +1035,15 @@ func (p *partitionProducer) Send(ctx context.Context, msg *ProducerMessage) (Mes
 	}, true)
 
 	// wait for send request to finish
-	<-doneCh
-
-	return msgID, err
+	for {
+		select {
+		case <-ctx.Done():
+			isDone.Store(true)
+			return msgID, ctx.Err()
+		case <-doneCh:
+			return msgID, err
+		}
+	}
 }
 
 func (p *partitionProducer) SendAsync(ctx context.Context, msg *ProducerMessage,
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -1035,14 +1035,12 @@ func (p *partitionProducer) Send(ctx context.Context, msg *ProducerMessage) (Mes
 	}, true)
 
 	// wait for send request to finish
-	for {
-		select {
-		case <-ctx.Done():
-			isDone.Store(true)
-			return msgID, ctx.Err()
-		case <-doneCh:
-			return msgID, err
-		}
+	select {
+	case <-ctx.Done():
+		isDone.Store(true)
+		return msgID, ctx.Err()
+	case <-doneCh:
+		return msgID, err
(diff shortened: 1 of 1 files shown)


Gleiphir2769:
Hi, could you upload the full error stack? I think we need to find out real reason for doneCh can not be closed. Because doneCh is never closed means that the callback of `sendRequest` is not be called. This shouldn't happen.

billowqiu:
> Hi, could you upload the full error stack? I think we need to find out the real reason for doneCh can not be closed. Because doneCh is never closed means that the callback of `sendRequest` is not be called. This shouldn't happen.

In the issue link have the stack. I think wait in a channel should with select timeout context to avoid stuck.

Gleiphir2769:
> In the issue link have the stack. I think wait in a channel should with select timeout context to avoid stuck.

You are right, timeout context can avoid `doneCh` stuck. 

But `send()` stuck shouldn't happen. I think add select just skip this problem.

pulsar/producer_partition.go
```diff
@@ -1035,9 +1035,15 @@ func (p *partitionProducer) Send(ctx context.Context, msg *ProducerMessage) (Mes
 	}, true)
 
 	// wait for send request to finish
-	<-doneCh
-
-	return msgID, err
+	for {
```
Gleiphir2769, line 1038:
```suggestion
```
Hi, I don't think a loop is needed here.

Gleiphir2769 (changes requested):

pulsar/producer_partition.go
billowqiu: I also think this loop is useless.

pulsar/producer_partition.go
```diff
@@ -1035,9 +1035,13 @@ func (p *partitionProducer) Send(ctx context.Context, msg *ProducerMessage) (Mes
...
 	// wait for send request to finish
-	<-doneCh
-
-	return msgID, err
+	select {
+	case <-ctx.Done():
+		isDone.Store(true)
+		return msgID, ctx.Err()
+	case <-doneCh:
+		return msgID, err
```
RobertIndie, line 1043: We shouldn't return msgID and err here. They are not assigned here.

RobertIndie (changes requested):

RobertIndie:
Fixed by https://github.com/apache/pulsar-client-go/pull/1053

Issue #879, opened 2022-10-31
Flaky test: Data Race detected in CI test 


zzzming:
CI's `make test` occasionally fails because of data race detected

Based on the stack, the problem seems to be connection is closed while consumer still tries to Ack()

```
=== RUN   TestLargeMessage
time="2022-10-31T01:58:18Z" level=info msg="Connecting to broker" remote_addr="pulsar://localhost:6650"
time="2022-10-31T01:58:18Z" level=info msg="TCP connection established" local_addr="127.0.0.1:51432" remote_addr="pulsar://localhost:6650"
time="2022-10-31T01:58:18Z" level=info msg="Connection is ready" local_addr="127.0.0.1:51432" remote_addr="pulsar://localhost:6650"
time="2022-10-31T01:58:18Z" level=info msg="Connected producer" cnx="127.0.0.1:51432 -> 127.0.0.1:6650" epoch=0 topic="persistent://public/default/my-topic-577261200"
time="2022-10-31T01:58:18Z" level=info msg="Created producer" cnx="127.0.0.1:51432 -> 127.0.0.1:6650" producerID=1 producer_name=standalone-0-172 topic="persistent://public/default/my-topic-577261200"
time="2022-10-31T01:58:18Z" level=info msg="Connected producer" cnx="127.0.0.1:51432 -> 127.0.0.1:6650" epoch=0 topic="persistent://public/default/my-topic-577261200"
time="2022-10-31T01:58:18Z" level=info msg="Created producer" cnx="127.0.0.1:51432 -> 127.0.0.1:6650" producerID=2 producer_name=standalone-0-173 topic="persistent://public/default/my-topic-577261200"
time="2022-10-31T01:58:18Z" level=info msg="Connected consumer" consumerID=1 name=sbiay subscription=chunk-subscriber topic="persistent://public/default/my-topic-577261200"
time="2022-10-31T01:58:18Z" level=info msg="Created consumer" consumerID=1 name=sbiay subscription=chunk-subscriber topic="persistent://public/default/my-topic-577261200"
time="2022-10-31T01:59:18Z" level=warning msg="Connection was closed" cnx="127.0.0.1:51432 -> 127.0.0.1:6650" producerID=1 producer_name=standalone-0-172 topic="persistent://public/default/my-topic-577261200"
[...]

PR #881, opened 2022-11-02
[Fix] Fix the data race of ackReq.err

Gleiphir2769:
Master Issue: #879 

### Motivation

#879 shows ackReq.err has data race problem.

Read
https://github.com/apache/pulsar-client-go/blob/44b64aa712a0d85ce7562134c0f7cf3d21f17ac0/pulsar/consumer_partition.go#L471-L473

Write
https://github.com/apache/pulsar-client-go/blob/44b64aa712a0d85ce7562134c0f7cf3d21f17ac0/pulsar/consumer_partition.go#L702-L707

After ackReq.doneCh has introduced in #777 , I think the final return value of `AckId` can be nil. Because ` ackReq.err` may not necessarily be set before the `AckID` returns. If user need to know the result of ack, just using `AckIDWithResponse` will be fine.

### Modifications

Modify the return value of `AckID`.

### Verifying this change

*(or)*

*(or)*

*(example:)*

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -469,7 +469,7 @@ func (pc *partitionConsumer) AckID(msgID MessageID) error {
 		pc.options.interceptors.OnAcknowledge(pc.parentConsumer, msgID)
 	}
 
-	return ackReq.err
+	return nil
 }
 
 func (pc *partitionConsumer) NackID(msgID MessageID) {


RobertIndie (approved):

nodece:
@Gleiphir2769 I think we should return this error, do you have any other way to fix this?

Gleiphir2769:
> @Gleiphir2769 I think we should return this error, do you have any other way to fix this?

@nodece Hi, thanks for your opinion. I think this error doesn't matter.

This error is set in another goroutine `internalAck`. When `AckId` returns, `internalAck` is not necessarily executing. It means that if `internalAck` really fails, the `ackReq.err` returned by `AckId` may still be nil instead of actual err.

After #777 , `AckId` has been specified as an async method. `AckIDWithResponse` should be called If user really need to know the ack result,.

Gleiphir2769:
@wolfstudy Hi, could you give a review when you r free?

PR #888, opened 2022-11-04
Fix load tls certificate

nodece:
### Motivation

Fix load tls certificate.

This is an incorrect condition. The `CertFile` and `KeyFile` cannot be empty, see the following code that directly read this file:

```
func LoadX509KeyPair(certFile, keyFile string) (Certificate, error) {
	certPEMBlock, err := os.ReadFile(certFile)
	if err != nil {
		return Certificate{}, err
	}
	keyPEMBlock, err := os.ReadFile(keyFile)
	if err != nil {
		return Certificate{}, err
	}
	return X509KeyPair(certPEMBlock, keyPEMBlock)
}
```

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -986,7 +986,7 @@ func (c *connection) getTLSConfig() (*tls.Config, error) {
 		c.log.Debugf("getTLSConfig(): setting tlsConfig.ServerName = %+v", tlsConfig.ServerName)
 	}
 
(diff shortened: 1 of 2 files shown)


tisonkun:
@nodece can you explain a bit more about the condition here instead of a simple word "fix", or add a test case that shows how the original condition fails?

nodece:
@tisonkun The PR description has been updated.

tisonkun (approved):

RobertIndie (approved):

liangyuanpeng (approved):

Issue #183, opened 2020-02-11
support Cumulative ack


adaiboy:
pulsar-client-go use CommandAck_Individual  now,  and we can subscrib a topic in exclusive, so  add CommandAck_Cumulative is usefull.

and it seems not very difficult, may is it already in your scheduler?

merlimat:
The cumulative ack was left out on purpose and the reasoning was (in weighted order):
 * Keep things simple in API
 * Keep things simple in client impl
 * Cumulative ack doesn't provide much more performance/effciency
 * Sequential individual acking will be equivalent

adaiboy:
> The cumulative ack was left out on purpose and the reasoning was (in weighted order):
> 
> * Keep things simple in API
> * Keep things simple in client impl
> * Cumulative ack doesn't provide much more performance/effciency
> * Sequential individual acking will be equivalent

i have doubts with the " doesn't provide much more peformance" than individual ack
- Cumulative ack doesn't provide much more performance/effciency
- Sequential individual acking will be equivalent

i set up a test cluster, with a persistent topic , just one subscription with exclusive type,
though some config in broker not be perfect, but i tuned the configure some days according to the source code
and i make a test with the test cluster
<img width="396" alt="截屏2020-02-1316 09 59" src="https://user-images.githubusercontent.com/3397286/74414209-ea798d80-4e7b-11ea-83a9-c1ccaa9432d8.png">

first,  i modified the consumer-series add AckCumulative to compare cumulative ack with individual ack and make a producer send msg with best effort
then,  use  cumulative consumer to consume entry , and cumulative ack when 10000 entries received, the consumer can catch the producer
lastly, clear all. restart the producer, but start a consumer with individual ack
just in the red frame， the consumer is much slower than producer

merlimat:
@flowchartsman In your example there would be no need to keep the message ids, you could use a Reader and do SeekByTime(): 

```
// Reset the subscription associated with this consumer to a specific message publish time.
	//
	// Note: this operation can only be done on non-partitioned topics. For these, one can rather perform the seek() on
	// the individual partitions.
	//
	// @param timestamp
	//            the message publish time where to reposition the subscription
	//
	SeekByTime(time time.Time) error

```

xiaofan-luan:
Cumulative commit seems to be a reasonable feature.
In our scenario, we use a chan to handle all message. If one ack failed(in current pulsar go client ack didn't throw error, I saw some pr is fixing it), what is the expected behaviour?
[...]

PR #903, opened 2022-12-07
[Issue 183] [Feat] Support cumulative acknowledge

Gleiphir2769:
Master Issue: #183 

### Motivation

Cumulative acknowledgement is a useful feature. Users can use this feature to ack messages in the stream up to (and including) provided message. 

Issue #183 shows more details.

### Modifications

- Add two api `AckCumulative` and `AckIDCumulative` for `Consumer`.

``` golang

// AckCumulative the reception of all the messages in the stream up to (and including)
// the provided message.
AckCumulative(msg Message) error

// AckIDCumulative the reception of all the messages in the stream up to (and including)
// the provided message, identified by its MessageID
AckIDCumulative(msgID MessageID) error

```

- Add the `AckCumulative` and `AckIDCumulative` implementation for `consumer`, `multiTopicConsumer`, `regexConsumer` and `mockConsumer`.

- Add the unit test `TestConsumerNoBatchCumulativeAck` `TestConsumerBatchCumulativeAck` `TestCumulativeAckWithResponse` for cumulative ack in `consumer_test.go`.

### Verifying this change

  - The public API: yes

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -231,13 +231,13 @@ type Consumer interface {
 	// AckID the consumption of a single message, identified by its MessageID
 	AckID(MessageID) error
 
-	// CumulativeAck the reception of all the messages in the stream up to (and including)
+	// AckCumulative the reception of all the messages in the stream up to (and including)
 	// the provided message.
-	CumulativeAck(msg Message) error
+	AckCumulative(msg Message) error
 
-	// CumulativeAckID the reception of all the messages in the stream up to (and including)
+	// AckIDCumulative the reception of all the messages in the stream up to (and including)
 	// the provided message, identified by its MessageID
-	CumulativeAckID(msgID MessageID) error
+	AckIDCumulative(msgID MessageID) error
 
 	// ReconsumeLater mark a message for redelivery after custom delay
 	ReconsumeLater(msg Message, delay time.Duration)
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -231,6 +231,14 @@ type Consumer interface {
 	// AckID the consumption of a single message, identified by its MessageID
 	AckID(MessageID) error
 
+	// CumulativeAck the reception of all the messages in the stream up to (and including)
+	// the provided message.
+	CumulativeAck(msg Message) error
+
+	// CumulativeAckID the reception of all the messages in the stream up to (and including)
+	// the provided message, identified by its MessageID
+	CumulativeAckID(msgID MessageID) error
+
 	// ReconsumeLater mark a message for redelivery after custom delay
 	ReconsumeLater(msg Message, delay time.Duration)
 
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -731,6 +731,72 @@ func TestConsumerAck(t *testing.T) {
 	}
 }
 
+func TestConsumerCumulativeAck(t *testing.T) {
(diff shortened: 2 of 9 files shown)


Gleiphir2769:
Hi, @RobertIndie @nodece,  could you give a review when you are free? Thanks.

pulsar/consumer.go
```diff
@@ -231,6 +231,14 @@ type Consumer interface {
 	// AckID the consumption of a single message, identified by its MessageID
 	AckID(MessageID) error
 
+	// CumulativeAck the reception of all the messages in the stream up to (and including)
+	// the provided message.
+	CumulativeAck(msg Message) error
+
+	// CumulativeAckID the reception of all the messages in the stream up to (and including)
+	// the provided message, identified by its MessageID
+	CumulativeAckID(msgID MessageID) error
```
BewareMyPower, line 240: Maybe it's better to use `AckIDCumulative` as the method name? Like the `acknowledge` and `acknowledgeCumulative` methods in other clients.

Gleiphir2769: Good suggestion. I will rename the two methods.

pulsar/impl_message_test.go
```diff
@@ -80,6 +80,12 @@ func TestAckTracker(t *testing.T) {
 
 	}
 	assert.Equal(t, true, tracker.completed())
+
+	// test large number 1000 cumulative
+	tracker = newAckTracker(1000)
+
+	assert.Equal(t, true, tracker.ackCumulative(999))
```
BewareMyPower, line 87: It's better to add a test to verify `ackCumulative` for 0 to 998 returns false.

```golang
        tracker = newAckTracker(1000)
        for i := 0; i < 999; i++ {
                assert.False(t, tracker.ackCumulative(i))
        }
        assert.False(t, tracker.completed())
```

pulsar/consumer_test.go
```diff
@@ -731,6 +732,161 @@ func TestConsumerAck(t *testing.T) {
...
+
+	for i := 0; i < 2*N; i++ {
+		msg, err := consumer.Receive(ctx)
+		assert.Nil(t, err)
+		assert.Equal(t, fmt.Sprintf("msg-content-%d", i), string(msg.Payload()))
+
+		if i == N-1 {
+			// cumulative acks the first half of messages
+			consumer.AckCumulative(msg)
+		}
```
BewareMyPower, line 866: I think you can test the case `i == N` to cover the following case

```
        } else if !trackingID.tracker.hasPrevBatchAcked() {
                // get previous batch message id
                ackReq.msgID = trackingID.prev()
                trackingID.tracker.setPrevBatchAcked()
```

Not sure if there is a way to reduce duplicated code. Maybe you can create another consumer to call `AckCumulative` when `i == N`?

```golang
if i == N - 1 {
    consumer.AckCumulative(msg)
} else if i == N {
    consumer2.AckCumulative(msg)
}
```

pulsar/impl_message_test.go
Gleiphir2769: Good suggestion! I will add test code to cover the cumulative tracker.

pulsar/consumer_test.go
Gleiphir2769: > I think you can test the case i == N to cover the following case

You are right. I will modify `TestConsumerBatchCumulativeAck` to cover this case.

BewareMyPower (approved):

Issue #906, opened 2022-12-08
Include original messageId and topic as part of properties for DLQ message.


GPrabhudas:
**Is your feature request related to a problem? Please describe.**
When a message is sent to DLQ topic and consumed later, currently there is no way to identify original messaged and topic of the message.

**Describe the solution you'd like**
Whenever message is being sent to DLQ topic, include messageId and topic of the original message as properties of DLQ message.
Java client already does it as follows 👇 
https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L2025
https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L715

RobertIndie:
Fixed by #907

PR #907, opened 2022-12-08
add messageId and topic as props of DLQ message

GPrabhudas:
addresses issue #906

### DIFF
--- a/pulsar/reader_test.go
+++ b/pulsar/reader_test.go
@@ -430,6 +430,14 @@ func (id *myMessageID) PartitionIdx() int32 {
 	return id.PartitionIdx()
 }
 
+func (id *myMessageID) String() string {
+	mid, err := DeserializeMessageID(id.data)
+	if err != nil {
+		return ""
+	}
+	return fmt.Sprintf("%d:%d:%d", mid.LedgerID(), mid.EntryID(), mid.PartitionIdx())
+}
+
 func TestReaderOnSpecificMessageWithCustomMessageID(t *testing.T) {
 	client, err := NewClient(ClientOptions{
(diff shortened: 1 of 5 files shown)


pulsar/reader_test.go
```diff
@@ -430,6 +430,14 @@ func (id *myMessageID) PartitionIdx() int32 {
 	return id.PartitionIdx()
 }
 
+func (id *myMessageID) String() string {
+	mid, err := DeserializeMessageID(id.data)
```
nodece, line 434:
```suggestion
        return fmt.Sprintf("%d:%d:%d", id.LedgerID(), id.EntryID(), id.PartitionIdx())
```

GPrabhudas:
```go
type myMessageID struct {
	data []byte
}

func (id *myMessageID) Serialize() []byte {
	return id.data
}

func (id *myMessageID) LedgerID() int64 {
	return id.LedgerID()
}

func (id *myMessageID) EntryID() int64 {
	return id.EntryID()
}

func (id *myMessageID) BatchIdx() int32 {
	return id.BatchIdx()
}

func (id *myMessageID) PartitionIdx() int32 {
	return id.PartitionIdx()
}
```

This is how ☝️ struct `myMessageID` implements interface `MessageID`. 

If I use `return fmt.Sprintf("%d:%d:%d", id.LedgerID(), id.EntryID(), id.PartitionIdx())` it would result into `fatal error: stack overflow` due to infinite recursive calls

nodece (approved):

wolfstudy (approved):
LGTM +1

PR #908, opened 2022-12-09
[improve][ci] add -failfast flag to improve test process

labuladong:
Even if one test case failed, all other tests will still run anyway. Add `failfast` can stop other test cases if any test failed.

FYI: https://stackoverflow.com/questions/32046192/stop-on-first-test-failure-with-go-test

### DIFF
--- a/scripts/run-ci.sh
+++ b/scripts/run-ci.sh
@@ -31,7 +31,7 @@ go build -o bin/pulsar-perf ./perf
 
 scripts/pulsar-test-service-start.sh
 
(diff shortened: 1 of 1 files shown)


nodece:
Usually, we disable the `failfast`, we can check all the test results.

labuladong:
It's hard for me to find the failed cases in lots of ci output, so I think `failfast` would be a solution. Now I found search keyword `FAIL:` can locate the failed test case, so seems `failfast` is not needed as you said. I'll close this pr.

PR #910, opened 2022-12-13
[improve][ci] change token name to `GITHUB_TOKEN`

### LINKED ISSUE #909: [CI] pulsar bot not work in this repo due to `GO_CLIENT_BOT_TOKEN` missing
I found that `/pulsarbot rerun-failure-checks` didn't work in this repo, there is some thing wrong when parse pr status:

https://github.com/apache/pulsar-client-go/actions/runs/3655434630/jobs/6176758886#step:4:8

~~The root cause is this repo didn't set `GO_CLIENT_BOT_TOKEN`~~:

https://github.com/apache/pulsar-client-go/blob/055b00b83ccfef4ea12a593e3678cb4bdd18311c/.github/workflows/bot.yml#L19-L25

~~see https://github.com/labuladong/pulsar-client-go/actions/runs/3662207400/jobs/6191157828#step:4:27 , `token` is empty so we cannot read pr info, and cannot trigger pulsar bot to rerun the failed job~~.

In my forked repo, I may not read the original `GO_CLIENT_BOT_TOKEN`, so I read the empty `GO_CLIENT_BOT_TOKEN`. Maybe we need someone who has upstream permission to fix this issue.

labuladong:
Fixes #909 

### Modifications

Change token to `GITHUB_TOKEN`.

After changing, the pulsar bot works fine. Example here:

https://github.com/labuladong/pulsar-client-go/actions/runs/3681253024/jobs/6227745474#step:4:27

### DIFF
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -20,6 +20,6 @@ jobs:
         id:   pulsarbot
         if: github.event_name == 'issue_comment' && startsWith(github.event.comment.body, '/pulsarbot')
         env:
-          GITHUB_TOKEN: ${{ secrets.GO_CLIENT_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           BOT_TARGET_REPOSITORY: apache/pulsar-client-go
         uses: apache/pulsar-test-infra/pulsarbot@master


tisonkun (approved):
Thank you! Although I suppose we can omit this env var at all and let the action read from preset ENV. But since it works and does no harm, let's move forward with this patch first.

tisonkun:
A possible issue is that this default token may not have the write permission required and thus fail to write comment/update status on demand. You can check it (it may work differently from you personal fork and in the upstream) or we just give it a try.

labuladong:
> A possible issue is that this default token may not have the write permission required.

I found the bot in the pulsar repo also uses another token:

image

So maybe you are right. I guess the default token doesn't have permission to restart the CI task.

labuladong:
Yes, the behavior in the upstream repo may be different from my fork. So this pr may not work. Do you have any ideas to fix the bug about the CI bot? @tisonkun

tisonkun:
This change doesn't make the case worse. Let's give it a try.

If it still won't work, we can find other approach or just remove the workflow since we actually live well without this functionality always.

tisonkun:
```
rerun-failed-jobs for 'CI'. Follow progress at https://github.com/apache/pulsar-client-go/actions/runs/2812684476
{
  "message": "Unable to retry this workflow run because it was created over a month ago",
  "documentation_url": "https://docs.github.com/rest/reference/actions#re-run-workflow-failed-jobs"
}
Failed.
```

https://github.com/apache/pulsar-client-go/actions/runs/4698390864/jobs/8330614311

It seems default token doesn't have enough permission to trigger rerun. Although we can file an INFRA ticket to add one, but IIRC INFRA will reject configure personal access token as we do before - we may be an actual bot account.

I'm OK to leave it as is or delete the workflow file since we can actually live well without this functionality.

PR #911, opened 2022-12-13
[build] update go version to 1.18

pgier:
### Motivation

Upgrade module golang version from 1.15 to 1.18.  Golang 1.18 is the oldest currently supported version.  See: https://go.dev/doc/devel/release.  Client applications using go 1.15 will still be able to compile against this module, but we should recommend that they upgrade their golang version or stick with a previous release of pulsar-client-go, e.g. 0.9.0.

Additional motivation for this is that I'd like to upgrade most/all of the dependencies in a separate commit, and some of them are likely not fully compatible with golang 1.15

### Modifications

Updated the golang version in go.mod and ran `go mod tidy`.

### Verifying this change

- [X] Make sure that the change passes the CI checks.

This change is covered by existing CI which runs the build using various golang versions.

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

  - If a feature is not applicable for documentation, explain why? Added a minor doc update to the README


README.md
```diff
@@ -36,7 +36,9 @@ CGo based library.
 
 ## Requirements
 
-- Go 1.15+
+- Go 1.18+
```
nodece, line 39:
```suggestion
- Go 1.18+  
```

Should I append two spaces?

pgier: Why would you append spaces here?

zzzming (approved):
LGTM

tisonkun:
> Client applications using go 1.15 will still be able to compile against this module

@pgier I'm curious about this statement. If we follow write some code depending on new added APIs, how can a lib compile in 1.15 compile with us as a dependency? Said `os.ReadFile`.

pgier:
@tisonkun You are right at that point we will no longer be compatible with go 1.15.  We should see a failure in the 1.15 integration tests at the point when a PR includes an incompatible change, and we would have to either not accept the change, or remove the 1.15 tests.  My point was just that this change by itself doesn't make us incompatible with 1.15.

I'll work on upgrading dependencies next and see if anything fails, and we can decide whether it's worth upgrading specific incompatible deps.

README.md
```diff
@@ -36,7 +36,9 @@ CGo based library.
...
+  Note that while this library should work with golang versions as early as go 1.15,
+  any bugs specific to Golang versions earlier than Go 1.18 may not be fixed.
```
tisonkun, line 41:
```suggestion

> **Note**:
>
> While this library should work with Golang versions as early as 1.15, any bugs specific to versions earlier than 1.18 may not be fixed.
```

tisonkun (approved):
Good to go. One comment inline.

nodece (approved):

Issue #863, opened 2022-10-15
why internal Provider？


tianjingle:
When I customized a authentication , I went to implement the `Provider` interface and found that the internal was used in Pulsar-Client-Go, so I couldn't customize the authentication. How do I implement custom authentication for this piece? Looking forward to your answer. Thank you.

```
func NewAuthenticationIAM(ak string, sk string) auth.Provider {
	var provider auth.Provider
	provider = &IAMProvider{
		ak:        ak,
		sk:        sk
	}
	return provider
}
```

PR #914, opened 2022-12-19
Move out the auth package from internal

nodece:
Fix https://github.com/apache/pulsar-client-go/issues/863

### Motivation

The auth package is internal, the developer cannot implement their auth provider, so we need to move out the auth package from internal.

### Modifications

Move out the auth package from `pulsar/internal/auth` to `pulsar/auth`.


liangyuanpeng (approved):

aristotelism:
@liangyuanpeng hi , is it possible to create a release with this change? Thanks

nodece:
Hi @aristotelism, you can use the master codebase by `go get github.com/apache/pulsar-client-go@master`.

Issue #918, opened 2022-12-21
Refactor the `clearMessageQueuesCh` and `clearQueueCh` in `dispatcher`


Gleiphir2769:
Now, there are two channels related to clearing messages from `queueCh` and `messageCh` when consume position is reset in the `consumer.dispatcher()`.

https://github.com/apache/pulsar-client-go/blob/d92fb1407d3d39c8a498dd7c7abdc0bbb3fc7e1a/pulsar/consumer_partition.go#L1150-L1164

https://github.com/apache/pulsar-client-go/blob/d92fb1407d3d39c8a498dd7c7abdc0bbb3fc7e1a/pulsar/consumer_partition.go#L1168-L1175

The first channel `clearQueueCh` is used to clear legacy messages `queueCh` when connection to broker is rebuild. This clear will reset the `startMessage`.

The second channel `clearMessageQueuesCh` is used to clear legacy messages in `queueCh` and `messageCh` when `Seek` is invoked.

There are some problems.
- The `clearQueueCh` and `clearMessageQueuesCh` have the same effect. It's enough to keep one.
- `clearQueueCh` does not clear the legacy messages in `messageCh`, which users can may get the legacy messages after reconnecting to broker.
- `clearMessageQueuesCh` is no need to send a flow rpc after clear. Because the permits in broker will not be clear after `Seek`.
https://github.com/apache/pulsar-client-go/blob/d92fb1407d3d39c8a498dd7c7abdc0bbb3fc7e1a/pulsar/consumer_partition.go#L1181-L1185
- `clearMessageQueuesCh` does not reset the `startMessage`, which causes the behavior of `Seek()` is different with Java client.

Overall, I think it's a good idea to refactor this part of code. 

Hi @RobertIndie @nodece , if you think it's need to be refactor, could you assign it to me?

Gleiphir2769:
Closed by 921.

PR #921, opened 2022-12-21
[Issue 918] [Refactor] Remove the `clearMessageQueuesCh` in `partitionConsumer.dispatcher()`

Gleiphir2769:
Master Issue: #918 

### Motivation

The two chanel `clearMessageQueuesCh` and `clearQueueCb` just need to keep one.

For more details please check #918.

This PR does not ony aim to clean up, but also fix the potential bug in `clearMessageQueuesCh`.

For example, the `clearMessageQueuesCh` do the jod including clearing messageCh. But it may cause problem when `SeekByTime` invoked on partition topic.

https://github.com/apache/pulsar-client-go/blob/1d3499a18d526b4b1aef0bdbbc54ac812b8ae0c0/pulsar/consumer_impl.go#L614-L626

https://github.com/apache/pulsar-client-go/blob/1d3499a18d526b4b1aef0bdbbc54ac812b8ae0c0/pulsar/consumer_partition.go#L1168-L1175

When consume the partition topic, all the `partitionConsumer` share the same `messageCh`. After `SeekByTime` on partitioned topic, the messageCh may be cleared more than one time which will cause the messages losing. Suppose there is such a situation, partitionConsumer-1 has cleared its messageCh and queueCh. When partitionConsumer-2 do the clear job, it can also exec this logic.

https://github.com/apache/pulsar-client-go/blob/1d3499a18d526b4b1aef0bdbbc54ac812b8ae0c0/pulsar/consumer_partition.go#L1172-L1174

But `messageCh` is a share chan, partitionConsumer-1 may received new messages and put them to `messageCh` at this moment. There is such a possibility that partitionConsumer-2 **cleared the new messages** from `messageCh`.

### Modifications

- Remove the `clearMessageQueuesCh` in `partitionConsumer.dispatcher()`
- Modify the `clearQueueCb` in `partitionConsumer.dispatcher()`

### Verifying this change

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -138,15 +138,14 @@ type partitionConsumer struct {
 	// the size of the queue channel for buffering messages
 	queueSize       int32
 	queueCh         chan []*message
-	startMessageID  trackingMessageID
+	startMessageID  atomicMessageID
 	lastDequeuedMsg trackingMessageID
 
-	eventsCh             chan interface{}
-	connectedCh          chan struct{}
-	connectClosedCh      chan connectionClosed
-	closeCh              chan struct{}
-	clearQueueCh         chan func(id trackingMessageID)
-	clearMessageQueuesCh chan chan struct{}
+	eventsCh        chan interface{}
+	connectedCh     chan struct{}
+	connectClosedCh chan connectionClosed
+	closeCh         chan struct{}
+	clearQueueCh    chan func(id trackingMessageID)
 
 	nackTracker *negativeAcksTracker
 	dlq         *dlqRouter
@@ -190,6 +189,24 @@ func (p *availablePermits) inc() {
 
 func (p *availablePermits) reset() {
 	atomic.StoreInt32(&p.permits, 0)
+}
+
+// atomicMessageID is a wrapper for trackingMessageID to make get and set atomic
+type atomicMessageID struct {
+	msgID trackingMessageID
+	sync.RWMutex
+}
+
+func (a *atomicMessageID) get() trackingMessageID {
+	a.RLock()
+	defer a.RUnlock()
+	return a.msgID
+}
+
+func (a *atomicMessageID) set(msgID trackingMessageID) {
+	a.Lock()
+	defer a.Unlock()
+	a.msgID = msgID
 }
 
 type schemaInfoCache struct {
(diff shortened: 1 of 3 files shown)


pulsar/consumer_partition.go
```diff
@@ -1134,37 +1144,19 @@ func (pc *partitionConsumer) dispatcher() {
...
-			initialPermits := uint32(pc.queueSize)
-
-			pc.log.Debugf("dispatcher requesting initial permits=%d", initialPermits)
-			// send initial permits
-			if err := pc.internalFlow(initialPermits); err != nil {
-				pc.log.WithError(err).Error("unable to send initial permits to broker")
-			}
-
-			close(doneCh)
+			clearQueueCb(nextMessageInQueue)
```
nodece, line 1159: Do you forget the reset available permits?

Gleiphir2769: > Do you forget the reset available permits?

Hi. @nodece, I don't think it's necessary to reset available permits. In the user side, if permits over the threshold, `internalFlow` will be invoked. 

https://github.com/apache/pulsar-client-go/blob/1d3499a18d526b4b1aef0bdbbc54ac812b8ae0c0/pulsar/consumer_partition.go#L196-L208

In the broker side, I checked the relative code and doesn't find any need to reset it. The Java Client does not reset it too.

So I think it's no need to reset the available permits. I don't really understand why should reset it in the legacy code and I guess it's just a mearsure taken for safe.

pulsar/consumer_partition.go
```diff
@@ -138,15 +138,14 @@ type partitionConsumer struct {
 	// the size of the queue channel for buffering messages
 	queueSize       int32
 	queueCh         chan []*message
-	startMessageID  trackingMessageID
+	startMessageID  atomicMessageID
```
nodece, line 141: Could `atomic.Value{}` instead of this?

Gleiphir2769: > Could `atomic.Value{}` instead of this?

I think atomicMessageID is better than `atomic.Value{}` because it's more simple and clearly.

For example, if we use `atomic.Value` as the startMessageID type, the original using of `startMessageID` will be like this.
``` golang
// original
return pc.startMessageID.greater(msgID.messageID)

// atomicMessageID
return pc.startMessageID.get().greater(msgID.messageID)

// atomic.Value{}
return pc.startMessageID.Load().(trackingMessageID).greater(msgID.messageID)
```
`atomic.Value{}` will need one more time type assertion.

But `atomicMessageID` may need a better name and declear position. Do you have more idea? Thanks.

nodece: LGTM

pulsar/consumer_test.go
```diff
@@ -3177,6 +3177,7 @@ func TestConsumerSeekByTimeOnPartitionedTopic(t *testing.T) {
 	// should be able to consume all messages once again
 	for i := 0; i < N; i++ {
 		msg, err := consumer.Receive(ctx)
+		fmt.Println(string(msg.Payload()) + "-" + strconv.Itoa(i))
```
nodece, line 3180:
```suggestion
```

nodece (approved):
LGTM

pulsar/consumer_test.go
Gleiphir2769: Got it. Thanks!

RobertIndie (approved):

PR #930, opened 2022-12-30
Handle nil value message correctly in table-view

Demogorgon314:
### Motivation

Currently, when we publish a nil value message([tombstones](https://en.wikipedia.org/wiki/Tombstone_(data_store))) to a compacted topic used to `TableView`, the `TableView` will store a nil value in the map, but we should remove this key since this is a tombstone.

### Modifications

* Handle nil value message correctly in table-view
* Add unit test to verify.

### DIFF
--- a/pulsar/table_view_test.go
+++ b/pulsar/table_view_test.go
@@ -78,3 +78,67 @@ func TestTableView(t *testing.T) {
 		assert.Equal(t, valuePrefix+k, *(v.(*string)))
 	}
 }
+
+func TestPublishNilValue(t *testing.T) {
+	client, err := NewClient(ClientOptions{
+		URL: lookupURL,
+	})
+
+	assert.NoError(t, err)
+	defer client.Close()
+
+	topic := newTopicName()
+	schema := NewStringSchema(nil)
+
+	// create producer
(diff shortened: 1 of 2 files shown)


pulsar/table_view_test.go
```diff
@@ -78,3 +78,67 @@ func TestTableView(t *testing.T) {
...
+
+	_, err = producer.Send(context.Background(), &ProducerMessage{
+		Key:   "key-1",
+		Value: "value-1",
+	})
+	assert.NoError(t, err)
+
+	for tv.Size() < 1 {
+		time.Sleep(time.Second * 1)
+	}
```
BewareMyPower, line 119: Is there any way to avoid the sleep call like the `awaitility` in Java? If not, we can write a util function.

tisonkun: Since we have `stretchr/testify` in deps, it's possible to try `require.Eventually` as [I did in Kvrocks](https://github.com/apache/incubator-kvrocks/blob/52b0a6102d83d006a0e1dfc55f9d054036cd6df4/tests/gocase/util/client.go#L44-L57).

nodece (approved):

BewareMyPower (approved):

PR #939, opened 2023-01-04
[feature] Max Retry per msg feature added

### LINKED ISSUE #890: How can we set retry logic per msg
Is it possible to set the retry logic per msg ?

We want to retry message to different differnt time. like few message should retry only x time and other msg retry at y time 

retry count per message how we can set at the time of producing/sending to pulsar

ngoyal16:

Fixes #890 

### Motivation

Add a feature to allow Max Retry per msg.

### Modifications

Msg will have max reconsume times property. if not present use default property from consumer max deliveries

### Verifying this change

- *Added Integration tests for per msg retry login*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -561,7 +561,14 @@ func (c *consumer) ReconsumeLaterWithCustomProperties(msg Message, customPropert
 			msgID:      msgID,
 		},
 	}
-	if uint32(reconsumeTimes) > c.dlq.policy.MaxDeliveries {
+
+	maxDeliveries := c.dlq.policy.MaxDeliveries
+	if s, ok := props[MsgPropertyMaxReconsumeTimes]; ok {
+		md, _ := strconv.Atoi(s)
+		maxDeliveries = uint32(md)
+	}
+
+	if uint32(reconsumeTimes) > maxDeliveries {
 		c.dlq.Chan() <- consumerMsg
 	} else {
 		c.rlq.Chan() <- RetryMessage{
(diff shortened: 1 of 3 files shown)


nodece:
It looks like you want to use the Properties to override the MaxDeliveries of dlq.

ngoyal16:
> It looks like you want to use the Properties to override the MaxDeliveries of dlq.

This is the simplest and best way I found for it.. because retry and dlq is managed at client side in all libs so we can use the same...

nodece:
> > It looks like you want to use the Properties to override the MaxDeliveries of dlq.
> 
> This is the simplest and best way I found for it.. because retry and dlq is managed at client side in all libs so we can use the same...

I see. I suggest you make a discussion to mailing lists, see https://pulsar.apache.org/community/#section-discussions.

wolfstudy (changes requested):
Thanks @ngoyal16 work for this. Sorry for request change.

Currently, we prefer users to use the nack backoff Policy to customize their own retry strategy. Currently NackBackoff is an interface that provides a default backoff retry strategy. If the strategy does not meet the needs, we can try to implement the NackBackoff interface to meet the current needs.

The current implementation of ReconsumeLater relies on the function of delaying messages, and the Go SDK still does not support the backoff retry function. Of course, it is not planned to continue to support it in the future. The following problems are found in actual production use:

1. Too many delayed messages will increase the load on Broker
2. NackBackoff can completely replace the usage logic of ReconsumeLater

If there is something unsatisfactory, we prefer to support more customized features in NackBackoff

nodece:
Close this PR, this idea has not been approved by PMC.

jcass8695:
@nodece what is PMC, and why was it not approved? I created an issue today (#1167) which I think this PR would have addressed.

RobertIndie:
> @nodece what is PMC, and why was it not approved? I created an issue today (#1167) which I think this PR would have addressed.

The discussion happened here: https://lists.apache.org/thread/b9rfv6t307z439xx3zt2ym4p140qzp06
For the term of PMC, please see: https://pulsar.apache.org/community/#section-governance

PR #942, opened 2023-01-14
Migrate from the deprecated io/ioutil package

reugn:
### Motivation
The `io/ioutil` package has been deprecated in go1.16, see the [release notes](https://go.dev/doc/go1.16#ioutil). This PR replaces the existing `io/ioutil` calls with their new definitions in the `io` and `os` packages.

### Verifying this change

### Documentation

### DIFF
--- a/.github/workflows/project.yml
+++ b/.github/workflows/project.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.15, 1.16, 1.17, 1.18, 1.19]
+        go-version: [1.16, 1.17, 1.18, 1.19]
     steps:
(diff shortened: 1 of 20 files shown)


nodece (approved):
For me, we should always use the supported version from the golang team.

pgier (approved):
LGTM, although I wonder if we should cut one more release before breaking compatibility with golang 1.15

nodece:
We can use a build flag to be compatible with the different versions.

pgier:
@nodece That's certainly possible, but in my opinion not worth the extra maintenance effort.  It would mean that any bug fixes and new features would also need to be tested against the older versions.  At some point we should just tell users to upgrade their Go version, especially since golang itself only maintains the two most recent versions.

Issue #931, opened 2022-12-30
[feature request] Support producer access mode.


shibd:
### Motivation

The current java client supports four access modes: `Shared`, `Exclusive`, `ExclusiveWithFencing`, and `WaitForExclusive`.

The issue is responsible for tracking, and specific tasks can be disassembled to do.

### Solution

Refer PIP:
1. [PIP 68: Exclusive Producer](https://github.com/apache/pulsar/wiki/PIP-68%3A-Exclusive-Producer)
2. [PIP161: ExclusiveWithFencing](https://github.com/apache/pulsar/issues/15528)

shibd:
Related: #896

avrahams:
Same issue here 
I would like to implement a leader election scheme with the go client

shibd:
I will work on it.

PR #944, opened 2023-01-17
[feat] Support Exclusive Producer access mode.

shibd:
Master Issue: #931

### Motivation

#931

### Modifications

- Support the `Exclusive` Producer access mode config in the `ProducerOptions`.

**Note**: `WaitForExclusive` and `ExclusiveWithFencing` will be supported in other PR.

### Verifying this change

- Add `TestExclusiveProducer` to cover it.

### Documentation

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -95,6 +95,7 @@ type partitionProducer struct {
 	metrics          *internal.LeveledMetrics
 	epoch            uint64
 	schemaCache      *schemaCache
+	topicEpoch       int64
 }
 
 type schemaCache struct {
@@ -161,6 +162,7 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		metrics:          metrics,
 		epoch:            0,
 		schemaCache:      newSchemaCache(),
+		topicEpoch:       -1,
 	}
 	if p.options.DisableBatching {
 		p.batchFlushTicker.Stop()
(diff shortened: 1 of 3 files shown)


pulsar/producer_partition.go
```diff
@@ -237,6 +239,11 @@ func (p *partitionProducer) grabCnx() error {
 		Schema:                   pbSchema,
 		Epoch:                    proto.Uint64(atomic.LoadUint64(&p.epoch)),
 		UserProvidedProducerName: proto.Bool(p.userProvidedProducerName),
+		ProducerAccessMode:       toProtoProducerAccessMode(p.options.ProducerAccessMode).Enum(),
+	}
+
+	if p.topicEpoch > -1 {
```
BewareMyPower, line 245: The default value should be 0, not -1.

nit: You should not use `> n` on a unsigned integer, even if you converted it to a signed integer. For example, 0xFFFFFFFFFFFFFFFFFFFF should be a valid `uint64` epoch (though in Java it's overflowed), but when it's converted to `int64`, it will be -1 and treated as negative. I think you'd better use `uint64` and avoid type cast.

https://github.com/apache/pulsar-client-go/blob/d9b18d0690c15d5da94c9f324d9c2618e3d60bfd/pulsar/internal/pulsar_proto/PulsarApi.pb.go#L3518-L3522

BewareMyPower (changes requested):

pulsar/producer_partition.go
shibd: > The default value should be 0, not -1.

Because Go natively does not support `optional`. We need to initialize it to `-1` to judge whether have received `topicEpoch` from the broker(The broker is incremented from 0). 

> nit: You should not use > n on a unsigned integer, even if you converted it to a signed integer. For example, 0xFFFFFFFFFFFFFFFFFFFF should be a valid uint64 epoch (though in Java it's overflowed), but when it's converted to int64, it will be -1 and treated as negative. I think you'd better use uint64 and avoid type cast.

I use uint64 ptr instead of it.

shibd:
@nodece @freeznet Can you help look at ci failed? https://github.com/apache/pulsar-client-go/actions/runs/4101318705/jobs/7072982772

It doesn't seem to have anything to do with this PR.

nodece:
> @nodece @freeznet Can you help look at ci failed? https://github.com/apache/pulsar-client-go/actions/runs/4101318705/jobs/7072982772
> 
> It doesn't seem to have anything to do with this PR.

It looks like a `golang-lint@1.51.x` issue,  which was released yesterday. Could you fix the `golang-lint` version to `1.50.1`?

BewareMyPower (approved):

Issue #932, opened 2023-01-01
[feature request] Support Transaction API 


liangyepianzhou:
### Motivation
Now, the pulsar-client-go does not support the go-client, so I want to support transaction API for the go client.  
### Solution
Add Transaction API for go client.

geniusjoe:
Since merged in [953](https://github.com/apache/pulsar-client-go/pull/953), can we close this issue as completed? @RobertIndie

crossoverJie:
@geniusjoe Done.

PR #953, opened 2023-02-13
[feat][txn]Implement transactionCoordinatorClient

liangyepianzhou:
Master Issue:https://github.com/apache/pulsar-client-go/issues/932
### Motivation
Implement transaction coordinator client.
### Modifications
1. Implement transaction coordinator
2. implement transaction API
3. Add metric and test
### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation


pulsar/transaction_test.go
```diff
@@ -0,0 +1,100 @@
...
+	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
+	"github.com/stretchr/testify/assert"
+
+	"testing"
+	"time"
+)
+
+func TestTCClient(t *testing.T) {
+	//1. Prepare: create PulsarClient and init transaction coordinator client.
+	topic := "my-topic"
```
shibd, line 30:
```suggestion
	topic := newTopicName()
```

pulsar/transaction_test.go
```diff
@@ -0,0 +1,100 @@
...
+	defer tc.close()
+	defer client.Close()
+}
+
+/*
+Create a transaction coordinator client to send request
+*/
+func createTcClient(t *testing.T) (*transactionCoordinatorClient, *client) {
+	c, err := NewClient(ClientOptions{
+		URL:                   webServiceURLTLS,
```
shibd, line 87: Does Transaction have to have TLS enabled?

Makefile
```diff
@@ -18,7 +18,7 @@
 #
 
 IMAGE_NAME = pulsar-client-go-test:latest
-PULSAR_VERSION ?= 2.8.3
+PULSAR_VERSION ?= 2.10.3
```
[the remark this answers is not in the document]
liangyepianzhou: CI failed again after I changed it to `latest`, so I can only change it to 2.10.3. The specific reason is not clear.

pulsar/transaction_test.go
liangyepianzhou: The transaction needs to get a connection from "pulsar/system/transaction_coordinator_assign".
And the client only has the authority of the `public` tenant in default.
So we need to add the authentication here.

pulsar/transactionCoordinatorClient.go
```diff
@@ -0,0 +1,223 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
nodece, line 1: `pulsar/transactionCoordinatorClient.go` rename to `pulsar/transaction_coordinator_client.go`

pulsar/transactionCoordinatorClient.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, con := range tc.cons {
+		con.Close()
+	}
+}
+
+/*
+*
+New a transactionImpl which can be used to guarantee exactly-once semantics.
+*/
+func (tc *transactionCoordinatorClient) newTransaction(timeout time.Duration) (TxnID, error) {
```
nodece, line 115: It is better `*TxnID, error`

Makefile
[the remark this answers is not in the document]
shibd: > CI failed again after I changed it to latest, so I can only change it to 2.10.3. The specific reason is not clear.
> Fixed version is right.

We can fix it in other PRs. But, Do we need the Go client to be compatible with the latest major version (the current is 2.11.0)?

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/*
+*
+New a transactionImpl which can be used to guarantee exactly-once semantics.
+*/
+func (tc *transactionCoordinatorClient) newTransaction(timeout time.Duration) (*TxnID, error) {
+	err := tc.canSendRequest()
+	if err != nil {
+		return &TxnID{}, err
```
nodece, line 118:
```suggestion
		return nil, err
```

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,223 @@
...
+	nextTcID := tc.nextTCNumber()
+	cmdNewTxn := &pb.CommandNewTxn{
+		RequestId:     proto.Uint64(requestID),
+		TcId:          proto.Uint64(nextTcID),
+		TxnTtlSeconds: proto.Uint64(uint64(timeout.Milliseconds())),
+	}
+
+	cnx, err := tc.client.rpcClient.RequestOnCnx(tc.cons[nextTcID], requestID, pb.BaseCommand_NEW_TXN, cmdNewTxn)
+	if err != nil {
+		return &TxnID{}, err
```
nodece, line 130:
```suggestion
		return nil, err
```

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,223 @@
...
+	_, err = tc.client.rpcClient.RequestOnCnx(tc.cons[id.mostSigBits], requestID,
+		pb.BaseCommand_ADD_SUBSCRIPTION_TO_TXN, cmdAddSubscription)
+	return err
+}
+
+/*
+*
+Commit or abort the transactionImpl.
+*/
+func (tc *transactionCoordinatorClient) endTxn(id *TxnID, action pb.TxnAction) error {
```
nodece, line 188:
```suggestion
func (tc *transactionCoordinatorClient) endTxn(id TxnID, action pb.TxnAction) error {
```

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,223 @@
...
+	_, err = tc.client.rpcClient.RequestOnCnx(tc.cons[id.mostSigBits], requestID,
+		pb.BaseCommand_ADD_PARTITION_TO_TXN, cmdAddPartitions)
+	return err
+}
+
+/*
+Register the subscription which acked messages with the transactionImpl.
+And this can be used when ending the transactionImpl.
+*/
+func (tc *transactionCoordinatorClient) addSubscriptionToTxn(id *TxnID, topic string, subscription string) error {
```
nodece, line 163:
```suggestion
func (tc *transactionCoordinatorClient) addSubscriptionToTxn(id TxnID, topic string, subscription string) error {
```

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,223 @@
...
+	return &TxnID{*cnx.Response.NewTxnResponse.TxnidMostBits,
+		*cnx.Response.NewTxnResponse.TxnidLeastBits}, nil
+}
+
+/*
+*
+Register the partitions which published messages with the transactionImpl.
+And this can be used when ending the transactionImpl.
+*/
+func (tc *transactionCoordinatorClient) addPublishPartitionToTxn(id *TxnID, partitions []string) error {
```
nodece, line 142:
```suggestion
func (tc *transactionCoordinatorClient) addPublishPartitionToTxn(id TxnID, partitions []string) error {
```

pulsar/helper_for_test.go
```diff
@@ -157,6 +157,13 @@ func topicStats(topic string) (map[string]interface{}, error) {
 	return metadata, err
 }
 
+func transactionStats(id *TxnID) (map[string]interface{}, error) {
```
nodece, line 160:
```suggestion
func transactionStats(id TxnID) (map[string]interface{}, error) {
```

pulsar/transaction_coordinator_client.go
liangyepianzhou: Why should we use TxnID instead of * TxnID here? I understand that using `*TxnID` can reduce the memory copy during value transfer.

pulsar/transaction_coordinator_client.go
liangyepianzhou: Why should we use TxnID instead of * TxnID here? I understand that using *TxnID can reduce the memory copy during value transfer.

nodece: Sounds good. But when using the `*TxnID`, I always want to check if it is `nil`, so I suggest you use `TxnID`, this mean is not `nil`.

liangyepianzhou: This is the logic of internal calls, not the interface exposed to users. So it is not necessary to determine whether it is `nil` here.

nodece: Ok.

nodece (approved):
LGTM

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,223 @@
...
+
+	return &TxnID{*cnx.Response.NewTxnResponse.TxnidMostBits,
+		*cnx.Response.NewTxnResponse.TxnidLeastBits}, nil
+}
+
+/*
+*
+Register the partitions which published messages with the transactionImpl.
+And this can be used when ending the transactionImpl.
+*/
```
BewareMyPower, line 141: Please make the comment style consistent. Use 

```go
// line 1...
// line 2...
```

or

```go
/**
 * line1
 * line2
 * ...
 */
```

BewareMyPower (changes requested):

integration-tests/conf/standalone.conf
```diff
@@ -111,8 +110,10 @@ superUserRoles=localhost,superUser,admin
 
 # Authentication settings of the broker itself. Used when the broker connects to other brokers,
 # either in same or other clusters
-brokerClientAuthenticationPlugin=
-brokerClientAuthenticationParameters=
+brokerClientTlsEnabled=true
+brokerClientTrustCertsFilePath=/pulsar/certs/cacert.pem
+brokerClientAuthenticationPlugin=org.apache.pulsar.client.impl.auth.AuthenticationTls
+brokerClientAuthenticationParameters={"tlsCertFile":"/pulsar/certs/client-cert.pem","tlsKeyFile":"/pulsar/certs/client-key.pem"}
```
BewareMyPower, line 116: Why did you add the TLS authentication? Does transaction depend on the TLS?

liangyepianzhou: Transaction coordinators use `brokerClient` to send commands to TB and TP.
So we need to add  TLS authentication to `brokerClient` here.

pulsar/client.go
```diff
@@ -142,6 +142,8 @@ type ClientOptions struct {
 	// Specify metric registerer used to register metrics.
 	// Default prometheus.DefaultRegisterer
 	MetricsRegisterer prometheus.Registerer
+
+	IsEnableTransaction bool
```
BewareMyPower, line 146: Could you remove the `Is` prefix? See other options that enable some feature:
- producer.go: EnableChunking
- consumer.go: EnableBatchIndexAcknowledgment, RetryEnable, EnableDefaultNackBackoffPolicy

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/apache/pulsar-client-go/pulsar/log"
+	"google.golang.org/protobuf/proto"
+)
+
+type transactionCoordinatorClient struct {
+	client                    *client
+	cons                      []internal.Connection
+	epoch                     uint64
+	semaphore                 internal.Semaphore
+	blockIfReachMaxPendingOps bool
```
BewareMyPower, line 37: Could this field be false? I see `transactionCoordinatorClient` should only be created by `newTransactionCoordinatorClientImpl`, which set it true. If it's always true, please remove this field.

pulsar/transaction_test.go
```diff
@@ -0,0 +1,100 @@
...
+		assert.Equal(t, err.Error(), "http error status code: 404")
+	}
+	defer consumer.Close()
+	defer tc.close()
+	defer client.Close()
+}
+
+/*
+Create a transaction coordinator client to send request
+*/
```
BewareMyPower, line 84:
```suggestion
// Create a transaction coordinator client to send request
```

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,210 @@
...
+	//The number of transactionImpl coordinators
+	tcNum uint64
```
BewareMyPower, line 39: This field is redundant. We can replace the use of it with `len(cons)`.

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,210 @@
...
+		con.Close()
+	}
+}
+
+// newTransaction new a transactionImpl which can be used to guarantee exactly-once semantics.
+func (tc *transactionCoordinatorClient) newTransaction(timeout time.Duration) (*TxnID, error) {
+	err := tc.canSendRequest()
+	if err != nil {
+		return nil, err
+	}
```
BewareMyPower, line 114:
```suggestion
        if err := tc.canSendRequest(); err != nil {
                return nil, err
        }
```

Simplify the code

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,210 @@
...
+		*cnx.Response.NewTxnResponse.TxnidLeastBits}, nil
+}
+
+// addPublishPartitionToTxn register the partitions which published messages with the transactionImpl.
+// And this can be used when ending the transactionImpl.
+func (tc *transactionCoordinatorClient) addPublishPartitionToTxn(id *TxnID, partitions []string) error {
+	err := tc.canSendRequest()
+	if err != nil {
+		return err
+	}
```
BewareMyPower, line 138:
```suggestion
        if err := tc.canSendRequest(); err != nil {
                return nil, err
        }
```

Simplify the code. The following `_, err = ...` in line 146 should become `_, err := ...` after this change.

pulsar/transaction_coordinator_client.go
```diff
@@ -0,0 +1,210 @@
...
+	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
+	"github.com/apache/pulsar-client-go/pulsar/log"
+	"google.golang.org/protobuf/proto"
+)
+
+type transactionCoordinatorClient struct {
+	client                    *client
+	cons                      []internal.Connection
+	epoch                     uint64
+	semaphore                 internal.Semaphore
```
BewareMyPower, line 36: It's never released. You should release the semaphore for each response received

pulsar/client_impl_test.go
```diff
@@ -425,7 +426,7 @@ func TestNamespaceTopics(t *testing.T) {
 		t.Fatal(err)
 	}
 	topic2 := fmt.Sprintf("%s/topic-2", namespace)
-	if err := httpPut("admin/v2/persistent/"+topic2, namespace); err != nil {
```
BewareMyPower, line 428: Could you explain changes in this test?

BewareMyPower (changes requested):

pulsar/client_impl_test.go
liangyepianzhou: This request got a 400 response after enabling TLS.
I confirmed to @freeznet  it is an error code.
So I change its logic to be like the logic of topic 1.

BewareMyPower (approved):

Issue #927, opened 2022-12-30
[feature request] Support client memory limit


shibd:
### Motivation

PIP: https://github.com/apache/pulsar/wiki/PIP-74%3A-Pulsar-client-memory-limits

### Alternatives

- Refer to the impl of the producer for the Java client: https://github.com/apache/pulsar/pull/8965
- Refer to the impl of the consumer for the Java client: (Consumer need Support auto scaled consumer receiver queue first)
   - https://github.com/apache/pulsar/pull/14494
   - https://github.com/apache/pulsar/pull/14400

shibd:
> @RobertIndie I'm interested in this feature. Could you assigned it to me? I will take some time to implement it.
Thanks!

Hi, @Gleiphir2769  Great! #955 implemented memory control on the production side. You can look at how to implement the consumer side

Gleiphir2769:
Hi @shibd. All the features have been finished. I think we can close this issue.

PR #955, opened 2023-02-15
[feat] Support memory limit for producer.

shibd:
Master Issue: #927 

### Motivation

#927 

### Modifications

- Added `channel_cond` class to enhance `sync.cond` to support wait that can accept context.
- Added the `memory_limit_controller` class.
- Producer adds memory limit relate logic.

### Verifying this change
- Added `channel_cond_test` to cover `channel_cond`
- Added `memory_limit_controller_test` to cover `mem_mory_limit_controller`
- Added `TestMemLimitRejectProducerMessages` and `TestMemLimitContextCancel ` to cover producer relate logic.

  - The public API: (yes)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/internal/channel_cond.go
+++ b/pulsar/internal/channel_cond.go
@@ -28,13 +28,13 @@ type chCond struct {
 	L sync.Locker
 	// The pointer to the channel, the channel pointed to may change,
 	// because we will use the channel's close mechanism to implement broadcast notifications.
-	notifyChanPointer unsafe.Pointer
+	notifyChPtr unsafe.Pointer
 }
 
 func newCond(l sync.Locker) *chCond {
 	c := &chCond{L: l}
 	n := make(chan struct{})
-	c.notifyChanPointer = unsafe.Pointer(&n)
+	c.notifyChPtr = unsafe.Pointer(&n)
 	return c
 }
 
@@ -65,12 +65,12 @@ func (c *chCond) waitWithContext(ctx context.Context) bool {
 // It is not required for the caller to hold c.L during the call.
 func (c *chCond) broadcast() {
 	n := make(chan struct{})
-	ptrOld := atomic.SwapPointer(&c.notifyChanPointer, unsafe.Pointer(&n))
+	ptrOld := atomic.SwapPointer(&c.notifyChPtr, unsafe.Pointer(&n))
 	// close old channels to trigger broadcast.
 	close(*(*chan struct{})(ptrOld))
 }
 
 func (c *chCond) notifyChan() <-chan struct{} {
-	ptr := atomic.LoadPointer(&c.notifyChanPointer)
+	ptr := atomic.LoadPointer(&c.notifyChPtr)
 	return *((*chan struct{})(ptr))
 }
--- a/pulsar/internal/channel_cond.go
+++ b/pulsar/internal/channel_cond.go
@@ -1,0 +1,73 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package internal
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"unsafe"
+)
+
+type ChCond struct {
+	L sync.Locker
+	n unsafe.Pointer
+}
+
+func NewCond(l sync.Locker) *ChCond {
+	c := &ChCond{L: l}
+	n := make(chan struct{})
+	c.n = unsafe.Pointer(&n)
+	return c
+}
+
+// Wait for Broadcast calls. Similar to regular sync.Cond
+func (c *ChCond) Wait() {
+	n := c.notifyChan()
+	c.L.Unlock()
+	<-n
+	c.L.Lock()
+}
+
+// WaitWithContext Same as Wait() call, but the end condition can also be controlled through the context.
+func (c *ChCond) WaitWithContext(ctx context.Context) bool {
+	n := c.notifyChan()
+	c.L.Unlock()
+	defer c.L.Lock()
+	select {
+	case <-n:
+		return true
+	case <-ctx.Done():
+		return false
+	default:
+		return true
+	}
+}
+
+// Broadcast wakes all goroutines waiting on c.
+// It is not required for the caller to hold c.L during the call.
+func (c *ChCond) Broadcast() {
+	n := make(chan struct{})
+	ptrOld := atomic.SwapPointer(&c.n, unsafe.Pointer(&n))
+	close(*(*chan struct{})(ptrOld))
+}
+
+func (c *ChCond) notifyChan() <-chan struct{} {
+	ptr := atomic.LoadPointer(&c.n)
+	return *((*chan struct{})(ptr))
+}
--- a/pulsar/internal/channel_cond.go
+++ b/pulsar/internal/channel_cond.go
@@ -24,28 +24,28 @@ import (
 	"unsafe"
 )
 
-type ChCond struct {
+type chCond struct {
 	L sync.Locker
 	n unsafe.Pointer
 }
 
-func NewCond(l sync.Locker) *ChCond {
-	c := &ChCond{L: l}
+func newCond(l sync.Locker) *chCond {
+	c := &chCond{L: l}
 	n := make(chan struct{})
 	c.n = unsafe.Pointer(&n)
 	return c
 }
 
-// Wait for Broadcast calls. Similar to regular sync.Cond
-func (c *ChCond) Wait() {
+// wait for broadcast calls. Similar to regular sync.Cond
+func (c *chCond) wait() {
 	n := c.notifyChan()
 	c.L.Unlock()
 	<-n
 	c.L.Lock()
 }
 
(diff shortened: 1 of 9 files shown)


pulsar/internal/memory_limit_controller.go
```diff
@@ -0,0 +1,101 @@
...
+type memoryLimitController struct {
+	limit        int64
+	chCond       ChCond
+	currentUsage int64
+}
+
+func NewMemoryLimitController(limit int64) MemoryLimitController {
+	mlc := &memoryLimitController{
+		limit:  limit,
+		chCond: *NewCond(&sync.Mutex{}),
```
RobertIndie, line 45: This seems will result in value copied. Better to use pointer type here.

pulsar/internal/channel_cond_test.go
```diff
@@ -0,0 +1,55 @@
...
+	wg.Wait()
+}
+
+func TestChCondWithContext(t *testing.T) {
+	cond := NewCond(&sync.Mutex{})
+	wg := sync.WaitGroup{}
+	ctx, cancel := context.WithCancel(context.Background())
+	wg.Add(1)
+	go func() {
+		cond.L.Lock()
```
RobertIndie, line 48: It's better to add `Lock` and `Unlock` method to `ChCond`.

pulsar/internal/memory_limit_controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	mlc := NewMemoryLimitController(100)
+
+	for i := 0; i < 101; i++ {
+		assert.True(t, mlc.TryReserveMemory(1))
+	}
+
+	assert.False(t, mlc.TryReserveMemory(1))
+	assert.Equal(t, int64(101), mlc.CurrentUsage())
+	assert.Equal(t, 1.01, mlc.CurrentUsagePercent())
```
RobertIndie, line 39: Can we check equal for float type here? Looks like `assert.Equal` doesn't due with the float case.

pulsar/producer_partition.go
```diff
@@ -494,14 +496,15 @@ func (p *partitionProducer) internalSend(request *sendRequest) {
...
 		return
 	}
 
 	if p.options.DisableMultiSchema {
 		if msg.Schema != nil && p.options.Schema != nil &&
 			msg.Schema.GetSchemaInfo().hash() != p.options.Schema.GetSchemaInfo().hash() {
 			p.publishSemaphore.Release()
+			p.client.memLimit.ReleaseMemory(uncompressedPayloadSize)
 			request.callback(nil, request.msg, fmt.Errorf("msg schema can not match with producer schema"))
 			p.log.WithError(err).Errorf("The producer %s of the topic %s is disabled the `MultiSchema`", p.producerName, p.topic)
```
RobertIndie, line 509: How about we combine these lines into one function like `completeCallbackAndReleaseSemaphore` in java? This will help subsequent developers not miss releasing any resources when adding new logic.

pulsar/producer_test.go
```diff
@@ -1779,3 +1779,126 @@ func TestWaitForExclusiveProducer(t *testing.T) {
...
+
+	_, err = producer2.Send(context.Background(), &ProducerMessage{
+		Payload: make([]byte, 1024),
+	})
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, getResultStr(ClientMemoryBufferIsFull))
+
+	// flush pending msg
+	producer1.Flush()
+	producer2.Flush()
```
RobertIndie, line 1840: lint warning here: `Unhandled error `
Same for line 1897

pulsar/internal/channel_cond_test.go
shibd: Here keep native Cond usage habits

shibd: The native usage should look like this:
```go
	cond := sync.NewCond(&sync.Mutex{})
	cond.L.Lock()
	cond.L.Unlock()
```

pulsar/internal/channel_cond.go
```diff
@@ -0,0 +1,73 @@
...
+package internal
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"unsafe"
+)
+
+type ChCond struct {
```
BewareMyPower, line 27: We should not expose this struct and its methods to users if `ChCond` is only used in the `internal` package.

BewareMyPower (changes requested):

BewareMyPower (commented):
And I don't get why do we have to wrap a conditional variable? We can achieve the same goal with the channel itself, or just reuse the `Semaphore` [here](https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/semaphore.go)

shibd:
> And I don't get why do we have to wrap a conditional variable? 

The native cond does not support timeout wait, and mainly wants to control the timeout or cancel the wait through context.

https://github.com/apache/pulsar-client-go/pull/955/files#diff-d4362cb0b0046577b68820bba2bde37b97ca5d7db7237e4e4b5eac54d7aa8bc3R89-R120

> We can achieve the same goal with the channel itself,

Do you want `mem_controller_limit` all the methods into the channel first, and then select to read the signal and process it? Similar to #957 `ackGroupingTracker `? 

> or just reuse the Semaphore [here](https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/semaphore.go)

This semaphore can only process applications or release 1 signal.

BewareMyPower:
For example, if we add a `ch chan bool` field to `memoryLimitController`, we can implement blocking `ReserveMemory` method like:

```go
func (m *memoryLimitController) ReserveMemory(ctx context.Context, size int64) bool {
        currentUsage := atomic.AddInt64(&m.currentUsage, size)
        for currentUsage > m.limit {
                select {
                case <-m.ch:
                        currentUsage = atomic.LoadInt64(&m.currentUsage)
                case <-ctx.Done():
                        return false
                }
        }
        return true
}

func (m *memoryLimitController) ReleaseMemory(size int64) {
        newUsage := atomic.AddInt64(&m.currentUsage, -size)
        // newUsage+size > m.limit means m was blocked in ReserveMemory method
        if newUsage+size > m.limit && newUsage <= m.limit {
                m.ch <- true
        }
}
```

The code above is not verified yet. But with the channel the code looks more simple and clear.

pulsar/internal/channel_cond.go
shibd: This implementation is under the `internal` package, which is not exposed to the user, right?

shibd:
> > This semaphore can only support acquire or release 1 signal.
> 
> Yes. But we can adopt a similar way to handle this case.
> 
> For example, if we add a `ch chan bool` field to `memoryLimitController`, we can implement blocking `ReserveMemory` method like:
> 
> ```go
> func (m *memoryLimitController) ReserveMemory(ctx context.Context, size int64) bool {
>         // NOTE: maybe we need to check if m.currentUsage > m.limit first
>         currentUsage := atomic.AddInt64(&m.currentUsage, size)
>         for currentUsage > m.limit {
>                 select {
>                 case <-m.ch:
>                         currentUsage = atomic.LoadInt64(&m.currentUsage)
>                 case <-ctx.Done(): // NOTE: Not sure if we need to reset some fields here
>                         return false
>                 }
>         }
>         return true
> }
> 
> func (m *memoryLimitController) ReleaseMemory(size int64) {
>         newUsage := atomic.AddInt64(&m.currentUsage, -size)
>         // newUsage+size > m.limit means m was blocked in ReserveMemory method
>         if newUsage+size > m.limit && newUsage <= m.limit {
>                 m.ch <- true
>         }
> }
> ```
> 
> The code above is not verified yet. But with the channel the code looks more simple and clear.

This implementation has a problem, and the `broadcasting` cannot be implemented using channels.

For example:
1. `currentUsage` = 100, `limit` = 100
2. goroutine `1` call `ReseveMemory(10)`, it will blocked.
3. goroutine `2` call `ReseveMemory(10)`, it will blocked.
4. goroutine `3` call `ReleaseMemory(20)`, it only wake one goroutine(1 or 2). The expectation is that both are woken up and return true.


// NOTE: maybe we need to check if m.currentUsage > m.limit first

Like this note. If we want to handle this case, we may need to introduce a variable `waitNum`. So I feel like it's complicated to implement `mem_controller_limit`, and it's better to introduce `channel_cond`.

BewareMyPower:
> If we want to handle this case, we may need to introduce a variable waitNum

I see. Currently let's go head with the conditional variable, but don't expose this struct to users.

pulsar/internal/channel_cond.go
BewareMyPower: Oh I see.

BewareMyPower (approved):

pulsar/internal/channel_cond_test.go
RobertIndie: Make sense to me.

RobertIndie:
> Like this note. If we want to handle this case, we may need to introduce a variable waitNum. So I feel like it's complicated to implement mem_controller_limit, and it's better to introduce channel_cond.

@shibd For this case, another workaround is to close the channel instead of sending value. But it's actually what this PR does.

RobertIndie (approved):

pulsar/internal/channel_cond.go
```diff
@@ -0,0 +1,73 @@
...
+	L sync.Locker
+	n unsafe.Pointer
```
BewareMyPower, line 29: A question about the name. What does `n` represent? It's better to give a meaningful name. See `sync.Cond`:

```go
type Cond struct {
        noCopy noCopy

        // L is held while observing or changing the condition
        L Locker

        notify  notifyList
        checker copyChecker
}
```

BewareMyPower: I see `n` is a channel in other places like `Broadcast` and `WaitWithContext`. In this case, `n` is a short name of `notifyChan`. It's okay because it's recommended to have short variable names, see https://github.com/golang/go/wiki/CodeReviewComments#variable-names

But the rule is not applied for struct fields, because

>  the further from its declaration that a name is used, the more descriptive the name must be

The field could be used in the method definitions, so it's far from where it's declared (in the struct)

RobertIndie: `notifyChPtr` or `notifyChanPtr` would be better. ([`be short rather than long`](https://github.com/golang/go/wiki/CodeReviewComments#variable-names))

PR #957, opened 2023-02-16
Support grouping ACK requests by time and size

### LINKED ISSUE #949: Consume Performance drops when set EnableBatchIndexAcknowledgment = true
According to PR: #938 
use master version(v0.9.1-0.20230117072740-d9b18d0690c1) to consume messages while EnableBatchIndexAcknowledgment set  
 true，but **consume performance drops to 2/3 of previous**。

The test situation is as follows:
Topic has 5 partitions, producer production rate is 20MB/s , 300000 rows/s.

and consumers consume situations:
| SDK Version  | Enable Batch Index Ack | Consumer Instances | Consume Rate |
| ------------- | ------------- | ------------- | ------------- |
| v0.9.1-0.20230117072740-d9b18d0690c1 | Yes  | 3 | 100000 rows/s  |
| v0.9.1-0.20230117072740-d9b18d0690c1 | No  | 3 | 300000+ rows/s  |

 Analyze the problem by pprof，we found that internal.(*connection).internalSendRequest and pulsar.(*partitionConsumer).internalAck are much more resource intensive when set EnableBatchIndexAcknowledgment as true.

Review the code:
```
func (pc *partitionConsumer) ackID(msgID MessageID, withResponse bool) error {
	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
		pc.log.WithField("state", state).Error("Failed to ack by closing or closed consumer")
		return errors.New("consumer state is closed")
	}

	if cmid, ok := toChunkedMessageID(msgID); ok {
		return pc.unAckChunksTracker.ack(cmid)
	}

	trackingID, ok := toTrackingMessageID(msgID)
	if !ok {
		return errors.New("failed to convert trackingMessageID")
	}

	ackReq := new(ackRequest)
	ackReq.doneCh = make(chan struct{})
	ackReq.ackType = individualAck
	if !trackingID.Undefined() && trackingID.ack() {
		pc.metrics.AcksCounter.Inc()
		pc.metrics.ProcessingTime.Observe(float64(time.Now().UnixNano()-trackingID.receivedTime.UnixNano()) / 1.0e9)
		ackReq.msgID = trackingID
		// send ack request to eventsCh
		pc.eventsCh <- ackReq

		if withResponse {
			<-ackReq.doneCh
		}

		pc.options.interceptors.OnAcknowledge(pc.parentConsumer, msgID)
	} else if pc.options.enableBatchIndexAck {
		ackReq.msgID = trackingID
		pc.eventsCh <- ackReq
	}

	if withResponse {
		return ackReq.err
	}
	return nil
}
```

Maybe problem is that partitionConsumer will send ack request to *Pulsar Server* by every MessageID, without waiting all msg of one batch be acked by ackTracker, it leads to ack requests becoming much more than BatchIndexAck disabled, performance drops bacause of much more processing requests. And backlog is lasting increasing, could not catch up with the production rate.

So, enableBatchIndexAck should follow the previous processing method or there is another way.

@BewareMyPower Thank you for developing this feature. Could you take a look at this problem ? Thanks a lot!

BewareMyPower:
Fixes #949 

### Motivation

Currently the Go client does not support grouping ACK requests, so each time `Ack` (or similar APIs) is called, a ACK request will be sent, which could downgrade the performance. We need to support configuring the time and size to cache `MessageID` before sending ACK requests.

### Modifications
- Add an `AckGroupingOptions` field to `ConsumerOptions`, when it's nil, use 100ms as the max time and 1000 as the max size.
- Add an `ackGroupingTracker` interface to support grouping ACK requests.
- When `AckWithResponse` is false, adding the `MessageID` instance to the tracker instead of sending the requests to `eventsCh`.

### Verifying this change

  - Added `ack_grouping_tracker_test.go` to verify `ackGroupingTracker` in various cases
  - The consumer side change can be covered by existing tests because the default `AckGroupingOptions` config is `{ MaxSize: 1000, MaxTime: 100*time.Millisecond }`.

  - The public API: (**yes**)

### Documentation

### DIFF
--- a/pulsar/ack_grouping_tracker.go
+++ b/pulsar/ack_grouping_tracker.go
@@ -1,0 +1,297 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package pulsar
+
+import (
+	"time"
+
+	"github.com/bits-and-blooms/bitset"
+)
+
+type ackGroupingTracker interface {
+	add(id MessageID)
+
+	addCumulative(id MessageID)
+
+	isDuplicate(id MessageID) bool
+
+	flush()
+
+	flushAndClean()
+
+	close()
+}
+
+type ackFlushType int
+
+const (
+	flushOnly ackFlushType = iota
+	flushAndClean
+	flushAndClose
+)
+
+func newAckGroupingTracker(options *AckGroupingOptions,
+	ackIndividual func(id MessageID),
+	ackCumulative func(id MessageID)) ackGroupingTracker {
+	if options == nil {
+		options = &AckGroupingOptions{
+			MaxSize: 1000,
+			MaxTime: 100 * time.Millisecond,
+		}
+	}
+
+	if options.MaxSize <= 1 {
+		return &immediateAckGroupingTracker{
+			ackIndividual: ackIndividual,
+			ackCumulative: ackCumulative,
+		}
+	}
+
+	c := &cachedAcks{
+		singleAcks:        make([]MessageID, options.MaxSize),
+		pendingAcks:       make(map[int64]*bitset.BitSet),
+		lastCumulativeAck: EarliestMessageID(),
+		ackIndividual:     ackIndividual,
+		ackCumulative:     ackCumulative,
+		ackList: func(ids []MessageID) {
+			// TODO: support ack a list of MessageIDs
+			for _, id := range ids {
+				ackIndividual(id)
+			}
+		},
+	}
+
+	timeout := time.NewTicker(time.Hour)
+	if options.MaxTime > 0 {
+		timeout = time.NewTicker(options.MaxTime)
+	} else {
+		timeout.Stop()
+	}
+	t := &timedAckGroupingTracker{
+		ackIndividualCh:   make(chan MessageID),
+		ackCumulativeCh:   make(chan MessageID),
+		duplicateIDCh:     make(chan MessageID),
+		duplicateResultCh: make(chan bool),
+		flushCh:           make(chan ackFlushType, 10),
+		waitFlushCh:       make(chan bool),
+	}
+	go func() {
+		for {
+			select {
+			case id := <-t.ackIndividualCh:
+				if c.addAndCheckIfFull(id) {
+					c.flushIndividualAcks()
+					if options.MaxTime > 0 {
+						timeout.Reset(options.MaxTime)
+					}
+				}
+			case id := <-t.ackCumulativeCh:
+				c.tryUpdateLastCumulativeAck(id)
+				if options.MaxTime <= 0 {
+					c.flushCumulativeAck()
+				}
+			case id := <-t.duplicateIDCh:
+				t.duplicateResultCh <- c.isDuplicate(id)
+			case <-timeout.C:
+				c.flush()
+			case ackFlushType := <-t.flushCh:
+				timeout.Stop()
+				c.flush()
+				t.waitFlushCh <- true
+				if ackFlushType == flushAndClose {
+					return
+				} else if ackFlushType == flushAndClean {
+					c.clean()
+				}
+			}
+		}
+	}()
+	return t
+}
+
+type immediateAckGroupingTracker struct {
+	ackIndividual func(id MessageID)
+	ackCumulative func(id MessageID)
+}
+
+func (i *immediateAckGroupingTracker) add(id MessageID) {
+	i.ackIndividual(id)
+}
+
+func (i *immediateAckGroupingTracker) addCumulative(id MessageID) {
+	i.ackCumulative(id)
+}
+
+func (i *immediateAckGroupingTracker) isDuplicate(id MessageID) bool {
+	return false
+}
+
+func (i *immediateAckGroupingTracker) flush() {
+}
+
+func (i *immediateAckGroupingTracker) flushAndClean() {
+}
+
+func (i *immediateAckGroupingTracker) close() {
+}
+
+type cachedAcks struct {
+	singleAcks []MessageID
+	index      int
+
+	// Key is the hash code of the ledger id and the netry id,
+	// Value is the bit set that represents which messages are acknowledged if the entry stores a batch.
+	// The bit 1 represents the message has been acknowledged, i.e. the bits "111" represents all messages
+	// in the batch whose batch size is 3 are not acknowledged.
+	// After the 1st message (i.e. batch index is 0) is acknowledged, the bits will become "011".
+	// Value is nil if the entry represents a single message.
+	pendingAcks map[int64]*bitset.BitSet
+
+	lastCumulativeAck     MessageID
+	cumulativeAckRequired bool
+
+	ackIndividual func(id MessageID)
+	ackCumulative func(id MessageID)
+	ackList       func(ids []MessageID)
+}
+
+func (t *cachedAcks) addAndCheckIfFull(id MessageID) bool {
+	t.singleAcks[t.index] = id
+	t.index++
+	key := messageIDHash(id)
+	ackSet, found := t.pendingAcks[key]
+	if !found {
+		if messageIDIsBatch(id) {
+			ackSet = bitset.New(uint(id.BatchSize()))
+			for i := 0; i < int(id.BatchSize()); i++ {
+				ackSet.Set(uint(i))
+			}
+			t.pendingAcks[key] = ackSet
+		} else {
+			t.pendingAcks[key] = nil
+		}
+	}
+	if ackSet != nil {
+		ackSet.Clear(uint(id.BatchIdx()))
+	}
+	return t.index == len(t.singleAcks)
+}
+
+func (t *cachedAcks) tryUpdateLastCumulativeAck(id MessageID) {
+	if messageIDCompare(t.lastCumulativeAck, id) < 0 {
+		t.lastCumulativeAck = id
+		t.cumulativeAckRequired = true
+	}
+}
+
+func (t *cachedAcks) isDuplicate(id MessageID) bool {
+	if messageIDCompare(t.lastCumulativeAck, id) >= 0 {
+		return true
+	}
+	ackSet, found := t.pendingAcks[messageIDHash(id)]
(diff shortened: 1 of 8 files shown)


pulsar/ack_grouping_tracker.go
```diff
@@ -0,0 +1,298 @@
...
+	singleAcks []MessageID
+	index      int
+
+	// Key is the hash code of the ledger id and the netry id,
+	// Value is the bit set that represents which messages are acknowledged if the entry stores a batch.
+	// The bit 1 represents the message has been acknowledged, i.e. the bits "111" represents all messages
+	// in the batch whose batch size is 3 are not acknowledged.
+	// After the 1st message (i.e. batch index is 0) is acknowledged, the bits will become "011".
+	// Value is nil if the entry represents a single message.
+	pendingAcks map[int64]*bitset.BitSet
```
shibd, line 164: Why define `pendingAcks` here? Does it look like it's to judge `isDuplicate`?

pulsar/ack_grouping_tracker.go
```diff
@@ -0,0 +1,298 @@
...
+		timeout = time.NewTicker(options.MaxTime)
+	} else {
+		timeout.Stop()
+	}
+	t := &timedAckGroupingTracker{
+		ackIndividualCh:   make(chan MessageID),
+		ackCumulativeCh:   make(chan MessageID),
+		duplicateIDCh:     make(chan MessageID),
+		duplicateResultCh: make(chan bool),
+		flushCh:           make(chan ackFlushType, 10),
```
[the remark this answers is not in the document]
BewareMyPower: I just set it with a large value in case `flush` could be called in different threads. I will try figuring out a more proper value.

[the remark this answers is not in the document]
BewareMyPower: Since all flush operations are synchronous, I changed it to a unbuffered channel now. Let's see if any test would fail.

pulsar/ack_grouping_tracker.go
shibd: I got it. thanks for your explain.

pulsar/consumer_partition_test.go
```diff
@@ -37,6 +37,8 @@ func TestSingleMessageIDNoAckTracker(t *testing.T) {
 		metrics:              newTestMetrics(),
 		decryptor:            crypto.NewNoopDecryptor(),
 	}
+	pc.ackGroupingTracker = newAckGroupingTracker(&AckGroupingOptions{MaxSize: 0},
```
shibd, line 40: Can you add more unit tests? Such as:
- MaxSize > 0
- A Individual ack in a  batch message.
- Other Similar to: [CPP test](https://github.com/apache/pulsar-client-cpp/blob/06eab69f4caedc1962a21ac52e5c19de868609cd/tests/AcknowledgeTest.cc)

BewareMyPower: This line is to fix the error that `pc.ackGroupingTracker` is `nil` in this test without the explicit initialization of `pc.ackGroupingTracker` because the `partitionConsumer` is created manually, not by the `Consumer`. The tests for different `AckGroupingOptions` are added into `ack_grouping_tracker_test.go`.

> MaxSize > 0

The default config (`MaxSize: 1000, MaxTime: 10ms`) has been verified in all existing tests that call acknowledge related methods.

> A Individual ack in a batch message.

TestBatchIndexAck

This PR has already changed 600+ lines. Adding these tests could make it very huge. What do you think of adding the different options into a single test like `TestBatchIndexAck`?

BewareMyPower: I've added the test cases to `TestBatchIndexAck` and fix the bug that `AckGroupingOptions` is not exposed, PTAL again.

shibd (approved):

panszobe:
Does Community has some Performance Test Data?

I used the latest version(v0.9.1-0.20230222003822-7d257b01b121) to test today, still found that which Test Case by EnableBatchIndexAcknowledgment=true consumed messages much more slower than EnableBatchIndexAcknowledgment=false, and could not catch up with the producer‘s production. 
As consuming from the same topic, consumers of JAVA Client by EnableBatchIndexAcknowledgment=true consumed messages smoothly, and backlog was not enlarged. 

I think that _Support grouping ACK_ feature is OK, but maybe _EnableBatchIndexAcknowledgment_ feature have much room for improvement.

shibd:
> Does Community has some Performance Test Data?
> 
> I used the latest version(v0.9.1-0.20230222003822-7d257b01b121) to test today, still found that which Test Case by EnableBatchIndexAcknowledgment=true consumed messages much more slower than EnableBatchIndexAcknowledgment=false, and could not catch up with the producer‘s production. As consuming from the same topic, consumers of JAVA Client by EnableBatchIndexAcknowledgment=true consumed messages smoothly, and backlog was not enlarged.
> 
> I think that _Support grouping ACK_ feature is OK, but maybe _EnableBatchIndexAcknowledgment_ feature have much room for improvement.

Maybe we should set ackIndividualCh and ackCumulativeCh size > 0.

BewareMyPower:
> Maybe we should set ackIndividualCh and ackCumulativeCh size > 0

@shibd What's the benefit of setting this channel size greater than 0?  I think we don't need to cache these MessageIDs again in the channel. They are already cached in `cachedAcks`.

> Does Community has some Performance Test Data?

Currently not. Could you provide more information to diagnosis? For example, the prometheues metrics  `ProcessingTime` records the processing time between a message is received and acknowledged. You can also adjust the `AckGroupingOption` to see if there are some improvements, e.g. increasing the `MaxTime` or `MaxSize`. @panszobe

shibd:
> @shibd What's the benefit of setting this channel size greater than 0? I think we don't need to cache these MessageIDs again in the channel. They are already cached in cachedAcks.

Just guessing. For example, if the channel has buffers while the thread is flushing data, then the add operation will not be blocked.

panszobe:
> 
> > Does Community has some Performance Test Data?
> 
> Currently not. Could you provide more information to diagnosis? For example, the prometheues metrics `ProcessingTime` records the processing time between a message is received and acknowledged. You can also adjust the `AckGroupingOption` to see if there are some improvements, e.g. increasing the `MaxTime` or `MaxSize`. A flamegraph pprof is also good for analysis. @panszobe

There are some prometheues metrics in Grafana Graph:

image
image
image

Version label stands for
enable_batch_master: the latest version and `EnableBatchIndexAcknowledgment` set true
disable_batch_master: the latest version and `EnableBatchIndexAcknowledgment` set false
disable_batch_v0.6.0: version 0.6.0 and `EnableBatchIndexAcknowledgment` not provided

As the metrics shows, receiving and acking messages for **enable_batch_master** is slower than other two versions, processing time metrics looks like it does't have a very large difference.

There are three flamegraph screenshot for **enable_batch_master**:
image
image
image

For *disable_batch_master**:
image
image


As above, compare to **disable_batch_master**, `runEventsLoop`  and `internalSendRequest` takes much more time.


Next I will try to adjust `AckGroupingOption` to see if there are some improvements.
Please have a look, if you can diagnose what's wrong from something provided above is better.
@BewareMyPower @shibd

BewareMyPower:
@panszobe From the flamegraph, I think the root case is the inefficiency process of sending ACKs. The most time of `internalAck` method is `internal.(*rpcClient).RequestOnCnxNoWait`. In this method, `SendRequestNoWait` takes 3/4 of the total time, while the `internal.baseCommand` function also takes about 1/4 of the total time.

While I'm looking forward to your test results of other `AckGroupingOption` configs, I'm going to test the performance locally. Could you give the information of your test data? Including:
- The batch configs of the producer
- The message size
- The ACK type (individual or cumulative?)

BTW, the `flush` method of the ACK grouping tracker only does the following things:
1. Call `partitionConsumer#sendIndividualAck` or `partitionConsumer#sendCumulativeAck`, which just send the ACKs to the channel `eventsCh`
2. Remove some elements from the map.

I didn't see the `flushIndividualAcks` method from the flamegraph so it looks like it's not the bottle neck. /cc @shibd

panszobe:
> @panszobe From the flamegraph, I think the root case is the inefficiency process of sending ACKs. The most time of `internalAck` method is `internal.(*rpcClient).RequestOnCnxNoWait`. In this method, `SendRequestNoWait` takes 3/4 of the total time, while the `internal.baseCommand` function also takes about 1/4 of the total time.
> 
> While I'm looking forward to your test results of other `AckGroupingOption` configs, I'm going to test the performance locally. Could you give the information of your test data? Including:
> 
> * The batch configs of the producer
> * The message size
> * The ACK type (individual or cumulative?)

The situations of producer: about 20MB/s, 200k messages/s, batch = 10. Through JAVA pressure test procedure.
ACK type is individual, we use `AckID` method to commit. SubscriptionType is Shared.

I adjusted the `AckGroupingOption` configs at 09:49, increasing MaxSize to 1000 and MaxTime to 1000, metrics monitor as below:
wecom-temp-21e4577459e75e0e33510de1559f2726
It seems that the adjustment does not have any improvement.

Looking forward to your local test. @BewareMyPower

BewareMyPower:
> increasing MaxSize to 1000 and MaxTime to 1000,

The default MaxSize is 1000, which means every time 1000 messages are acknowledged, the ACK request will be sent. Increasing MaxTime to 1000 won't help because the receiving speed is far more than 1000 messages per second.

BTW, I just remembered an important thing that this PR does not support the list ACK. See https://github.com/apache/pulsar-client-go/blob/7d257b01b1216e0009495a4aaf83b1716ba458a9/pulsar/ack_grouping_tracker.go#L72. i.e. each time 1000 message IDs are flushed, 1000 ACK requests will be sent. So it does not make a difference.

I'm going to implement the list ACK soon and see if it works.

panszobe:
> BTW, I just remembered an important thing that this PR does not support the list ACK. See
> 
> https://github.com/apache/pulsar-client-go/blob/7d257b01b1216e0009495a4aaf83b1716ba458a9/pulsar/ack_grouping_tracker.go#L72
> 
Emmm, I made a mistake while adjusting the options. As your diagnosis, I adjusted the options by increasing the two settings to 10000, which still didn't make a difference.  
Maybe the break point is your new PR for supporting the list ACK.

BewareMyPower:
Okay. When I open a new PR, I will also perform a simple test in my local env to see if it works.

panszobe:
@BewareMyPower  hi~
Seeing the new PR, I used latest verision(v0.9.1-0.20230301160414-42ded0d59c46) to test today, but it seems that it has no improvements.
image
image
image
image

According to the flamegraphs as above,   SendRequest and MessageReceive took much time.

image
As above, consuming rates and message receving rates also show the test result.


The subscription is very simple, just receiving and acking the Message, has no business processing logic, as below:
image

But it still could not catch up with the producer's production while the other one which disables `BatchIndexAcknowledgment` has no large backlog and keeps stable.

Do you have any other idea? Or there is any optimizations doing now?

BewareMyPower:
@panszobe I'm working on this issue currently. But since I'm busy on something else this week, the PR was not opened. If you mean https://github.com/apache/pulsar-client-go/pull/968, then it's not related to this issue. It's just a problem I found when I wrote the code.

panszobe:
> @panszobe I'm working on this issue currently. But since I'm busy on something else this week, the PR was not opened. If you mean #968, then it's not related to this issue. It's just a problem I found when I wrote the code.

OK. Looking forward to the new PR. Thanks~

PR #959, opened 2023-02-17
add pulsar admin ability in pulsar-client-go

Shoothzj:
### Motivation
I would like to express that the current Pulsar client for Go (pulsar-client-go) is missing the pulsar Admin API. As such, I would like to propose that we work towards adding this feature to pulsar-client-go.
see mail list:
https://lists.apache.org/thread/pd07btsz37mn1hw9o68ll88vf3toxs3m


BewareMyPower:
There is already a Golang admin tool project: https://github.com/streamnative/pulsarctl

Not sure if it has been exposed the API yet. If yes, we can add the dependency.

In addition, this PR is really a huge PR, it's better to split it into multiple PRs.

Shoothzj:
@BewareMyPower 
This is a massive PR submission, but I currently have no plans to merge it. I have marked it to draft.

I'd like to open a thread on whether some of the HTTP APIs can be merged into the official Apache client repository, and merging some of the features of pulsarctl is also awesome.

PR #960, opened 2023-02-21
[improve] Athenz auth provider supports Copper Argos

massakam:
### Motivation

As with the Java client (cf. https://github.com/apache/pulsar/pull/19445), make the Athenz auth provider for the Go client support [Copper Argos](https://github.com/AthenZ/athenz/blob/master/docs/copper_argos.md).

### Modifications

Added new parameters `x509CertChain` and `caCert` to the Athenz auth provider. If the parameter `x509CertChain` is passed, the auth provider assumes that Copper Argos is used and presents the X.509 certificate to obtain a role token from ZTS. Existing users will not be affected by this change.

### Verifying this change

### DIFF
--- a/pulsar/auth/athenz.go
+++ b/pulsar/auth/athenz.go
@@ -32,28 +32,33 @@ import (
 )
 
 const (
-	minExpire            = 2 * time.Hour
-	maxExpire            = 24 * time.Hour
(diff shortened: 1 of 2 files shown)


hrsakai (approved):
LGTM

nodece (approved):
Good!

Issue #937, opened 2023-01-03
[feature request] Support reader interceptor


shibd:
## Motivation

Refer https://github.com/apache/pulsar/issues/14642

PR #961, opened 2023-02-21
[feat] Add Reader interceptor

dao-jun:
Fixes https://github.com/apache/pulsar-client-go/issues/937

### Motivation

Add Reader interceptor

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation


dao-jun:
Tests to be completed

shibd:
@dao-jun Can you resolve the conflict and add a unit test?

treuherz:
It would be much easier to use this interceptor with tracing libraries like OpenTelemetry if a Context was included in the signature. I've gone into more detail in [this other PR](https://github.com/apache/pulsar-client-go/pull/1157) but the short version is that if span IDs, timings, etc can be passed through the context then it removes the need to maintain an external data structure, managed by the interceptor, for every request.

PR #969, opened 2023-02-28
Upgrade the pulsar version that unit tests depend on to 2.11.0

shibd:
### Motivation
Pulsar 2.11.0 has been [released](https://github.com/apache/pulsar/releases/tag/v2.11.0). Upgrade the versions that unit tests depend on so that they are compatible with the latest versions and support new features.

### Modifications
- Upgrade the pulsar version that unit tests depend on to 2.11.0.
- The `GetTopicsOfNamespace` method filters system topics. (This is the break change for broker v2.11.0, which has started an [email discussion](https://lists.apache.org/thread/qhm0zbfw0wfd1m5vj977w002qmkr5lt9). Temporarily fix it on the client side first)
- Fix incorrect HTTP requests. (Because in v2.11.0 will [validate bad requests](https://github.com/apache/pulsar/pull/16577))
- Remove key_shared subscribe consumer unit test-related assert(that every consumer will receive the message), Because consistent hashing is [enabled by default in v2.11.0](https://github.com/apache/pulsar/pull/13352), in scenarios where the number of keys is small, all messages may be sent to the same consumer.

### Verifying this change
- All tests passed.

  - Dependencies (does it add or upgrade a dependency): (upgrade)
  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

  - If yes, how is the feature documented? (not documented)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ ARG GOLANG_IMAGE=golang:latest
 FROM $PULSAR_IMAGE as pulsar
 FROM $GOLANG_IMAGE
 
-RUN apt-get update && apt-get install -y openjdk-11-jre-headless ca-certificates
+RUN apt-get update && apt-get install -y openjdk-17-jre-headless ca-certificates
 
 COPY --from=pulsar /pulsar /pulsar
 
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@
 #
 
 IMAGE_NAME = pulsar-client-go-test:latest
-PULSAR_VERSION ?= 2.10.3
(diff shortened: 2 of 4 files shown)


shibd:
> I think we should not remove the Key_Shared subscription related tests. They are used to protect the case when `subscriptionKeySharedUseConsistentHashing` is false.
> 
> See #970 and [apache/pulsar#13963](https://github.com/apache/pulsar/pull/13963)

Agree, In fact, it was introduced by #953

shibd:
wait v2.11.1 release of pulsar

Issue #971, opened 2023-02-28
Flaky test: TestConsumerSeekByTimeOnPartitionedTopic


BewareMyPower:
See [example failure](https://github.com/apache/pulsar-client-go/actions/runs/4293872087/jobs/7483063543)

logs: 
[logs.txt](https://github.com/apache/pulsar-client-go/files/10851964/logs.txt)

stacks: 
[stacks.txt](https://github.com/apache/pulsar-client-go/files/10851952/stacks.txt)

PR #972, opened 2023-02-28
[refactor]: Refactor the toTrackingMessageID()

Gleiphir2769:
### Motivation

`toTrackingMessageID()` is a function that is widely used in `consumer` implementation. It can convert an interface type  `MessageID` into an struct type `trackingMessageID`. In addition, the second return value also plays a role in checking the `MessageID` type. In other words, it indicates that `MessageID` **cannot** be a user-defined type. From the perspective of code readability, `toTrackingMessageID()` should not do both.

**Note**: After #968 , `toTrackingMessageID()` returns only a pointer now. The role of original `ok` is replaced by nil pointer  now. However, the main content discussed in this PR has not changed.

For example.

https://github.com/apache/pulsar-client-go/blob/e2ea255052e8a527091791ef368851d885ee2d45/pulsar/consumer_regex.go#L176-L181

This example is the correct usage. The `ok` returned by `toTrackingMessageID()` is used to reject user-defined `MessageID`.

https://github.com/apache/pulsar-client-go/blob/e2ea255052e8a527091791ef368851d885ee2d45/pulsar/consumer_partition.go#L470-L473

This example is a bit vague. The actual effect here is the same as the previous example. But it return an error `failed to convert trackingMessageID` which is confusing. 

https://github.com/apache/pulsar-client-go/blob/e2ea255052e8a527091791ef368851d885ee2d45/pulsar/consumer_partition.go#L1816-L1820

In this case. We just want to convert `MessageID` into `trackingMessageID`. We do not care what it really is because it's not possible an invalid `MessageID` implementation.

So, original `toTrackingMessageID()` needs to require a careful look to use it correctly. I think it would be better to split it into two different method. `toTrackingMessageID()` just do the struct conversion, which it's more clearly. And when the new messageID type is added, we can just modify the `checkMessageIDType`.

### Modifications

- Refactor the `toTrackingMessageID()`
- Add the `checkMessageIDType()` to check whether `MessageID` is user-defined.

### Verifying this change

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -538,6 +538,10 @@ func (c *consumer) ReconsumeLaterWithCustomProperties(msg Message, customPropert
 	}
 
 	msgID := c.messageID(msg.ID())
+	if msgID == nil {
+		return
+	}
+
 	props := make(map[string]string)
 	for k, v := range msg.Properties() {
 		props[k] = v
@@ -589,6 +593,10 @@ func (c *consumer) Nack(msg Message) {
 	}
 	if c.options.EnableDefaultNackBackoffPolicy || c.options.NackBackoffPolicy != nil {
 		mid := c.messageID(msg.ID())
+		if mid == nil {
+			return
+		}
+
 		if mid.consumer != nil {
 			mid.consumer.NackID(msg.ID())
 			return
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -531,10 +531,13 @@ func (c *consumer) ReconsumeLaterWithCustomProperties(msg Message, customPropert
 	if delay < 0 {
 		delay = 0
 	}
-	msgID := c.messageID(msg.ID())
-	if msgID == nil {
+
+	if !checkMessageIDType(msg.ID()) {
+		c.log.Warnf("invalid message id type %T", msg.ID())
 		return
 	}
+
+	msgID := c.messageID(msg.ID())
 	props := make(map[string]string)
 	for k, v := range msg.Properties() {
 		props[k] = v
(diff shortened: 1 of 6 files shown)


Gleiphir2769:
> /pulsarbot run-failure-checks

Still doesn't work...

BewareMyPower:
The flaky test is related to https://github.com/apache/pulsar-client-go/issues/971

Gleiphir2769:
> The flaky test is related to #971

@BewareMyPower yes, rerun CI failed again. Looks like CI is stucked in `TestConsumerSeekByTimeOnPartitionedTopic` now.

BewareMyPower (approved):

pulsar/consumer_impl.go
```diff
@@ -580,12 +583,12 @@ func (c *consumer) ReconsumeLaterWithCustomProperties(msg Message, customPropert
 }
 
 func (c *consumer) Nack(msg Message) {
+	if !checkMessageIDType(msg.ID()) {
+		c.log.Warnf("invalid message id type %T", msg.ID())
+		return
+	}
 	if c.options.EnableDefaultNackBackoffPolicy || c.options.NackBackoffPolicy != nil {
 		mid := c.messageID(msg.ID())
-		if mid == nil {
```
shibd, line 585: Even if we check the type of msgID, we can't remove this check, because there is other logic that returns `nil`

https://github.com/apache/pulsar-client-go/blob/48b7d0195327a81f638a3c51df2010d7eb244bbe/pulsar/consumer_impl.go#L755-L759

We can add `checkMessageIdType` to `c. messageID` method, and keep this check.

shibd (changes requested):

pulsar/consumer_impl.go
Gleiphir2769: Great suggestion! I will fix it.

shibd (approved):

RobertIndie (approved):

PR #976, opened 2023-03-03
[feat]: Support auto scaled consumer receiver queue

Gleiphir2769:
Master Issue: #927 

### Motivation
**Note: This is part of the work for [PIP 74](https://github.com/apache/pulsar/wiki/PIP-74%3A-Pulsar-client-memory-limits) in go client.**

The memory limit in consumer side relys on flow control. So we need to firstly support auto scaled consumer receiver queue before we implement the consumer side memory limitation.

#### Why should we support auto scaled consumer first before supporting consumer client memory limit?

The main memory consumption on the consumer side comes from the messages that the consumer prefetches from the broker. Currently, the number of prefetched messages depends on the "current available permits" which represents the messages that have been delivered to the user. When the `availablePermits` reach a certain threshold, the client will send a `Flow` request to the broker to prefetch messages. The current threshold is half of the "receiverQueueSize" (default 1000).

https://github.com/apache/pulsar-client-go/blob/352c463194916cab7772c0729c9cdac14404e8bd/pulsar/consumer_partition.go#L194-L207

If we do not support auto scaled consumer receiver queue, the `Flow` threshold will be fixed. We cannot dynamically adjust the number of prefetched messages, resulting in the inability to control consumer memory. So auto scaled consumer receiver queue needs to be supported first.

#### How does "auto scaled consumer receiver queue" work?

In consumer side, all controls related to message pulling are actually controlling the `Flow`. For example, `receiverQueueSize` controls the `Flow` by controlling the size of the receiver queue.  ReceiverQueue `queueCh` can receive more messages than `receiverQueueSize` because queueCh is a `chan []*message`.

https://github.com/apache/pulsar-client-go/blob/352c463194916cab7772c0729c9cdac14404e8bd/pulsar/consumer_partition.go#L296

If we want to control the `Flow` threshold, we should make the `queueSize` can be auto scaled. There are mainly two problems to be solved.

- Under what condition is the queue considered to need to be scaled?

When the consumption rate required by the user is greater than the configured rate, we increase the configured rate. This condition can be expressed as, before triggering a new Flow to obtain new messages, the number of messages that the current consumer has received (i.e., current available permits + current messages in `queueCh`) is greater than the current queue size. 

- When is the timing to scale receiver queue size?

In the Java client, `expectMoreIncomingMessages` is triggered by "empty reveive".

https://github.com/apache/pulsar/blob/da30b2e211d0a097abfa8d0392f1fd4b13a46328/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L478-L482

In the Go Client, we can not get the "empty reveive" because the `messageCh` is exposed to user to async consume. So we `expectMoreIncomingMessages` when `messageCh` is not full.

https://github.com/apache/pulsar-client-go/blob/352c463194916cab7772c0729c9cdac14404e8bd/pulsar/consumer_partition.go#L1278-L1283

More details in #927 .

### Modifications

- Add `currentQueueSize` and `scaleReceiverQueueHint` to control `Flow` rate.
- Add `AutoScaledReceiverQueueSize` option for `Consumer`

### Verifying this change

Here is a test with pulsar-perf on local standalone server to confirm the effect of this PR.

Non-partitioned topic consumer

```
$  ./pulsar-perf produce my-test -r 1000 -s 128 -b 0 
$  ./pulsar-perf consume my-test -r 10000 --enable-auto-scaled-queue-size=true
```

| RATE  | Final receiver queue size |
| ----- | ------------------------- |
| 1     | 1                         |
| 5     | 2                         |
| 10    | 2                         |
| 100   | 8                         |
| 1000  | 64                        |
| 10000 | 1000                      |

3-partitioned topic consumer

```
$  ./pulsar-perf produce my-test-partitioned -r 1000 -s 128 -b 0 
$  ./pulsar-perf consume my-test-partitioned -r 10000 --enable-auto-scaled-queue-size=true
```

| RATE  | Final total receiver queue size | Sub-consumers receiver queue size |
| ----- | ------------------------------- | --------------------------------- |
| 1     | 4                               | 2,1,1                             |
| 10    | 8                               | 4,2,2                             |
| 100   | 16                              | 4,8,4                             |
| 1000  | 192                             | 64,64,64                          |
| 10000 | 1536                            | 512,512,512                       |
| 20000 | 3000                            | 1000,1000,1000                    |

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"container/list"
+	"context"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -78,6 +79,10 @@ const (
 )
 
 const (
+	initialReceiverQueueSize = 1
+)
+
+const (
 	noMessageEntry = -1
 )
 
@@ -89,6 +94,7 @@ type partitionConsumerOpts struct {
 	subscriptionInitPos         SubscriptionInitialPosition
 	partitionIdx                int
 	receiverQueueSize           int
+	autoReceiverQueueSize       bool
 	nackRedeliveryDelay         time.Duration
 	nackBackoffPolicy           NackBackoffPolicy
 	metadata                    map[string]string
@@ -142,10 +148,13 @@ type partitionConsumer struct {
 	availablePermits *availablePermits
 
 	// the size of the queue channel for buffering messages
-	queueSize       int32
-	queueCh         chan []*message
+	maxQueueSize    int32
+	queueCh         chan *message
 	startMessageID  atomicMessageID
 	lastDequeuedMsg *trackingMessageID
+
+	currentQueueSize       atomic.Int32
+	scaleReceiverQueueHint atomic.Bool
 
 	eventsCh        chan interface{}
 	connectedCh     chan struct{}
@@ -187,16 +196,32 @@ type availablePermits struct {
 
 func (p *availablePermits) inc() {
 	// atomic add availablePermits
-	ap := atomic.AddInt32(&p.permits, 1)
+	p.add(1)
+}
 
+func (p *availablePermits) add(delta int32) {
+	atomic.AddInt32(&p.permits, delta)
+	p.flowIfNeed()
+}
+
+func (p *availablePermits) reset() {
+	atomic.StoreInt32(&p.permits, 0)
+}
+
+func (p *availablePermits) get() int32 {
+	return atomic.LoadInt32(&p.permits)
+}
+
+func (p *availablePermits) flowIfNeed() {
 	// TODO implement a better flow controller
 	// send more permits if needed
-	flowThreshold := int32(math.Max(float64(p.pc.queueSize/2), 1))
-	if ap >= flowThreshold {
-		availablePermits := ap
-		requestedPermits := ap
+	current := p.get()
+	flowThreshold := int32(math.Max(float64(p.pc.currentQueueSize.Load()/2), 1))
+	if current >= flowThreshold {
+		availablePermits := current
+		requestedPermits := current
 		// check if permits changed
-		if !atomic.CompareAndSwapInt32(&p.permits, ap, 0) {
+		if !atomic.CompareAndSwapInt32(&p.permits, current, 0) {
 			return
 		}
 
@@ -205,10 +230,6 @@ func (p *availablePermits) inc() {
 			p.pc.log.WithError(err).Error("unable to send permits")
 		}
 	}
-}
-
-func (p *availablePermits) reset() {
-	atomic.StoreInt32(&p.permits, 0)
 }
 
 // atomicMessageID is a wrapper for trackingMessageID to make get and set atomic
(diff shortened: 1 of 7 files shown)


pulsar/consumer_partition.go
```diff
@@ -1045,8 +1072,14 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 		messages = append(messages, msg)
 	}
 
+	if len(pc.queueCh) == 0 && len(pc.messageCh) == 0 {
```
shibd, line 1075: Why judge the length of `queueCh` and `messageCh`?

shibd (commented):
Great! After a preliminary look, I left some comments.

I see that the unit tests are commented on, and you can finish them up. Can you do a benchmark? Just like this one: https://github.com/apache/pulsar/pull/14494#issuecomment-1072991108

pulsar/consumer_test.go
```diff
@@ -3988,3 +3988,89 @@ func runBatchIndexAckTest(t *testing.T, ackWithResponse bool, cumulative bool, o
...
+		Payload: []byte("hello"),
+	})
+	assert.NoError(t, err)
+
+	// this will trigger receiver queue size expanding to 2 because we have prefetched 1 message >= currentSize 1.
+	_, err = c.Receive(context.Background())
+	assert.Nil(t, err)
+
+	// waiting for prefetched message passing from queueCh to dispatcher()
+	time.Sleep(time.Second)
```
shibd, line 4034: Is there any other way to avoid sleep?

pulsar/consumer_partition.go
```diff
@@ -1273,6 +1325,10 @@ func (pc *partitionConsumer) dispatcher() {
 
 			messages = nil
 
+			if pc.options.autoReceiverQueueSize {
+				pc.incomingMessages.Store(0)
```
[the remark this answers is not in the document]
Gleiphir2769: Makes sense, I'll modify it.

pulsar/consumer_test.go
Gleiphir2769: L4034 time.Sleep can be removed. But if remove L4046, `currentQueueSize` may be 2 or 3 because `flow` and `messageReceive` are concurrent.

pulsar/consumer_test.go
```diff
@@ -4030,9 +4030,6 @@ func TestConsumerWithAutoScaledQueueReceive(t *testing.T) {
 	_, err = c.Receive(context.Background())
 	assert.Nil(t, err)
 
-	// waiting for prefetched message passing from queueCh to dispatcher()
-	time.Sleep(time.Second)
```
shibd: CI failed, it seems we can't remove this.

shibd: You can use `retryAssert` instead of sleep.

https://github.com/apache/pulsar-client-go/blob/e2ea255052e8a527091791ef368851d885ee2d45/pulsar/helper_for_test.go#L176-L185

shibd (approved):

PR #471, opened 2021-02-22
[Issue 444][fix] Use newError to build return error

jonyhy96:
Fixes #444

### Motivation

For better error assertion in client side.

### Modifications

Fix newError and use newError to build return error in most case.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

This change is a trivial rework / code cleanup without any test coverage.

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency):  no
  - The public API:  no
  - The schema:  no
  - The default values of configurations:  no
  - The wire protocol: no

### Documentation

  - Does this pull request introduce a new feature? no
  - If yes, how is the feature documented? GoDocs

PR #984, opened 2023-03-07
[feat][txn]Implement transactionImpl

liangyepianzhou:
Master Issue:https://github.com/apache/pulsar-client-go/issues/932
### Motivation
Implement transaction coordinator client.
### Modifications
1. Implement transaction coordinator
2. implement transactionImpl
3. Implement transaction in producer and consumer API
### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/error.go
+++ b/pulsar/error.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"fmt"
+
 	proto "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
 )
 
--- a/pulsar/error.go
+++ b/pulsar/error.go
@@ -109,9 +109,6 @@ const (
 	SchemaFailure
 	// InvalidStatus means the component status is not as expected.
 	InvalidStatus
-	// TransactionNoFoundError The transaction is not exist in the transaction coordinator, It may be an error txn
-	// or already ended.
-	TransactionNoFoundError
 	// ClientMemoryBufferIsFull client limit buffer is full
 	ClientMemoryBufferIsFull
 )
@@ -228,12 +225,5 @@ func getResultStr(r Result) string {
 }
 
 func getErrorFromServerError(serverError *proto.ServerError) error {
-	switch *serverError {
-	case proto.ServerError_TransactionNotFound:
-		return newError(TransactionNoFoundError, serverError.String())
-	case proto.ServerError_InvalidTxnStatus:
-		return newError(InvalidStatus, serverError.String())
-	default:
-		return newError(UnknownError, serverError.String())
-	}
+	return newError(Result(*serverError), serverError.String())
 }
--- a/pulsar/error.go
+++ b/pulsar/error.go
@@ -109,6 +109,9 @@ const (
 	SchemaFailure
 	// InvalidStatus means the component status is not as expected.
 	InvalidStatus
+	// TransactionNoFoundError The transaction is not exist in the transaction coordinator, It may be an error txn
+	// or already ended.
+	TransactionNoFoundError
 	// ClientMemoryBufferIsFull client limit buffer is full
 	ClientMemoryBufferIsFull
 )
@@ -219,11 +222,20 @@ func getResultStr(r Result) string {
 		return "SchemaFailure"
 	case ClientMemoryBufferIsFull:
 		return "ClientMemoryBufferIsFull"
+	case TransactionNoFoundError:
+		return "TransactionNoFoundError"
 	default:
 		return fmt.Sprintf("Result(%d)", r)
 	}
 }
 
 func getErrorFromServerError(serverError *proto.ServerError) error {
-	return newError(Result(*serverError), serverError.String())
+	switch *serverError {
+	case proto.ServerError_TransactionNotFound:
+		return newError(TransactionNoFoundError, serverError.String())
+	case proto.ServerError_InvalidTxnStatus:
+		return newError(InvalidStatus, serverError.String())
+	default:
+		return newError(UnknownError, serverError.String())
+	}
 }
--- a/pulsar/error.go
+++ b/pulsar/error.go
@@ -17,7 +17,10 @@
 
 package pulsar
 
-import "fmt"
+import (
+	"fmt"
+	proto "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
+)
 
 // Result used to represent pulsar processing is an alias of type int.
 type Result int
@@ -103,14 +106,11 @@ const (
 	ProducerClosed
 	// SchemaFailure means the payload could not be encoded using the Schema
 	SchemaFailure
-
-	// ReachMaxPendingOps means the pending operations in transaction_impl coordinator reach the maximum.
-	ReachMaxPendingOps
 	// InvalidStatus means the component status is not as expected.
 	InvalidStatus
-	// TransactionError means this is a transaction related error
-	TransactionError
-
+	// TransactionNoFoundError The transaction is not exist in the transaction coordinator, It may be an error txn
+	// or already ended.
+	TransactionNoFoundError
 	// ClientMemoryBufferIsFull client limit buffer is full
 	ClientMemoryBufferIsFull
 )
@@ -223,5 +223,16 @@ func getResultStr(r Result) string {
 		return "ClientMemoryBufferIsFull"
 	default:
 		return fmt.Sprintf("Result(%d)", r)
+	}
+}
+
+func getErrorFromServerError(serverError *proto.ServerError) error {
+	switch *serverError {
+	case proto.ServerError_TransactionNotFound:
+		return newError(TransactionNoFoundError, serverError.String())
+	case proto.ServerError_InvalidTxnStatus:
+		return newError(InvalidStatus, serverError.String())
+	default:
+		return newError(UnknownError, serverError.String())
 	}
 }
--- a/pulsar/transaction.go
+++ b/pulsar/transaction.go
@@ -17,7 +17,32 @@
 
 package pulsar
 
+import (
+	"context"
+)
+
+const (
+	Open = iota //The init state is open
+	Committing
+	Aborting
+	Committed
+	Aborted
+	Errored
+	TimeOut
+)
+
 type TxnID struct {
 	mostSigBits  uint64
 	leastSigBits uint64
+}
+
+// Transaction used to guarantee exactly-once
+type Transaction interface {
+	Commit(context.Context) error
+
+	Abort(context.Context) error
+
+	GetState() State
+
+	GetTxnID() TxnID
 }
--- a/pulsar/transaction.go
+++ b/pulsar/transaction.go
@@ -42,7 +42,7 @@ type Transaction interface {
 
 	Abort(context.Context) error
 
-	GetState() State
+	GetState() TxnState
 
 	GetTxnID() TxnID
 }
--- a/pulsar/transaction.go
+++ b/pulsar/transaction.go
@@ -21,8 +21,11 @@ import (
 	"context"
 )
 
+type TxnState int32
+
 const (
-	Open = iota //The init state is open
+	_ TxnState = iota
+	Open
 	Committing
 	Aborting
 	Committed
--- a/pulsar/transaction.go
+++ b/pulsar/transaction.go
@@ -27,13 +27,24 @@ type TxnState int32
 
 const (
 	_ TxnState = iota
+	// Open The transaction in Open state can be used to send/ack messages.
 	Open
+	// Committing The state of the transaction will be Committing after the commit method is called.
+	// The transaction in Committing state can be committed again.
 	Committing
+	// Aborting The state of the transaction will be Aborting after the abort method is called.
+	// The transaction in Aborting state can be aborted again.
 	Aborting
+	// Committed The state of the transaction will be Committed after the commit method is executed success.
+	// This means that all the operations with the transaction are success.
 	Committed
+	// Aborted The state of the transaction will be Aborted after the abort method is executed success.
+	// This means that all the operations with the transaction are aborted.
 	Aborted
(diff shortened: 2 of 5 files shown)


pulsar/transaction_impl.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func (txn *transaction) registerSendOrAckOp() error {
+	if txn.opsCount.Inc() == 1 {
+		//There are new operations that not completed
+		select {
+		case <-txn.opsFlow:
+			return nil
+		case <-time.After(txn.opTimeout):
+			if txn.state != Open {
```
nodece, line 147: Please call the `txn.checkIfOpen`.

nodece (approved):

pulsar/transaction.go
```diff
@@ -17,7 +17,32 @@
 
 package pulsar
 
+import (
+	"context"
+)
+
+const (
+	Open = iota //The init state is open
```
BewareMyPower, line 25:
```suggestion
	Open = iota
```

Please add the API docs instead of just a simple comment.

pulsar/transaction_impl.go
```diff
@@ -0,0 +1,235 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
+	"github.com/apache/pulsar-client-go/pulsar/log"
+	uAtomic "go.uber.org/atomic"
+)
+
+type TxnState int32
```
BewareMyPower, line 31: I think `TxnState` should be moved to `transaction.go` because it's a public type.

```go
type TxnState int32

const (
    Open TxnState = iota
    // ...
)

File: pulsar/transaction_impl.go [hunk 4]
```diff
@@ -0,0 +1,235 @@
...
+	*        Write:
+	* 		      1. When the transaction is created, a new empty struct{} will be written to opsFlow chan.
+	*			  2. When the opsCount decrement from 1 to 0, a new empty struct{} will be written to opsFlow chan.
+	* 			  3. When get a retryable error after committing or aborting the transaction,
+	*	             a new empty struct{} will be written to opsFlow chan.
+	*        Read:
+	*			  1. When the transaction is committed or aborted, an empty struct{} will be read from opsFlow chan.
+	*			  2. When the opsCount increment from 0 to 1, an empty struct{} will be read from opsFlow chan.
+	 */
+	opsFlow   chan void
```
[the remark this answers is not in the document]
BewareMyPower: I agree with @shibd. The `opsFlow` is hard to understand. We usually create a channel to handle all events in a separated goroutine, and send events to this channel from various goroutines.

See https://go.dev/blog/codelab-share

pulsar/error.go
```diff
@@ -225,3 +226,14 @@ func getResultStr(r Result) string {
 		return fmt.Sprintf("Result(%d)", r)
 	}
 }
+
+func getErrorFromServerError(serverError *proto.ServerError) error {
```
BewareMyPower, line 230: Maybe we can replace this function with a simple call like:

```go
newError(Result(*serverError), serverError.String())
```

Because `Result` and `proto.ServerError` should be just the same.

BewareMyPower (changes requested):

pulsar/transaction_impl.go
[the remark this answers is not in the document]
liangyepianzhou: The `opsFlow` has two effects:
1. It can be a protection mechanism to prevent sending or confirming messages with an ended transaction.
     Using the end transaction to send a message makes it easy to cause the stable position to not move for a long time. And this causes a read delay.
2. It can be used as a future to wait for all operations complete before committing or aborting transactions.

pulsar/transaction_impl.go
```diff
@@ -0,0 +1,232 @@
...
+		txn.opsFlow <- true
+	}
+}
+
+func (txn *transaction) registerProducerTopic(topic string) error {
+	isOpen, err := txn.checkIfOpen()
+	if !isOpen {
+		return err
+	}
+	_, ok := txn.registerPartitions[topic]
```
shibd, line 169: This map also needs to be protected with a lock, right?

pulsar/transaction_impl.go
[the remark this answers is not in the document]
shibd: oh, I see.

When `registerProducerTopic` and `registerAckTopic` operations are being performed, commits and aborts are also not allowed, right?

I see that the future of these Java operations is linked:
- `registerAckedTopic` future -> `registerAckOp`
- `registerProducedTopic` future -> `registerSendOp`


I'm not quite sure what to do with the go client next, maybe you can add some more complete tests?

`registerSendOrAckOp` and `endSendOrAckOp` and `registerProducerTopic` and `registerAckTopic`  these complete a test scenario overlay.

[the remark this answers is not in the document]
liangyepianzhou: There are some tests for `registerSendOrAckOp` and `endSendOrAckOp` and `registerProducerTopic` and `registerAckTopic`.
* `TestRegisterOpAndEndOp`
* `TestRegisterTopic`

pulsar/transaction_impl.go
liangyepianzhou: Yeah, we should check the map in the lock again.

pulsar/transaction_impl.go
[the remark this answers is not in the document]
shibd: Overall LGTM, I'll wait to see how subsequent PRs use these methods together.

shibd (approved):

pulsar/error.go
```diff
@@ -103,14 +107,8 @@ const (
...
 	// SchemaFailure means the payload could not be encoded using the Schema
 	SchemaFailure
-
-	// ReachMaxPendingOps means the pending operations in transaction_impl coordinator reach the maximum.
-	ReachMaxPendingOps
 	// InvalidStatus means the component status is not as expected.
 	InvalidStatus
-	// TransactionError means this is a transaction related error
-	TransactionError
-
```
BewareMyPower, line 113: Just a note here: it's a breaking change. But these error codes were added in http://github.com/apache/pulsar-client-go/pull/953, which is not included in any official release yet. So it's okay.

liangyepianzhou: Yes, we don't need this exception for now. so you can delete it first.

pulsar/error.go
BewareMyPower: Sorry I might give a wrong suggestion. I just reviewed the `Result` enum again, the `ServerError` cannot be converted to the `Error`. You have adjusted the order to make `InvalidStatus` (`Result`) equal to `ServerError_InvalidTxnStatus` (`ServerError`). However, other error codes might not be the same. So it's better to map `ServerError` to `Result` manually here, like what you did before.

pulsar/transaction_test.go
```diff
@@ -79,6 +81,138 @@ func TestTCClient(t *testing.T) {
...
+*       2. The internal API, registerSendOrAckOp and endSendOrAckOp
+*			1. Register 4 operation but only end 3 operations, the transaction can not be committed or aborted.
+*			2. Register 3 operation and end 3 operation the transaction can be committed and aborted.
+*			3. Register an operation and end the operation with an error,
+*			and then the state of the transaction should be replaced to errored.
+*		3. The internal API, registerAckTopic and registerProducerTopic
+*			1. Register ack topic and send topic, and call http request to get the stats of the transaction
+*              to do verification.
+ */
+/** TestTxnImplCommitOrAbort Test abort and commit txn */
```
BewareMyPower, line 98: Please don't use such type of comments. Use:

```go
// xxx
// yyy
// zzz
```

pulsar/transaction_test.go
```diff
@@ -79,6 +81,138 @@ func TestTCClient(t *testing.T) {
...
+	if commit {
+		err = txn.Commit(context.Background())
+	} else {
+		err = txn.Abort(context.Background())
+	}
+	return err
+}
+
+func createTxn(tc *transactionCoordinatorClient, t *testing.T) *transaction {
+	id3, err := tc.newTransaction(time.Hour)
```
BewareMyPower, line 209: Please don't use a number suffix, I think we can just use `id` here.

pulsar/transaction_test.go
```diff
@@ -79,6 +81,138 @@ func TestTCClient(t *testing.T) {
...
+func TestTxnImplCommitOrAbort(t *testing.T) {
+	tc, _ := createTcClient(t)
+	//1. Open a transaction and then commit it.
+	//The operations of committing txn1 should success at the first time and fail at the second time.
+	txn1 := createTxn(tc, t)
+	err := txn1.Commit(context.Background())
+	if err != nil {
+		t.Fatalf("Failed to commit the transaction %d:%d, %s\n", txn1.txnID.mostSigBits, txn1.txnID.leastSigBits,
+			err.Error())
+	}
```
BewareMyPower, line 108:
```suggestion
        assert.Nil(t, err, "Failed to commit the transaction %d:%d, %s\n",
                txn1.txnID.mostSigBits, txn1.txnID.leastSigBits, err.Error())
```

pulsar/transaction.go
```diff
@@ -17,7 +17,53 @@
...
+	// The transaction in Aborting state can be aborted again.
+	Aborting
+	// Committed The state of the transaction will be Committed after the commit method is executed success.
+	// This means that all the operations with the transaction are success.
+	Committed
+	// Aborted The state of the transaction will be Aborted after the abort method is executed success.
+	// This means that all the operations with the transaction are aborted.
+	Aborted
+	// Errored The state of the transaction will be Errored after the operation of transaction get a non-retryable error.
+	Errored
```
BewareMyPower, line 45: I see the error code is `ERROR` in the Java client. Should we name it with `Error` here? `Errored` looks confusing. We use the `ed` and `ing` suffixes to differ whether a step is complete or ongoing.

pulsar/transaction_test.go
```diff
@@ -79,6 +81,138 @@ func TestTCClient(t *testing.T) {
...
+	txn1.opsFlow <- true
+	err = txn1.Commit(context.Background())
+	assert.Equal(t, err.(*Error).Result(), Result(pb.ServerError_TransactionNotFound))
+	assert.Equal(t, txn1.GetState(), Errored)
+	//2. Open a transaction and then abort it.
+	//The operations of aborting txn2 should success at the first time and fail at the second time.
+	id2, err := tc.newTransaction(time.Hour)
+	if err != nil {
+		t.Fatalf("Failed to new a transaction %s", err.Error())
+	}
```
BewareMyPower, line 119:
```suggestion
	assert.Nil(t, err, "Failed to new a transaction %s\n", err.Error())
```

pulsar/transaction_test.go
```diff
@@ -79,6 +81,138 @@ func TestTCClient(t *testing.T) {
...
+	txn2 := newTransaction(*id2, tc, time.Hour)
+	err = txn2.Abort(context.Background())
+	if err != nil {
+		t.Fatalf("Failed to abort the transaction %d:%d, %s\n", id2.mostSigBits, id2.leastSigBits, err.Error())
+	}
```
BewareMyPower, line 124:
```suggestion
        assert.Nil(t, err, "Failed to abort the transaction %d:%d, %s\n",
                id2.mostSigBits, id2.leastSigBits, err.Error())
```

pulsar/transaction_test.go
```diff
@@ -79,6 +81,138 @@ func TestTCClient(t *testing.T) {
...
+	txn := createTxn(tc, t)
+	//3. Create a topic and register topic and subscription.
+	err = txn.registerAckTopic(topic, sub)
+	if err != nil {
+		t.Fatalf("Failed to register ack topic %s", err.Error())
+	}
+	err = txn.registerProducerTopic(topic)
+	if err != nil {
+		t.Fatalf("Failed to register ack topic %s", err.Error())
+	}
```
BewareMyPower, line 179: Please also replace the check with the `assert.Nil` call.

pulsar/transaction.go
liangyepianzhou: The Error is a conflict with `Error` in error.go.
```java
type Error struct {
	msg    string
	result Result
}
```

liangyepianzhou: We get an unretrievable error and then change the state of the transaction to `Errored`.
So the `Errored` is OK.

BewareMyPower: > The Error is a conflict with Error in error.go.

Okay. Then I think it's better to add a prefix to these enum constants. e.g. `TxnState_Open`, `TxnState_Error`. Otherwise, it could be very easy to cause the naming conflicts. You can also see the generated code in `PulsarApi.pb.go`, the enum constants all have their prefixes like `CompressionType_` and `ProducerAccessMode_`.

BewareMyPower: I just found https://github.com/apache/pulsar-client-go/pull/471 renames the error code so that the prefix `Result` has been removed for all `Result` constants. It's a bad breaking change and I don't know why the reviewers still approved.

But we should avoid the wrong design like that.

pulsar/error.go
BewareMyPower: @liangyepianzhou Please also take a look at this comment. I've thought `Result` and `proto.ServerError` should be just the same before, but it seems that they are already different currently.

pulsar/transaction.go
liangyepianzhou: Okay, I add a prefix to the state. Because the `TxnState_Open` does not conform to Go's specification, I change it to `TxnOpen`

BewareMyPower (approved):

pulsar/error.go
liangyepianzhou: Done

PR #991, opened 2023-03-17
[feat] Support consumer client memory limit

Gleiphir2769:
Master Issue: #927 

### Motivation

This is part of the work for [PIP 74](https://github.com/apache/pulsar/wiki/PIP-74%3A-Pulsar-client-memory-limits) in go client.

More details in https://github.com/apache/pulsar-client-go/issues/927 .

### Modifications

- Consumer adds memory limit relate logic.

### Verifying this change

### Documentation

  - Does this pull request introduce a new feature? (**yes** / no)
  - If yes, how is the feature documented? (not applicable / docs / **GoDocs** / not documented)

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -79,7 +79,6 @@ const (
 const (
 	initialReceiverQueueSize           = 1
 	receiverQueueExpansionMemThreshold = 0.75
-	receiverQueueShrinkMemThreshold    = 0.95
 )
 
 const (
@@ -335,7 +334,7 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 	}
 	if pc.options.autoReceiverQueueSize {
 		pc.currentQueueSize.Store(initialReceiverQueueSize)
-		pc.client.memLimit.RegisterTrigger(receiverQueueShrinkMemThreshold, pc.shrinkReceiverQueueSize)
+		pc.client.memLimit.RegisterTrigger(pc.shrinkReceiverQueueSize)
 	} else {
 		pc.currentQueueSize.Store(int32(pc.options.receiverQueueSize))
 	}
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -155,7 +155,6 @@ type partitionConsumer struct {
 
 	currentQueueSize       uAtomic.Int32
 	scaleReceiverQueueHint uAtomic.Bool
-	receiverQueueShrinking uAtomic.Bool
 	incomingMessages       uAtomic.Int32
 
 	eventsCh        chan interface{}
@@ -336,6 +335,7 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 	}
 	if pc.options.autoReceiverQueueSize {
 		pc.currentQueueSize.Store(initialReceiverQueueSize)
+		pc.client.memLimit.RegisterTrigger(receiverQueueShrinkMemThreshold, pc.shrinkReceiverQueueSize)
 	} else {
 		pc.currentQueueSize.Store(int32(pc.options.receiverQueueSize))
 	}
@@ -1124,7 +1124,7 @@ func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, header
 	}
 
 	if pc.options.autoReceiverQueueSize {
-		pc.reserveMemory(int64(bytesReceived))
+		pc.client.memLimit.ForceReserveMemory(int64(bytesReceived))
 		pc.incomingMessages.Add(int32(len(messages)))
 		pc.markScaleIfNeed()
 	}
@@ -1347,7 +1347,7 @@ func (pc *partitionConsumer) dispatcher() {
 
 			if pc.options.autoReceiverQueueSize {
 				pc.incomingMessages.Dec()
-				pc.releaseMemory(int64(nextMessageSize))
+				pc.client.memLimit.ReleaseMemory(int64(nextMessageSize))
 				pc.expectMoreIncomingMessages()
 			}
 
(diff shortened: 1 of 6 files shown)


pulsar/consumer_partition.go
```diff
@@ -1752,6 +1762,37 @@ func (pc *partitionConsumer) markScaleIfNeed() {
 	}
 }
 
+func (pc *partitionConsumer) reserveMemory(size int64) {
+	pc.client.memLimit.ForceReserveMemory(size)
+	if pc.client.memLimit.CurrentUsagePercent() >= receiverQueueShrinkMemThreshold {
```
shibd, line 1767: This shrinking strategy is different from java client implementation. 

The java client implementation:

Will try to trigger shrinking memory after each `ReserveMemory`, this means that the producer will also trigger the consumer to reduce the queue when sending data. 

https://github.com/shibd/pulsar/blob/82237d3684fe506bcb6426b3b23f413422e6e4fb/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MemoryLimitController.java#L65

Your implementation will cause the following scenario where producers cannot produce data.
- One client instance has `1` producer instance and `3` consumer.
- If `3` consumers run out of memory, the producer's production data will block.

shibd (changes requested):

pulsar/consumer_partition.go
Gleiphir2769: I think `shrink` revoked by `memory limiter` and revoked by `consumer` are equivalent.

> One client instance has 1 producer instance and 3 consumer.
> If 3 consumers run out of memory, the producer's production data will block.

In this scenario, the Java client will also block. Because `consumer` use the `forceReserve`.

By the way, is it necessary for the producer to trigger the consumer `shrink`? If the memory usage exceeds the threshold, the consumer will always shrink it's receiver queue, right?

Gleiphir2769: The reasons here are not the same as those implemented by the Java client: 
1. I don’t think it matters whether this is triggered by `memory limiter` as long as the `shrink` is not triggered repeated. 
3. If we want to implement it like the Java client, we need to modify the `memory limiter` to allow it to support a trigger, but I feel like this in itself is an unnecessary code coupling.

shibd: > By the way, is it necessary for the producer to trigger the consumer shrink? If the memory usage exceeds the threshold, the consumer will always shrink it's receiver queue, right?

Sorry, the previous example is not quite accurate.

I think the key difference is whether one `consumer` or `producer` needs to trigger all consumers to reduce the queue size.

- The java client implementation:
When a producer or consumer call the `ReserveMemory` option, If the memory exceeds the threshold, it triggers all consumers to shrink the queue size

- Your implementation:
When a consumer checks that memory exceeds a threshold, it will only try to reduce its own queue size.

I think there is a memory usage starvation issue here:

- One client instance has `1` producer instance and `3` consumer.
- If the queue of these three consumers is already relatively large, if the user of `consumer-1` and `consumer 2` is not actively called (there are many messages piled up in queurCh), then if `consumer-3` checks and reduces its queue size after receiving a message, although the memory limit is reduced, but `consumer-1` and `consumer-2` still occupy more memory (a lot of pending consumption), 
- The same way, if the producer wants to send a message, Since it cannot trigger other consumers to reduce the queue size, it will also send more slowly.

In short, in this case, Memory is wasted on inactive consumers(consumer-1, consumer-2)

If it is in implementing the Java client, the problem does not exist. 
Each time, all the consumers will reduce the queue size, and then let them re-promote the  queue size according to the activity of the consumers.

Gleiphir2769: > In short, in this case, Memory is wasted on inactive consumers(consumer-1, consumer-2)

OK, I see. But even in Java client, `reduceReceiverQueue()` can not free memory immediately. It just controls next flow to prefetch fewer messages. IMO, the memory limit on the consumer side is lazy. The memory will only be freed when pending messages are consumed.

So I think this question can be summed up like this.
- Even `memory limiter` can trigger all the consumers to reduce it's queue size. The actual memory release still depends on the consumer's consumption.
- If the consumer is not active, even reducing the queue size will not reduce the memory.

shibd: > OK, I see. But even in Java client, reduceReceiverQueue() can not free memory immediately. It just controls next flow to prefetch fewer messages.

> It just controls next flow to prefetch fewer messages.

Yes, We need these inactive consumers to reduce the size of the pull flow so that it can give memory to more active consumers.

pulsar/consumer_partition.go
```diff
@@ -1760,6 +1770,21 @@ func (pc *partitionConsumer) markScaleIfNeed() {
 	}
 }
 
+func (pc *partitionConsumer) shrinkReceiverQueueSize() {
+	if !pc.options.autoReceiverQueueSize {
+		return
+	}
+
+	oldSize := pc.currentQueueSize.Load()
```
shibd, line 1778: We still need `receiverQueueShrinking ` to protect can only be shrinking once at a time.

shibd: oh, I see. Already protected in `memory_limit_controller`.

shibd (approved):

pulsar/consumer_partition.go
```diff
@@ -333,6 +335,7 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 	}
 	if pc.options.autoReceiverQueueSize {
 		pc.currentQueueSize.Store(initialReceiverQueueSize)
+		pc.client.memLimit.RegisterTrigger(receiverQueueShrinkMemThreshold, pc.shrinkReceiverQueueSize)
```
shibd, line 338: This threshold should be managed by the ·memory_limit_controller·, and we don't need to set it when we register trigger.

Gleiphir2769: But `memory_limit_controller` is created by client.  If threshold is managed by `memory_limit_controller`, will the threshold become a magic number? 
I'm not sure if my thinking is correct.

shibd: I think it can be passed in via the constructor(`NewMemoryLimitController`).

In the short, We don't have to set a threshold for each trigger, it should be unique in `memory_limit_controller`

Gleiphir2769: > 

I see. I will modify it.

pulsar/internal/memory_limit_controller.go
```diff
@@ -31,18 +31,41 @@ type MemoryLimitController interface {
...
+
+func (t *thresholdTrigger) canTryRunning() bool {
+	return atomic.CompareAndSwapInt32(&t.triggerRunning, 0, 1)
+}
+
+func (t *thresholdTrigger) setRunning(isRunning bool) {
+	if isRunning {
+		atomic.StoreInt32(&t.triggerRunning, 1)
+	}
+	atomic.StoreInt32(&t.triggerRunning, 0)
```
shibd, line 60:
```suggestion
	} else {
		  atomic.StoreInt32(&t.triggerRunning, 0)
	}
```

Gleiphir2769: Sorry for missing it.
Done.

pulsar/consumer_test.go
```diff
@@ -4113,3 +4113,204 @@ func TestConsumerBatchIndexAckDisabled(t *testing.T) {
...
+	// Use mem-limited client 2 to create consumer c1
+	c2, err := cli2.Subscribe(ConsumerOptions{
+		Topic:                             topic,
+		SubscriptionName:                  "my-sub-2",
+		Type:                              Exclusive,
+		EnableAutoScaledReceiverQueueSize: true,
+	})
+	assert.Nil(t, err)
+	defer c2.Close()
+	pc2 := c1.(*consumer).consumers[0]
```
RobertIndie, line 4277:
```suggestion
	pc2 := c1.(*consumer).consumers[1]
```

RobertIndie (changes requested):

pulsar/consumer_test.go
Gleiphir2769: Done. 
```
pc2 := c2.(*consumer).consumers[0]
```

RobertIndie (approved):

Anonymitaet:
Hi @shibd, thanks for introducing this great feature!

========================================

As we discussed just now, for the doc side:

### ✅ Pulsar feature matrix

I've ticked the boxes of ["Memory Limits" for Go](https://docs.google.com/spreadsheets/d/1YHYTkIXR8-Ql103u-IMI18TXLlGStK8uJjDsOOA0T20/edit#gid=1784579914&range=22:22).

### ✅ Pulsar client docs

#### 🌈 Collobration way

To add docs for it, feel free to choose your preferred method (either way works for me):

- You provide tech inputs to me. I'll write docs based on them and request your review. Or

- You submit a PR to the pulsar-site repo directly and request my review.

#### 🌈 TO-DO tasks

1. Create a tab for Go and a topic named "Set memory limits" in [work with clients](https://pulsar.apache.org/docs/next/client-libraries-clients/). It should explain:
  - What is it? 
  - How to use it? (e.g., any considerations? Sample code, etc)

image

2. This doc is only added to `NEXT` (submit a PR [here](https://github.com/apache/pulsar-site/edit/main/docs/client-libraries-clients.md)) since this feature is server-independent. 

========================================

Feel free to reach out to me if you have questions. TIA!

PR #992, opened 2023-03-17
[Issue 468] Add Support for NonDurable subscriptions

### LINKED ISSUE #468: Add Support for NonDurable subscriptions
The Java API supports Non-Durable Subscriptions via SubscriptionMode
https://github.com/apache/pulsar/blob/master/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ConsumerBuilder.java#L243-L256

but it appears that the Go client does not
https://github.com/apache/pulsar-client-go/blob/25f3075176f853f5f09d89b3b625e0646432735f/pulsar/consumer.go#L76

This is useful in our use case where we have lots of topics with short-lived subscriptions that last minutes to hours and then once data is read the topic is no longer needed.
To ensure that data is compacted in the bookies currently we have to set the subscription expiration time at the namespace level

Please add support for Subscription Mode

Continue to use subscription expiration time

**Additional context**

dinghram:
<!--
### Contribution Checklist
  
  - Name the pull request in the form "[Issue XYZ][component] Title of the pull request", where *XYZ* should be replaced by the actual issue number.
    Skip *Issue XYZ* if there is no associated github issue for this pull request.
    Skip *component* if you are unsure about which is the best component. E.g. `[docs] Fix typo in produce method`.

  - Fill out the template below to describe the changes contributed by the pull request. That will give reviewers the context they need to do the review.
  
  - Each pull request should address only one issue, not mix up code from multiple issues.
  
  - Each commit in the pull request has a meaningful commit message

  - Once all items of the checklist are addressed, remove the above text and this checklist, leaving only the filled out template below.

Fixes #468 

### Motivation
This change allows the Go Pulsar Client to create consumers with non-durable subscriptions. We use Pulsar in an enterprise level application that processes large amounts of data throughput. We are adding a component that creates a consumer on that throughput, and it is very important that the throughput is not affected by this component. If the component creates a durable subscription, then goes down for whatever reason, the backlog caused by a durable subscription could bring our enterprise application down. Allowing a consumer to be non-durable alleviates this problem. 

It has been suggested that we could use a Pulsar Reader instead of a Consumer, but then we cannot easily spread the load of messages across multiple pods, as each pod would not know which messages the other pod had processed. It has also been suggested that we could use "aggressive subscription/topic removal policies" to achieve this, but our enterprise level application has the policies already defined and changing them could have other, unintended side-effects.

This change brings a feature to the Go Pulsar Client which is already supported by the Java client.

### Modifications

This change exposes the SubscriptionMode outside the pulsar package, and adds SubscriptionMode as an option in ConsumerOptions. The default remains "Durable".

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation


shibd (approved):
Nice work!

BewareMyPower (approved):

Issue #993, opened 2023-03-18
Messages unacked when EnableBatchIndexAck is false 


panszobe:
#### Expected behavior

All of the messages should be acked when `EnableBatchIndexAck` is false, as the same behavior of setting `EnableBatchIndexAck` true.

#### Actual behavior

When setting `EnableBatchIndexAck` to false, there were more and more messaged unacked, Grafana dashboards as below:

<img width="1377" alt="image" src="https://user-images.githubusercontent.com/75996911/226072338-060010cb-73ef-451a-a7ff-0212222f921e.png">

And it led to that Pulsar Server would take more and more memory to handle unacked messages though Client could consume messages at the same time, but backlog became larger, and will cause consuming duplicated messages when restarting consumer.

#### System configuration

pulsar-client-go version: master(**v0.9.1-0.20230313030101-bcbac9f2ae5c**)

#### Code Review
I think the root cause is as below:

```
func (pc *partitionConsumer) ackID(msgID MessageID, withResponse bool) error {
	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
		pc.log.WithField("state", state).Error("Failed to ack by closing or closed consumer")
		return errors.New("consumer state is closed")
	}

	if cmid, ok := msgID.(*chunkMessageID); ok {
		return pc.unAckChunksTracker.ack(cmid)
	}

	trackingID := toTrackingMessageID(msgID)

	if trackingID != nil && trackingID.ack() {
		pc.metrics.AcksCounter.Inc()
		pc.metrics.ProcessingTime.Observe(float64(time.Now().UnixNano()-trackingID.receivedTime.UnixNano()) / 1.0e9)
	} else if !pc.options.enableBatchIndexAck {
		return nil
	}

	var ackReq *ackRequest
	if withResponse {
		ackReq := pc.sendIndividualAck(trackingID)
		<-ackReq.doneCh
	} else {
		pc.ackGroupingTracker.add(trackingID)
	}
	pc.options.interceptors.OnAcknowledge(pc.parentConsumer, msgID)
	if ackReq == nil {
		return nil
	}
	return ackReq.err
}

func (t *timedAckGroupingTracker) tryAddIndividual(id MessageID) map[[2]uint64]*bitset.BitSet {
	t.Lock()
	defer t.Unlock()
	key := [2]uint64{uint64(id.LedgerID()), uint64(id.EntryID())}

	batchIdx := id.BatchIdx()
[...]

PR #994, opened 2023-03-18
Fix batched messages not ACKed correctly when batch index ACK is disabled

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-go/issues/993

### Motivation

When batch index ACK is disabled, if N messages in a batch are acknowledged, currently only the batched message ID of the last message will be acknowledged. This behavior is wrong because we need to acknowledge the whole batch.

### Modifications

- Create a new trackingMessageID that has no batch index and size fields for this case
- Add `TestConsumerBatchIndexAckDisabled` to cover this case

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -497,8 +497,13 @@ func (pc *partitionConsumer) ackID(msgID MessageID, withResponse bool) error {
 	}
 
 	trackingID := toTrackingMessageID(msgID)
+	var msgIDToAck MessageID = trackingID
 
 	if trackingID != nil && trackingID.ack() {
+		msgIDToAck = &messageID{
+			ledgerID: trackingID.ledgerID,
+			entryID:  trackingID.entryID,
+		}
 		pc.metrics.AcksCounter.Inc()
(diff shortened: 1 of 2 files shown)


pulsar/consumer_partition.go
```diff
@@ -499,6 +499,12 @@ func (pc *partitionConsumer) ackID(msgID MessageID, withResponse bool) error {
 	trackingID := toTrackingMessageID(msgID)
 
 	if trackingID != nil && trackingID.ack() {
+		trackingID = &trackingMessageID{
```
shibd, line 502: It is best to add some comments explaining why, which can be intuitively understood by developers who are not familiar with this implementation

shibd (approved):

pulsar/consumer_partition.go
BewareMyPower: Good suggestion. I agree with you.

RobertIndie: +1

RobertIndie (approved):

pulsar/consumer_partition.go
BewareMyPower: Addressed.

PR #1006, opened 2023-04-10
[feat][schema] Support the schema type ProtoNativeSchema

gaoran10:
### Motivation

The pulsar-client-go supports ProtoSchema based on the Avro format, users must provide the schema definition string when using struct schema(JSON, Avro, Proto), it's hard to use and we can't get the entire original proto definition. 

The difference between ProtoSchema and ProtoNativeSchema is the schema data, the ProtoNativeSchema will contain the entire original proto definition, and we can use the schema data of the ProtoNativeSchema to generate the corresponding Descriptor and get any field, we can leverage this feature in Pulsar SQL.

The schema data of the ProtoNativeSchema follow the Pulsar [format](https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/protocol/schema/ProtobufNativeSchemaData.java), we can generate JSON data for `ProtobufNativeSchemaData` as the schema data.

```
public class ProtobufNativeSchemaData {

    private byte[] fileDescriptorSet;
    private String rootMessageTypeName;
    private String rootFileDescriptorName;

}
```

The field `fileDescriptorSet` is the serialized data of the message `FileDescriptorSet`, it contains all definition proto files data.

```
message FileDescriptorSet {
  repeated FileDescriptorProto file = 1;
}
```

```
message FileDescriptorProto {
  optional string name = 1;
  optional string package = 2;
  repeated string dependency = 3;
  repeated DescriptorProto message_type = 4;
  repeated EnumDescriptorProto enum_type = 5;
  repeated ServiceDescriptorProto service = 6;
  repeated FieldDescriptorProto extension = 7;
  optional FileOptions options = 8;
  optional SourceCodeInfo source_code_info = 9;
  repeated int32 public_dependency = 10;
  repeated int32 weak_dependency = 11;
  optional string syntax = 12;
} 
```

### Modifications

Add a new schema type `ProtoNative(20)`.
Add ProtoNativeSchema.
Add encode and decode methods for ProtoNativeSchema.

### Verifying this change

Add an integration test to make sure the broker can deserialize the schema data of the ProtoNativeSchema.

  - The public API: (yes)
  - The schema: (yes)
  - The wire protocol: (no)

### Documentation


BewareMyPower:
Could you fix lint error? You can use `golangci-lint run` to check lint errors in your local env.

RobertIndie:
https://github.com/apache/pulsar-client-go/actions/runs/4670464787/jobs/8270224455?pr=1006#step:5:789

pulsar/schema.go
```diff
@@ -219,6 +226,107 @@ func (ps *ProtoSchema) GetSchemaInfo() *SchemaInfo {
...
+	unResolvedDependencies := make([]string, 0)
+	for i := 0; i < fileDesc.Imports().Len(); i++ {
+		if _, ok := protoMap[fileDesc.Imports().Get(i).Path()]; ok {
+			if !ok {
+				unResolvedDependencies = append(unResolvedDependencies, fileDesc.Imports().Get(i).Path())
+			}
+		}
+	}
+	if len(unResolvedDependencies) > 0 {
+		log.Fatalf("%v can't resolve dependency [%v].",
```
nodece, line 308: Could you return an error instead of the log?

pulsar/schema.go
```diff
@@ -219,6 +226,107 @@ func (ps *ProtoSchema) GetSchemaInfo() *SchemaInfo {
...
+	RootFileDescriptorName string `json:"rootFileDescriptorName"`
+}
+
+func GetFileProto(fileDesc protoreflect.FileDescriptor, protoMap map[string]*descriptorpb.FileDescriptorProto) {
+	if fileDesc.Imports().Len() > 0 {
+		for i := 0; i < fileDesc.Imports().Len(); i++ {
+			GetFileProto(fileDesc.Imports().Get(i).ParentFile(), protoMap)
+		}
+	}
+	unResolvedDependencies := make([]string, 0)
```
nodece, line 299: Can we ignore this check?

nodece (changes requested):
Great work! I left some comments.

pulsar/schema.go
```diff
@@ -219,6 +226,107 @@ func (ps *ProtoSchema) GetSchemaInfo() *SchemaInfo {
...
 
+type ProtoNativeSchema struct {
+	SchemaInfo
+}
+
+func NewProtoNativeSchemaWithMessage(message proto.Message, properties map[string]string) *ProtoNativeSchema {
+	return NewProtoNativeSchema(GetProtoNativeSchemaInfo(message), properties)
+}
+
+func NewProtoNativeSchema(protoNativeSchemaDef string, properties map[string]string) *ProtoNativeSchema {
```
BewareMyPower, line 237: I think this function should not be exposed to users. It's hard for users to know which string should be passed to the 1st parameter. I suggest only exposing the method that accepts a `proto.Message`:

```go
func NewProtoNativeSchema(message proto.Message, properties map[string]string) *ProtoNativeSchema {
        return newProtoNativeSchema(GetProtoNativeSchemaInfo(message), properties)
}

func newProtoNativeSchema(protoNativeSchemaDef string, properties map[string]string) *ProtoNativeSchema {
```

BewareMyPower (changes requested):

pulsar/schema.go
BewareMyPower: The `GetProtoNativeSchemaInfo` and `GetFileProto` functions should not be exposed as well.

pulsar/schema.go
```diff
@@ -219,6 +226,107 @@ func (ps *ProtoSchema) GetSchemaInfo() *SchemaInfo {
...
+	RootMessageTypeName    string `json:"rootMessageTypeName"`
+	RootFileDescriptorName string `json:"rootFileDescriptorName"`
+}
+
+func GetFileProto(fileDesc protoreflect.FileDescriptor, protoMap map[string]*descriptorpb.FileDescriptorProto) {
+	if fileDesc.Imports().Len() > 0 {
+		for i := 0; i < fileDesc.Imports().Len(); i++ {
+			GetFileProto(fileDesc.Imports().Get(i).ParentFile(), protoMap)
+		}
+	}
```
BewareMyPower, line 298:
```suggestion
        for i := 0; i < fileDesc.Imports().Len(); i++ {
                GetFileProto(fileDesc.Imports().Get(i).ParentFile(), protoMap)
        }
```

Remove unnecessary check.

pulsar/schema.go
BewareMyPower: +1, I didn't get why we need to check it again. I think we only need to implement it like:

```go
func GetFileProto(fileDesc protoreflect.FileDescriptor, protoMap map[string]*descriptorpb.FileDescriptorProto) {
        for i := 0; i < fileDesc.Imports().Len(); i++ {
                GetFileProto(fileDesc.Imports().Get(i).ParentFile(), protoMap)
        }
        protoMap[fileDesc.Path()] = protodesc.ToFileDescriptorProto(fileDesc)
}
```

pulsar/schema.go
```diff
@@ -219,6 +226,107 @@ func (ps *ProtoSchema) GetSchemaInfo() *SchemaInfo {
...
+			PublicDependency: fileProtoMap[k].PublicDependency,
+			MessageType:      fileProtoMap[k].MessageType,
+			EnumType:         fileProtoMap[k].EnumType,
+			Service:          fileProtoMap[k].Service,
+			Extension:        fileProtoMap[k].Extension,
+			Options:          fileProtoMap[k].Options,
+			SourceCodeInfo:   fileProtoMap[k].SourceCodeInfo,
+			Syntax:           fileProtoMap[k].Syntax,
+		}
+		fileDescList = append(fileDescList, &value)
```
BewareMyPower, line 266:
```suggestion
        for _, v := range fileProtoMap {
                fileDescList = append(fileDescList, v)
```

Simplify the code. We don't need to copy the object that the pointer `v` references. Golang will handle the pointer well.

BewareMyPower (changes requested):

pulsar/schema.go
```diff
@@ -219,6 +226,107 @@ func (ps *ProtoSchema) GetSchemaInfo() *SchemaInfo {
...
+	}
+	fileDescSet := descriptorpb.FileDescriptorSet{
+		File: fileDescList,
+	}
+	bytesData, err := proto.Marshal(&fileDescSet)
+	if err != nil {
+		log.Fatalf("get serialized proto file data error: %v", err)
```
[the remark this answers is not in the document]
gaoran10: I found other schema-initialize functions also handle errors like this.

pulsar/schema.go
gaoran10: I think so, but I found the Java client also has this [check](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/ProtobufNativeSchemaUtils.java#L80), I'm not sure if there is any potential problem.

gaoran10: I think so, I just found the Java client also has this [check](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/schema/ProtobufNativeSchemaUtils.java#L80), I'm not sure if there is any potential problem. I remove it first.

pulsar/schema.go
gaoran10: I remove this check.

pulsar/schema.go
gaoran10: Fixed, thanks.

pulsar/schema.go
gaoran10: fixed

pulsar/schema.go
gaoran10: fixed

.gitignore
```diff
@@ -14,3 +14,4 @@ coverage.html
 perf/perf
 pulsar-perf
 bin
+integration-tests/pkg
```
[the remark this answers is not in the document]
gaoran10: I found this directory in my project, but I'm not sure which operation will create this directory, I remove this first.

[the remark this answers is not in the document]
BewareMyPower: I think this directory might be generated by a wrong `go` related operation. It should not be added to the `.gitignore` file.

pulsar/schema.go
```diff
@@ -219,6 +226,107 @@ func (ps *ProtoSchema) GetSchemaInfo() *SchemaInfo {
...
+	}
+	schemaData := ProtoNativeSchemaData{
+		FileDescriptorSet:      bytesData,
+		RootMessageTypeName:    string(message.ProtoReflect().Descriptor().FullName()),
+		RootFileDescriptorName: fileDesc.Path(),
+	}
+	jsonData, err := json.Marshal(schemaData)
+	if err != nil {
+		log.Fatalf("get json schema data for proto native schema error: %v", err)
```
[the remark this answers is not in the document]
gaoran10: fixed

.gitignore
```diff
@@ -13,4 +13,4 @@ coverage.html
 
 perf/perf
 pulsar-perf
-bin
+bin
```
BewareMyPower, line 16: Please do not remove the line break at the end of a file. If you're using the JetBrains' IDEs, you might do that by selecting **Editer - General - Ensure every saved file ends with a line break**.

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

gaoran10: Fixed, thanks.

BewareMyPower (approved):

BewareMyPower:
@nodece Could you take a look again?

nodece (approved):
LGTM

PR #1022, opened 2023-05-29
[Issue 1021][client] FIX: use maphash instead of crypto/sha256 for hash function of hashmap in Schema.hash()

### LINKED ISSUE #1021: Schema.hash is ressource intensive
#### Expected behavior

Usage of pulsar producer with Schema performs well or similar when producing without Schema.

#### Actual behavior
When a schema is enabled in the producer, the produce throughput is degraded.

I profiled the application and the result indicates that `Schema.hash` is using 30% CPU time.

https://github.com/apache/pulsar-client-go/commit/ea6eccf7ddad548a1ad977a0b72a61b894e6d248 introduced a function `Schema.hash` with `sha256`.
This might be not noticeable on low event throughput, but indicates a problem with high volume.
Producing 300k events/s is not a problem at all when no schema is used.

profile002

I can see at https://github.com/apache/pulsar-client-go/blob/27e36250f35094b203483c946bc4406c6119ce62/pulsar/producer_partition.go#L537 that there is some caching logic for schema, and `Schema.Get` uses the hash of the Schema for cache lookup.

May I suggest to use a more performant hash algorithm than sha256? Maybe there is also another possibility to achive the goal of looking up the schema cache in a more performant way.

I'm open to contribute and test different approaches.

#### Steps to reproduce

- Pulsar Producer with Schema
- Enable Profiling
- Produce 200k events/s

#### System configuration
**Pulsar version**: 2.11.1

bpereto:
use golang/hash/maphash instead of crypto/sha256 for hash function of hashmap in Schema.hash()
as the hash function is used for determining the hash key, it is safe to use a non-cryptographic function.

Fixes #1021

### Motivation

Improve performance of the SchemaCache by replacing the cryptographic hash function.

### Modifications

Replaced the crypto/sha256 with golang/hash/maphash function.

### Verifying this change

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -101,12 +101,12 @@ type partitionProducer struct {
 
 type schemaCache struct {
 	lock    sync.RWMutex
-	schemas map[string][]byte
+	schemas map[uint64][]byte
 }
 
 func newSchemaCache() *schemaCache {
 	return &schemaCache{
-		schemas: make(map[string][]byte),
+		schemas: make(map[uint64][]byte),
 	}
 }
 
@@ -122,9 +122,9 @@ func (s *schemaCache) Get(schema *SchemaInfo) (schemaVersion []byte) {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
 
-	key := schema.hash()
-	return s.schemas[key]
+	return s.schemas[schema.hash()]
 }
+
 func newPartitionProducer(client *client, topic string, options *ProducerOptions, partitionIdx int,
 	metrics *internal.LeveledMetrics) (
 	*partitionProducer, error) {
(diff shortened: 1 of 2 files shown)


kevinkl3 (approved):
LGTM

bpereto:
I saw the seed is random. so theres a method needed to fix the seed.

bpereto:
Added a static seed with `MakeSeed`. 
tested it with my application, works as intended for me.

tisonkun (approved):
Thank you!

To double-check the patch - it changes local access to the schema map, without any changes to the actual schema go through the wire, right?

bpereto:
> Thank you!
> 
> To double-check the patch - it changes local access to the schema map, without any changes to the actual schema go through the wire, right?

yes, no changes to the schema itself, just the hash method to get a key for the local hashmap struct to update or access the cached schema.

tisonkun:
Merging...

tisonkun:
Thank you!

PR #1033, opened 2023-06-25
[Improve] improve the perf of schema and schema cache

gunli:

Fixes #<xyz>

SchemaInfo.hash() and schemaCache.Get() are frequently called by partitionProducer.internalSend(), current implementation of these functions use sync.RWMutex to sync the cache map and calculate the hash every call, use the sync.Map and cache the hash value can improve the performance.

Master Issue: #<xyz>

### Motivation

### Modifications

- Modified: pulsar/producer_partition.go
- Modified: pulsar/schema.go

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)
  - **The schema: (yes)**
  - The wire protocol: ( no)

### Documentation


pulsar/schema.go
```diff
@@ -37,27 +38,27 @@ import (
 type SchemaType int
 
 const (
-	NONE        SchemaType = iota //No schema defined
```
shibd, line 40: Please do not submit format changes that are not related to this PR.

pulsar/schema.go
```diff
@@ -66,13 +67,21 @@ type SchemaInfo struct {
...
+	if oldHash != 0 {
+		return oldHash
+	}
+
 	h := maphash.Hash{}
 	h.SetSeed(seed)
 	h.Write([]byte(s.Schema))
-	return h.Sum64()
+	newHash := h.Sum64()
+	s.hashVal.CompareAndSwap(oldHash, newHash)
```
shibd, line 83: We can use `sync.Once` to ensure `hashVal` just set value on first call time.

``` go
	s.onc.Do(func() {
		h := maphash.Hash{}
		h.SetSeed(seed)
		h.Write([]byte(s.Schema))
		s.hashVal = h.Sum64()
	})
	return s.hashVal
```

shibd (commented):
Thanks for your contribution. Left some small comments.

pulsar/schema.go
gunli: OK, but I think it is better to keep a same comment format for a project, so I configured my IDE to run gofmt when I save the files, I have to update my IDE config when I prepare a PR, and revert the config when I do my own job, it not so convinient...

pulsar/schema.go
gunli: I think `atomic` is enough, 'cause in sync.Once, there is an `atomic` and a `mutex`, when we read the hash, it will `atomic.LoadUint32(&o.done)` first, the costs are the same.

``` go
type Once struct {
	// done indicates whether the action has been performed.
	// It is first in the struct because it is used in the hot path.
	// The hot path is inlined at every call site.
	// Placing done first allows more compact instructions on some architectures (amd64/386),
	// and fewer instructions (to calculate offset) on other architectures.
	done uint32
	m    Mutex
}

func (o *Once) Do(f func()) {
	// Note: Here is an incorrect implementation of Do:
	//
	//	if atomic.CompareAndSwapUint32(&o.done, 0, 1) {
	//		f()
	//	}
	//
	// Do guarantees that when it returns, f has finished.
	// This implementation would not implement that guarantee:
	// given two simultaneous calls, the winner of the cas would
	// call f, and the second would return immediately, without
	// waiting for the first's call to f to complete.
	// This is why the slow path falls back to a mutex, and why
	// the atomic.StoreUint32 must be delayed until after f returns.

	if atomic.LoadUint32(&o.done) == 0 {
		// Outlined slow-path to allow inlining of the fast-path.
		o.doSlow(f)
	}
}
```

pulsar/schema.go
gunli: Hi shibd, I have revert the comment

pulsar/schema.go
shibd: But, Using `sync.once` is more intuitive.

And, In multi-thread, your implementation of this is not guaranteed to be hashed only once.

pulsar/schema.go
shibd: Yes, If there is a good way to automate, you can contribute to the code format alone,

RobertIndie (approved):

pulsar/schema.go
gunli: @shibd I have update to sync.Once since there is no atomic.Uint64 in go 1.18

shibd (approved):

PR #1029, opened 2023-06-12
[Issue 1027][producer] fix: split sendRequest and make reconnectToBroker and other operate in the same coroutine

zengguan:
Fixes #1027

### Motivation

fix send timeout error cause by reconnection failures

### Modifications

After #691 , client avoid blocking the reconnection logic, but it also make reconnection and internalSend/internalFlush have concurrency issues。

when the client connection is block, it run into `p.grabCnx()` to get new connection and resend pendingQueue messages. At the same time, the internalSend/internalFlush keep running by use pendingQueue and  send mssage by `p._getConn().WriteData`. Therefore, it received an ack larger than expected, and trigger the ConnectionClosed logic, then the client enter `p.reconnectToBroker()` and repeat the concurrency issues with internalSend/internalFlush, fall into a reconnect-resend-connectionClosed loop

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API: (no)
  - The schema: (no)
  - The default values of configurations: (no)
  - The wire protocol: (no)

### Documentation

  - Does this pull request introduce a new feature? (no)
  - If yes, how is the feature documented? (not applicable / docs / GoDocs / not documented)
  - If a feature is not applicable for documentation, explain why?
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

PR #1058, opened 2023-07-18
[fix] [issue 1057]:  Fix the producer flush opertion is not guarantee to flush all messages

### LINKED ISSUE #1057: [BUG] The producer flush opertion is not guarantee to flush all messages
#### Expected behavior

The `producer.Flush` should guarantee that all the messages should be flushed to the broker.

#### Actual behavior

The flush operation may skip some messages in the producer.

#### Steps to reproduce

```go
		Payload: []byte("hello"),
	}, nil)
err = producer.Flush()
if err != nil {
	return
}
```
Currently, the Flush cannot guarantee that the message `hello` is actually flushed to the broker.

The root cause is that when calling the SendAsync, the producer will create another go routine to put the message to the queue.
Therefore, when flush is called, the message may not have entered the queue yet.

#### System configuration
**Pulsar Go version**: master

Gleiphir2769:
Fixes #1057 

### Motivation

`dataChan` is introduced by #1029  to fix the problem of reconnectToBroker. But it missed that if a flush operation excuted, there may still be some messages in `dataChan`. And these messages can't be flushed.

### Modifications

- Fix the producer flush opertion is not guarantee to flush all messages

### Verifying this change

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -1041,11 +1041,13 @@ func (p *partitionProducer) internalFlushCurrentBatches() {
 }
 
 func (p *partitionProducer) internalFlush(fr *flushRequest) {
-	oldDataChan := p.dataChan
-	p.dataChan = make(chan *sendRequest, p.options.MaxPendingMessages)
-	for len(oldDataChan) != 0 {
-		pendingData := <-oldDataChan
-		p.internalSend(pendingData)
+	if len(p.dataChan) != 0 {
+		oldDataChan := p.dataChan
+		p.dataChan = make(chan *sendRequest, p.options.MaxPendingMessages)
+		for len(oldDataChan) != 0 {
+			pendingData := <-oldDataChan
+			p.internalSend(pendingData)
+		}
 	}
 
 	p.internalFlushCurrentBatch()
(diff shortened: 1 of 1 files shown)


pulsar/producer_partition.go
```diff
@@ -1041,6 +1041,12 @@ func (p *partitionProducer) internalFlushCurrentBatches() {
 }
 
 func (p *partitionProducer) internalFlush(fr *flushRequest) {
+	oldDataChan := p.dataChan
+	p.dataChan = make(chan *sendRequest, p.options.MaxPendingMessages)
+	for len(oldDataChan) != 0 {
+		pendingData := <-oldDataChan
+		p.internalSend(pendingData)
+	}
```
RobertIndie, line 1049:
```suggestion
	if len(p.dataChan) != 0 {
		oldDataChan := p.dataChan
		p.dataChan = make(chan *sendRequest, p.options.MaxPendingMessages)
		for len(oldDataChan) != 0 {
			pendingData := <-oldDataChan
			p.internalSend(pendingData)
		}
	}
```

Gleiphir2769: You r right. We should check length of dataChan first.

RobertIndie (approved):

PR #1059, opened 2023-07-18
[Fix][producer] Fail all messages that are pending requests when closing like java

### LINKED ISSUE #1042: [Bug][Producer]Forget to consume the sendRequests before closing the producer
#### Expected behavior

Consume the sendRequests before closing the producer, send and flush them, invoke the callbacks of the input messages, so that the application can know the producing is succeed or failed.

#### Actual behavior

Currently, when we close the producer, we forget to consume the sendRequests in partitionProducer.dataChan before closing.
In the case that producing is faster than consuming in partitionProducer, we have send a lot of sendRequests  into partitionProducer.dataChan, when closing, many of them are not consumed, these sendRequests will looks like get lost, their callback won't get invoked, the application won't know the producing result of these messages.

#### Steps to reproduce

Review the code of partitionProducer
``` go
func (p *partitionProducer) internalSendAsync(ctx context.Context, msg *ProducerMessage,
	callback func(MessageID, *ProducerMessage, error), flushImmediately bool) {
	//Register transaction operation to transaction and the transaction coordinator.
	var newCallback func(MessageID, *ProducerMessage, error)
	if msg.Transaction != nil {
		transactionImpl := (msg.Transaction).(*transaction)
		if transactionImpl.state != TxnOpen {
			p.log.WithField("state", transactionImpl.state).Error("Failed to send message" +
				" by a non-open transaction.")
			callback(nil, msg, newError(InvalidStatus, "Failed to send message by a non-open transaction."))
			return
		}

		if err := transactionImpl.registerProducerTopic(p.topic); err != nil {
			callback(nil, msg, err)
			return
		}
		if err := transactionImpl.registerSendOrAckOp(); err != nil {
			callback(nil, msg, err)
		}
		newCallback = func(id MessageID, producerMessage *ProducerMessage, err error) {
			callback(id, producerMessage, err)
			transactionImpl.endSendOrAckOp(err)
		}
	} else {
		newCallback = callback
	}
		// Producer is closing
		newCallback(nil, msg, errProducerClosed)
		return
	}

	// bc only works when DisableBlockIfQueueFull is false
	bc := make(chan struct{})

	// callbackOnce make sure the callback is only invoked once in chunking
	callbackOnce := &sync.Once{}
	var txn *transaction
	if msg.Transaction != nil {
		txn = (msg.Transaction).(*transaction)
	}
	sr := &sendRequest{
		ctx:              ctx,
		msg:              msg,
		callback:         newCallback,
		callbackOnce:     callbackOnce,
		flushImmediately: flushImmediately,
		publishTime:      time.Now(),
		blockCh:          bc,
		closeBlockChOnce: &sync.Once{},
		transaction:      txn,
	}
	p.options.Interceptors.BeforeSend(p, msg)

	p.dataChan <- sr

	if !p.options.DisableBlockIfQueueFull {
		// block if queue full
		<-bc
	}
}

func (p *partitionProducer) internalClose(req *closeProducer) {
	defer close(req.doneCh)
	if !p.casProducerState(producerReady, producerClosing) {
		return
	}

	p.log.Info("Closing producer")

	id := p.client.rpcClient.NewRequestID()
	_, err := p.client.rpcClient.RequestOnCnx(p._getConn(), id, pb.BaseCommand_CLOSE_PRODUCER, &pb.CommandCloseProducer{
		ProducerId: &p.producerID,
		RequestId:  &id,
	})

	if err != nil {
		p.log.WithError(err).Warn("Failed to close producer")
	} else {
		p.log.Info("Closed producer")
	}

	if p.batchBuilder != nil {
		if err = p.batchBuilder.Close(); err != nil {
			p.log.WithError(err).Warn("Failed to close batch builder")
		}
	}

	p.setProducerState(producerClosed)
	p._getConn().UnregisterListener(p.producerID)
	p.batchFlushTicker.Stop()
}
```

And in the interface of the Producer, we saied **Waits until all pending write request are persisted. In case of errors, pending writes will not be retried.**, I think the messages that have been summit to `SendAsync` should be treated as **pending write requests**, so we should send and flush them to make sure they are done with a result.

``` go
// Producer is used to publish messages on a topic
type Producer interface {
	...

	// Close the producer and releases resources allocated
	// No more writes will be accepted from this producer. Waits until all pending write request are persisted. In case
	// of errors, pending writes will not be retried.
	Close()
}
```
#### System configuration
**Pulsar version**: x.y

@zengguan @merlimat @wolfstudy

### LINKED ISSUE #1057: [BUG] The producer flush opertion is not guarantee to flush all messages

graysonzeng:
Fixes #1042 , link #1057

### Motivation

producer not fail all messages that are in the pending queue when closing

### Modifications

add `failPendingMessages()` to fail the pending requests when closing producer

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)


Gleiphir2769:
Hi @graysonzeng. I think dataChan should be cleared in the `internalFlush`. Because `internalFlush` should guarantee flush all the messages in the dataChan too.

I submit a fix PR in #1058 and I think `internalClose` can just invoke `internalFlush` to clear `dataChan`.

gunli:
> I submit a fix PR in #1058 and I think `internalClose` can just invoke `internalFlush` to clear `dataChan`.

I think that will be better, I think we can update this PR after @Gleiphir2769's PR is merged. @graysonzeng

graysonzeng:
> Hi @graysonzeng. I think dataChan should be cleared in the `internalFlush`. Because `internalFlush` should guarantee flush all the messages in the dataChan too.
> 
> I submit a fix PR in #1058 and I think `internalClose` can just invoke `internalFlush` to clear `dataChan`.

I agree. I will update this PR after #1057  is merged

RobertIndie:
Hi @graysonzeng #1058 has been merged.

graysonzeng:
update completed @RobertIndie

RobertIndie:
Seems that in the Java client, the producer will fail all messages that are in the pending queue when closing:
https://github.com/apache/pulsar/blob/2bede012c73c301b73c079aaeb122cbb472728c1/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L1079
https://github.com/apache/pulsar/blob/2bede012c73c301b73c079aaeb122cbb472728c1/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L1098
```java
cnx.sendRequestWithId(cmd, requestId).handle((v, exception) -> {
            cnx.removeProducer(producerId);
            if (exception == null || !cnx.ctx().channel().isActive()) {
                // Either we've received the success response for the close producer command from the broker, or the
                // connection did break in the meantime. In any case, the producer is gone.
                log.info("[{}] [{}] Closed Producer", topic, producerName);
                closeAndClearPendingMessages();  // <- we call failPendingMessages in closeAndClearPendingMessages
                closeFuture.complete(null);
            } else {
                closeFuture.completeExceptionally(exception);
            }

            return null;
        });
```

gunli:
> Seems that in the Java client, the producer will fail all messages that are in the pending queue when closing: 

From the user's perspective, I would like to have the submitted data sent out as much as possible when closing the client. The cost is I have to wait until timeout.

RobertIndie:
> From the user's perspective, I would like to have the submitted data sent out as much as possible when closing the client. The cost is I have to wait until timeout. 

The user can call the `flush` and then call the `close`. IMO, the `close` means that we are going to terminate the producer and release all the resources. If there are any ongoing requests, we need to fail them.

gunli:
> The user can call the `flush` and then call the `close`. IMO, the `close` means that we are going to terminate the producer and release all the resources. If there are any ongoing requests, we need to fail them.

Hmm, this implies that users must call flush() before closing, and if they do so but still have unsent data, it looks like that the flush() implementation is not perfect. On the other hand, if users don't call flush(), they will have to bear the consequences themselves, in this case, it seems better to encapsulate flush() in close() to reduce the burden on users.

 It seems a bit contradictory.

graysonzeng:
https://github.com/apache/kafka/blob/4ea9394e7ebf71e2dcf96e9b96191dac253f4930/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L1294C15-L1294C22

In kafka. When the user actively closes, the waiting time can be specified until all messages in the queue are cleared. I think it is a good choice to use flush to send the data in the remaining queue before closing。

If you think it's better to let the message fail directly. i can change it

RobertIndie:
> and if they do so but still have unsent data, it looks like that the flush() implementation is not perfect.

The flush should guarantee that messages which are at least pushed into the data chan should be flushed to the broker. And that's the issue @Gleiphir2769 has fixed:https://github.com/apache/pulsar-client-go/issues/1057.

>  if users don't call flush(), they will have to bear the consequences themselves

Yes. That's right. The user can choose to fail or flush the messages.

IMO, the close should have only one responsibility: just close and release all the resources of the producer.
 I still think it's more important to keep the behavior consistent with the Java client. If we want to change this behavior, then we need to change it in the Java client first. And this needs a discussion on the mailing list. Besides, I still think it's better to fail the messages directly.

gunli:
> And, as a user, when it reach the time to call close(), I won't care about the results of the async callback of the failed requests, unless the close() returns the failed requests to me as the return value, in this case, it seems failing the pending reqeusts is of little meaning.

OK. But I think we should think about the concerns aboved for the users, return the failed requests to the users rather than just calling the async callbacks, so that they can decide what to do to these failed requests.

RobertIndie:
> OK. But I think we should think about the concerns aboved for the users, return the failed requests to the users rather than just calling the async callbacks, so that they can decide what to do to these failed requests.

@gunli Thanks. You're right. I have created an issue to track this improvement: https://github.com/apache/pulsar-client-go/issues/1063

RobertIndie (changes requested):
We should fail the pending requests in the pending queue in the producer instead of the connection. Just like `failTimeoutMessages` did:
https://github.com/apache/pulsar-client-go/blob/e45122c2defc5efd4efc493d0acef278a7ccfc01/pulsar/producer_partition.go#L895

RobertIndie (commented):
Overall looks good to me. Could you add a test?

graysonzeng:
HI @RobertIndie . How can I stably construct some messages that exist in pendQueue ?

RobertIndie:
> HI @RobertIndie . How can I stably make some messages that exist in pendQueue ?

You can create a producer with relative large BatchingMaxPublishDelay and BatchingMaxMessages so that messages could stay in the pending queue stably.

graysonzeng:
PTAL @RobertIndie

tisonkun (approved):
Thank you!

PR #1055, opened 2023-07-11
[fix] [issue 1051]: Fix inaccurate producer mem limit in chunking and schema

Gleiphir2769:
Fixes #1051 

### Motivation

The producer memory limit have some problem when `EnableChunking=true` or `Schema` is set.
- When `Schema` is set, the actual message payload is `msg.Value`. The `len(msg.Payload)` may be 0 and memory can not be reserved acurate.
https://github.com/apache/pulsar-client-go/blob/be3574019383ac0cdc65fec63e422fcfd6c82e4b/pulsar/producer_partition.go#L479-L494

- In chunking, if producer meets the memory limit, it should release the memory for **chunks which has send out**. But the calculate for this release is not accurate, it should be `uncompressedPayloadSize - int64(lhs)` instead of `uncompressedPayloadSize - int64(rhs)`
https://github.com/apache/pulsar-client-go/blob/be3574019383ac0cdc65fec63e422fcfd6c82e4b/pulsar/producer_partition.go#L662-L664

- In chunking, if `internalSingleSend` is failed, it should release the memory for **single chunk**. But we release all the chunks memory repeatly now.
https://github.com/apache/pulsar-client-go/blob/be3574019383ac0cdc65fec63e422fcfd6c82e4b/pulsar/producer_partition.go#L838-L843

- When producer received the receipt from broker, it should release the memory **it reserved before sending**. But it releases wrong size in `chunking` and `schema`.
https://github.com/apache/pulsar-client-go/blob/be3574019383ac0cdc65fec63e422fcfd6c82e4b/pulsar/producer_partition.go#L1221-L1230

### Modifications

- Fix all the memory limit problems relative to `chunking` and `schema`
- Add unit tests to cover these scenarios

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (yes / no)
  - The public API: (yes / no)
  - The schema: (yes / no / don't know)
  - The default values of configurations: (yes / no)
  - The wire protocol: (yes / no)

### Documentation

  - Does this pull request introduce a new feature? (yes / no)
  - If yes, how is the feature documented? (not applicable / docs / GoDocs / not documented)
  - If a feature is not applicable for documentation, explain why?
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

PR #1068, opened 2023-07-21
[fix] [issue 1067] Fix the excessive dataChan capacity

### LINKED ISSUE #1067: [BUG] The capacity of dataChan is too big
Hi community. I noticed that the capacity of dataChan is too big. We make the `dataChan` with capacity `MaxPendingMessages`  and the account of pending messages actual in queue may exceed the limit by `MaxPendingMessages` a lot. 

This is because we use `publishSemaphore` instead of capacity of `dataChan` to limit pending messages. The actual limit is 2X the `MaxPendingMessages`.

https://github.com/apache/pulsar-client-go/blob/9867c29ca329302e97ddd9c6a99f66853c7f447f/pulsar/producer_partition.go#L148

Gleiphir2769:
Fixes #1067 

### Motivation

wait to edit...

### Modifications

### Verifying this change

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -145,7 +145,7 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		log:              logger,
 		options:          options,
 		producerID:       client.rpcClient.NewProducerID(),
-		dataChan:         make(chan *sendRequest, maxPendingMessages),
+		dataChan:         make(chan *sendRequest, 10),
 		cmdChan:          make(chan interface{}, 10),
 		connectClosedCh:  make(chan connectionClosed, 10),
 		batchFlushTicker: time.NewTicker(batchingMaxPublishDelay),


gunli:
Hi @Gleiphir2769, as discussed in  #1055 , we will move some logic from `internalSend()` back to `internalAsyncSend()`, reserving resource for the requset will be done before putting it into dataChan, if there is enough semaphore and pending queue spaces, continue, or it will be block or failed, as the JAVA client did, and this will be not a problem any more.

Gleiphir2769:
> we will move some logic from internalSend() back to internalAsyncSend(), reserving resource for the requset will be done before putting it into dataChan

If we have made some progress in this work, this PR can be closed. But I think we should fix the existing bugs first and this fix has no side effects on this work. What do you think? @gunli

gunli:
> If we have made some progress in this work, this PR can be closed. But I think we should fix the existing bugs first and this fix has no side effects on this work. What do you think? @gunli

@Gleiphir2769 I will do the refactoring work this week.

Gleiphir2769:
> I will do the refactoring work this week.
Good! I will close this PR after your refactor work.

gunli:
@Gleiphir2769 I have pushed the PR #1071 , please help to review it, thanks!

tisonkun:
Closing...

You may review #1121 and I still think that we can have a much better solution than even after #1121 merged.

Issue #1060, opened 2023-07-19
[Bug][Producer]Inaccurate transaction endSendOrAckOp for chunked message


gunli:
#### Expected behavior

When chunking, call `transaction.endSendOrAckOp()` only for the last chunked reqeust???

#### Actual behavior
I am NOT sure if it is a bug.

In `internalSend()`, when chunking, we share the `transaction` between all the chunked request:
```go
nsr := &sendRequest{
	ctx:              request.ctx,
	msg:              request.msg,
	callback:         request.callback,
	callbackOnce:     request.callbackOnce,
	publishTime:      request.publishTime,
	blockCh:          request.blockCh,
	closeBlockChOnce: request.closeBlockChOnce,
	totalChunks:      totalChunks,
	chunkID:          chunkID,
	uuid:             uuid,
	chunkRecorder:    cr,
	transaction:      request.transaction,
}
```

In `ReceivedSendReceipt()`, we call `transaction.endSendOrAckOp()` for every request without checking if it is the last chunked one
```
if sr.transaction != nil {
	sr.transaction.endSendOrAckOp(nil)
}
```

In `failTimeoutMessages()`, we call `transaction.endSendOrAckOp()` for every request without checking if it is the last chunked one
```
if sr.transaction != nil {
	sr.transaction.endSendOrAckOp(nil) // And I think it should be errSendTimeout here
}
```
#### Steps to reproduce

Code review.

#### System configuration
**Pulsar version**: master

@RobertIndie @Gleiphir2769

RobertIndie:
Could you take a look at this? @liangyepianzhou

liangyepianzhou:
@gunli You are right. Thanks for you point it out.

liangyepianzhou:
>In failTimeoutMessages(), we call transaction.endSendOrAckOp() for every request without checking if it is the last >chunked one
>```
>if sr.transaction != nil {
>	sr.transaction.endSendOrAckOp(nil) // And I think it should be errSendTimeout here
>}
>```

The errSendTimeout already was passed into the callback just before `sr.transaction.endSendOrAckOp(nil)`.
``` go
if sr.callback != nil {
	sr.callbackOnce.Do(func() {
		sr.callback(nil, sr.msg, errSendTimeout)
	})
}
if (sr.totalChunks <= 1 || sr.chunkID == sr.totalChunks-1) && sr.transaction != nil {
	sr.transaction.endSendOrAckOp(nil)
}
```

PR #1069, opened 2023-07-22
[fix][txn] Fix the transaction acknowledgement and send logic for  chunk message

liangyepianzhou:
Master https://github.com/apache/pulsar-client-go/issues/1060
### Motivation
1. For the chunk message, we only register the send operation once but end the send operation multiple times when receiving the send response. It will make the transaction can be committed before all the operations are completed.
2. When we use transaction ack for chunk messages, the provided transaction is ignored, resulting in the chunk message actually being acknowledged using the non-transactional ack method.
### Modifications
1. Only end the send operation when receive the last chunk message.
2. Add the check for the transaction when the massage is a chunk message.
### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -979,13 +979,10 @@ func (p *partitionProducer) failTimeoutMessages() {
 						WithField("properties", sr.msg.Properties)
 				}
 
-				if sr.callback != nil {
+				if sr.totalChunks <= 1 || sr.chunkID == sr.totalChunks-1 {
 					sr.callbackOnce.Do(func() {
 						sr.callback(nil, sr.msg, errSendTimeout)
 					})
-				}
-				if (sr.totalChunks <= 1 || sr.chunkID == sr.totalChunks-1) && sr.transaction != nil {
-					sr.transaction.endSendOrAckOp(nil)
 				}
 			}
 
@@ -1253,9 +1250,6 @@ func (p *partitionProducer) ReceivedSendReceipt(response *pb.CommandSendReceipt)
 						sr.callback(msgID, sr.msg, nil)
 					}
 					p.options.Interceptors.OnSendAcknowledgement(p, sr.msg, msgID)
-					if sr.transaction != nil {
-						sr.transaction.endSendOrAckOp(nil)
-					}
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


pulsar/producer_partition.go
```diff
@@ -984,7 +984,7 @@ func (p *partitionProducer) failTimeoutMessages() {
 						sr.callback(nil, sr.msg, errSendTimeout)
 					})
 				}
-				if sr.transaction != nil {
+				if (sr.totalChunks <= 1 || sr.chunkID == sr.totalChunks-1) && sr.transaction != nil {
 					sr.transaction.endSendOrAckOp(nil)
```
gunli, line 988: I think it should passed `errSendTimeout` to `endSendOrAckOp` to update the transcaction to `TxnError`
```go
func (txn *transaction) endSendOrAckOp(err error) {
	if err != nil {
		atomic.StoreInt32((*int32)(&txn.state), int32(TxnError))
	}
	if atomic.AddInt32(&txn.opsCount, -1) == 0 {
		//This means there are not pending send/ack requests
		txn.opsFlow <- true
	}
}
```

pulsar/producer_partition.go
```diff
@@ -979,14 +979,11 @@ func (p *partitionProducer) failTimeoutMessages() {
 						WithField("properties", sr.msg.Properties)
 				}
 
-				if sr.callback != nil {
+				if sr.totalChunks <= 1 || sr.chunkID == sr.totalChunks-1 {
 					sr.callbackOnce.Do(func() {
 						sr.callback(nil, sr.msg, errSendTimeout)
```
gunli, line 984: run callback without checking if it is nil

gunli:
@liangyepianzhou The bug of producer I have fixed in #1071

liangyepianzhou:
> @liangyepianzhou The bug of producer I have fixed in #1071

Maybe you can merge the master after this PR is merged?

gunli:
Or you can just fix the consumer, I have refactored the producer, that modified a lot of code.

gunli:
You are welcome. So, if you are free, would you please help to review my PR #1071.

RobertIndie (approved):

PR #819, opened 2022-07-29
[go]: reduce dependency

ttys3:
### Motivation

currently we have `github.com/spf13/cobra v1.2` dependency, which depends on `spf13/viper` (which is a heavy mod)

the good news is that github.com/spf13/cobra version >= v1.4.0 has `viper` mod dependency removed.

yes, but we can still avoid introduce unnecessary dependency to SDK client users.

### Modifications

make the `pulsar-perf` binary a standalone mod, so that its `cobra` dependency does not affect pulsar-client-go

nothing is break, but a more clean go mod

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

This change is a trivial rework / code cleanup without any test coverage.

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (yes)

Dependencies removed from module `github.com/apache/pulsar-client-go`:

```go
github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b

github.com/spf13/cobra v1.2.1
```

new module `github.com/apache/pulsar-client-go/pulsar-perf` introduced (which is not a package, but binary only package)

old `github.com/spf13/cobra v1.2.1` updated to `github.com/spf13/cobra v1.5.0` since this version has `github.com/spf13/viper` removed  (which does not affect the feature of cobra, see https://github.com/spf13/cobra/releases/tag/v1.4.0 )

```go
	github.com/apache/pulsar-client-go v0.8.1
	github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b
	github.com/prometheus/client_golang v1.12.2
	github.com/sirupsen/logrus v1.9.0
	github.com/spf13/cobra v1.5.0
```

  - The public API: (no)
  - The schema: (no)
  - The default values of configurations: (no)
  - The wire protocol: (no)

### Documentation

  - Does this pull request introduce a new feature? (no)

PR #1076, opened 2023-08-13
Feat: Integration teting with testcontainers go

flowchartsman:
### Motivation

Instead of building a customized container, test using base image and testcontainers-go. This enables manual individual test execution, and will only create the integration container if tests are run which require it.

### Modifications

- Integration container singleton using testcontainers-go
- Modify helper URLs to pull from mapped ports on integration container, ensuring that only tests which access these URLS will cause the integration container to get created.

### Verifying this change

This change is already covered by existing integration tests.

  - Dependencies (does it add or upgrade a dependency): yes
    - docker library dependencies have been updated to avoid the `http: invalid Host header` error (moby/moby#45935)
  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

    - this is a test-only feature

### WIP

Some connection tests are failing, such as the "retry on multipl hostname" or producer connection tests, however it's unclear how these changes would suddenly cause this test to fail, so it's possible these are flakey.


flowchartsman:
Some other benefits:

- no need for scripts in the happy pathway; the container is automatically created if needed and torn down after testing is complete
- code doesn't need to be copied into the container and built for every change, and with a little modification, the container could be made to persist by disabling the reaper container through the use of an environment variable or even a `go run` style container invocation allowing fast iteration on a single test or feature.
- the integration container and associated helper functions could be moved into another package, or even another module entirely, which would allow projects based on `pulsar-client-go` to use the same testing infrastructure.  In particular, I'm thinking about #819, which would split `pusar-perf` into its own repo. Even though this particular code path does not do IT, it potentially could do so by using the shared IT container code from the main repo.

flowchartsman:
Pending the changes in #1077, this PR might end up getting reworked such that the following code in `helper_for_test.go`

```go
func serviceURL() string {
	return integration.URL("pulsar", "6650")
}

func serviceURLTLS() string {
	return integration.URL("pulsar+ssl", "6651")
}

func webServiceURL() string {
	return integration.URL("http", "8080")
}

func webServiceURLTLS() string {
	return integration.URL("https", "8443")
}
```

Becomes

```go
AdminClient() *pulsaradmin.Client {
        return pulsaradmin.NewClient(pulsaradmin.ClientConfig{
		WebServiceURL:   integration.URL("https", "8443"),
		BKWebServiceURL: integration.URL("pulsar", "6650"),
                //etc
	})
}

AdminClientTLS() *pulsaradmin.Client {
        return pulsaradmin.NewClient(pulsaradmin.ClientConfig{
		WebServiceURL:   integration.URL("http", "8080"),
		BKWebServiceURL: integration.URL("pulsar+ssl", "6651"),
                //etc
	})
}
```

Which would then create the integration container only when one of the helper admin clients is used, while simultaneously dogfooding `pulsar/pulsaradmin` and increasing coverage. Which would be pretty darn slick.

pulsar/helper_for_test.go
```diff
@@ -29,24 +29,49 @@ import (
...
+
 	"github.com/stretchr/testify/assert"
 
 	pkgerrors "github.com/pkg/errors"
 )
 
 const (
-	serviceURL    = "pulsar://localhost:6650"
-	serviceURLTLS = "pulsar+ssl://localhost:6651"
+// serviceURL    = "pulsar://localhost:6650"
```
merlimat, line 40: Let's remove these instead of commenting out

Issue #1075, opened 2023-08-12
Tracking issue for merging pulsar-admin-go


tisonkun:
## Background

* Vote thread - https://lists.apache.org/thread/txpxdkzf4n5dd6bm0goqxfycr69g9gc4
* IP clearance - https://lists.apache.org/thread/85hynjryvr2869v9wtrqb8ym1853vx88

tisonkun:
@flowchartsman I outline the subtasks above and will start roughly adding source code the next week. You may propose the refactor ideas and I can help in review.

cc @RobertIndie @shibd @nodece @wolfstudy

flowchartsman:
Thanks @tFor those interested, I have an

tisonkun:
Thanks for your feedback and participation @flowchartsman!

I've submitted a PR https://github.com/apache/pulsar-client-go/pull/1077 for roughly adding the sources of `pulsar-admin-go`, so that we have a baseline to improve and the downstream projects can first migrate to that hash for updating module/pkg path, and later catch up any refactors.

Your refactoring suggestions can be applied onto that patch.

> [package-level error interfaces along with helper methods](https://github.com/streamnative/pulsar-admin-go/pull/41/files#diff-76a7e0e299c7417bae32d3098e71370980157fe29e68a366671491d147d40572)

I like this.

> exported Admin field

As we discussed on Slack, perhaps a parallel interface design for client API and admin API can be better. This is the design of Java APIs and it reflects that admin APIs (control panel) are at the same level of client API (data panel) - client APIs don't "own" admin APIs. Also, this can make the auth model clear - generally, users configure less permission for client APIs while admin APIs requires high trust level.

> Because pulsar-client-go already handles (most) auth plugins natively, the functional auth API I designed for my PR would no longer be necessary, and authentication would come from the auth settings specified in pulsar.ClientConfig

This is the right direction. We will factor out shared/common configs and the auth package. But how to reduce the impact for current users is a point that needs a second consideration.

> exercise most if not all endpoints using an actual pulsar container 

This is cool! We definitely need it.

PR #1077, opened 2023-08-14
Adopt pulsar admin go

tisonkun:
This refers to https://github.com/apache/pulsar-client-go/issues/1075.

Roughly add source code with ASF header without any nontrivial changes.

cc @flowchartsman @Gleiphir2769 @gunli

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,4 +1,3 @@
-#
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
@@ -15,7 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-#
 
 github:
   description: "Apache Pulsar Go Client Library"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - run: make lint
+      - name: Check license header
+        run: docker run --rm -v $(pwd):/github/workspace ghcr.io/korandoru/hawkeye-native:v3 check
+      - name: Run golangci-lint
+        run: make lint
 
   integration-tests:
     runs-on: ubuntu-latest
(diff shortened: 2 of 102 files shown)


flowchartsman:
As an administrative note now that we have an actual PR, I am prepped to replay the changes from streamnative/pulsar-admin-go#41 on top of these changes to refactor the API, either as a PR on [this branch](tisonkun:adopt-pulsar-admin-go) branch, against master, or as a separate PR, whichever is most appropriate. See #1075 for context. I would then begin the work of merging auth functionality and creating an admin client type.  This would remove the `pulsaradmin.Client` alias, replacing it with either an interface or a concrete type which would return concrete types representing the different administrative APIs currently modeled as interfaces (see [`FunctionsWorker` here](https://github.com/apache/pulsar-client-go/pull/1077/files#diff-77a7be6f50bcdbe42702153ff4b70c7cc136540832f57ed44c6b503780a25327R24-R39))

Thus,
```go
type Client interface {
	Functions() Functions
        // ...
}
```

would become:
```go
type Client interface {
	Functions() *FunctionsWorker
        // ...
}

// or

type Client struct{
       // internal fields
       Functions *FunctionsWorker
       //etc
}
```

I lean towards the concrete type if I'm being honest, since the construction of calling `client.Functions().Somemethod()` is awkward and making fat interfaces and `_impl` types is a bit of a Go antipattern, IMHO. The concrete client type can be tested using integration testing, since it consists almost entirely of HTTP calls, and users that wish to mock functionality in their code can simply make their own, smaller interfaces using just the functionality they need, such as:

```go
type MyFunctionMaintainer struct {
        functionAPI functionAPI
}

type functionAPI interface {
        CreateFuncWithURL(data *utils.FunctionConfig, pkgURL string) error
	GetFunction(tenant, namespace, name string) (utils.FunctionConfig, error)
        UpdateFunction(functionConfig *utils.FunctionConfig, fileName string, updateOptions *utils.UpdateOptions) error
	DeleteFunction(tenant, namespace, name string) error
}
```

A further PR could integrate the admin client into the normal pulsar tests, providing a single place to access the integration container, removing some of the awkward constructions used in #1076 and using the test admin client as the single point of access to the integration container. Which would be awesome :)

tisonkun:
> Can this PR include the pulsar-go-admin commit?

That's possible. Let me prepare a patch for this requirement. We should be able to use git filter-repo/filter-branch for such case.

tisonkun:
I'd minimize the changeset to replace the license checker in this PR and prepare a new one to do the filter-branch work which can be much more complicated.

PR #1087, opened 2023-09-01
[improve][client] Implement GetLastMSgID for Reader

liangyepianzhou:
### Motivation
Implement the GetLastMSgID API for Reader.

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/reader.go
+++ b/pulsar/reader.go
@@ -136,5 +136,5 @@ type Reader interface {
 	SeekByTime(time time.Time) error
 
 	// GetLastMSgID get the last message id available for consume.
-	GetLastMSgID() (MessageID, error)
+	GetLastMessageID() (MessageID, error)
 }
--- a/pulsar/reader.go
+++ b/pulsar/reader.go
@@ -135,6 +135,6 @@ type Reader interface {
 	//
 	SeekByTime(time time.Time) error
 
-	// GetLastMSgID get the last message id available for consume.
+	// GetLastMessageID get the last message id available for consume.
 	GetLastMessageID() (MessageID, error)
 }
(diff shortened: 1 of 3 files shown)


liangyepianzhou:
Leave a question: Do we have a plan to implement a multiple-topic-reader?

pulsar/reader.go
```diff
@@ -134,4 +134,7 @@ type Reader interface {
 	//            the message publish time where to reposition the subscription
 	//
 	SeekByTime(time time.Time) error
+
+	// GetLastMSgID get the last message id available for consume.
+	GetLastMSgID() (MessageID, error)
```
BewareMyPower, line 139: Please use a correct name like `GetLastMsgID`.

BewareMyPower (changes requested):

pulsar/reader.go
BewareMyPower: To keep the consistency with other clients, maybe `GetLastMessageID` will be the best name.

BewareMyPower:
> Do we have a plan to implement a multiple-topic reader?

Not yet. But I think we can do that.

In addition, please add tests for the exposed API.

pulsar/reader.go
```diff
@@ -134,4 +134,7 @@ type Reader interface {
 	//            the message publish time where to reposition the subscription
 	//
 	SeekByTime(time time.Time) error
+
+	// GetLastMSgID get the last message id available for consume.
```
BewareMyPower, line 138:
```suggestion
	// GetLastMessageID get the last message id available for consume.
```

BewareMyPower (approved):

liangyepianzhou:
@BewareMyPower Do you know which version of the go-client release will contain it?

BewareMyPower:
Currently 0.11.0 is in release, so I think it's 0.12.0

fud:
Just wondering when 0.12.0 would be released as we would love to use this feature?

PR #1104, opened 2023-10-19
fix: flush block

rfyiamcool:
### summary

When the producer's close method is executed, the runEventsLoop coroutine will exit. If producer.flush is executed at this time, there is a possibility of blocking.

`If you close first and then flush, `Flush()` will block.`

```
go func() {
      close() := func() {
            close(internalQueue)
            producer.Close()
      }
}()

go func() {
      for _, msg := range internalQueue {
          producer.Send(...)
      }
      producer.Flush() // block
}
```

Sometimes, we cannot control the order of concurrent execution. 😁

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -1381,6 +1381,11 @@ func (p *partitionProducer) LastSequenceID() int64 {
 }
 
 func (p *partitionProducer) Flush() error {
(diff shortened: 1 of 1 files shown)


tisonkun (commented):
This patch doesn't "fix" an issue if any because it's possible that the check passed first but the producer gets closed just after the check passed.

PR #1107, opened 2023-10-20
[Issue 1105][pulsaradmin] fix AutoTopicCreation for type non-partitioned

### LINKED ISSUE #1105: Impossible to set AutoTopicCreation policy to non-partitioned
#### Expected behavior

Client.Namespaces().SetTopicAutoCreation() allows setting a TopicAutoCreation _allowed_ policy on the namespace with default topic type _non-partitioned_.

#### Actual behavior

SetTopicAutoCreation has a parameter utils.TopicAutoCreationConfig, this struct has an int field Partitions (defaultNumPartitions) which cannot be nil. When trying to SetTopicAutoCreation with config (Allow=true,Type="non-partitioned",Partitions=0) the API call will fail with: 

`{"reason":"Invalid configuration for autoTopicCreationOverride. the detail is [defaultNumPartitions] is not allowed to be set when the type is non-partition."}`

This happens with any value for Partitions. In the pulsar admin rest-api the defaultNumPartitions field is optional, it should either be made optional as well in TopicAutoCreationConfig or the library should not include the field when Type="non-partitioned".

With the rest API I can do:
`
curl --data '{"allowAutoTopicCreation":true,"topicType":"non-partitioned"}' --header "Content-Type: application/json" --header "Accept: application/json" $APIHOST/admin/v2/namespaces/$tenant/$namespace/autoTopicCreation`

#### Steps to reproduce

Call Client.Namespaces().SetTopicAutoCreation() with a TopicAutoCreation with Allow=true and Type="non-partitioned"

#### System configuration
**Pulsar version**: 2.11.2

tomjo:
Fixes #1105

### Motivation

To allow setting the AutoTopicCreation policy to non-partitioned it is required to not send the partitions parameter, by making Partitions a pointer instead of an int this is possible.

### Modifications

Change type of Partitions in TopicAutoCreationConfig from int to *int

### Verifying this change

I tested it by using it in our application as the admin code currently does not have much test code. I saw here https://github.com/apache/pulsar-client-go/pull/1085 that there are reorganization plans + setting up test scaffolding in the future.

### DIFF
--- a/pulsaradmin/pkg/utils/topic_auto_creation_config.go
+++ b/pulsaradmin/pkg/utils/topic_auto_creation_config.go
@@ -20,5 +20,5 @@ package utils
 type TopicAutoCreationConfig struct {
 	Allow      bool      `json:"allowAutoTopicCreation"`
 	Type       TopicType `json:"topicType"`
-	Partitions int       `json:"defaultNumPartitions"`
+	Partitions *int      `json:"defaultNumPartitions"`
 }


tisonkun (commented):
@tomjo can you elaborate a bit for the use case where this change makes difference?

tomjo:
We're using a component that publishes messages to a topic with a generated name, so we want to use topic auto creation for the namespace so that these topics are automatically created. We don't need partitions for these topics so we want the default for topic auto creation to be non-partitioned. This is not possible now (see linked issue) since the Partitions int field in the struct has a default value of 0 causing it to be added to the request body of the api call, which pulsar rejects with an error. By making it a pointer, it can be nil and it is not serialized in the body allowing the api call to succeed in setting the auto topic creation policy to enabled and non-partitioned.

We're currently using topic auto creation with type partitioned and 1 partition which works but it would be nicer if the client supports what the rest api supports

RobertIndie (approved):

Issue #1103, opened 2023-10-12
Unable to read BOOLEAN value from pulsar message


Sonalit23:
#### Expected behavior

Sending pulsar message with value type Boolean (wrapper class) in java with RSA encryption enabled but unable to read it at consumer side with go client. correct boolean value should be received provided correct decryption keys.

#### Actual behavior

value is always empty and not sure what will be the right schema type for boolean value.

#### Steps to reproduce

Here I am writing producer in go (integration test) instead of JAVA.
producer code:
```
keyReader := crypto.NewFileKeyReader('PROVIDE PATH TO PUBLIC KEY', 'PROVIDE PATH TO PRIVATE KEY')
	producer, _ := client.CreateProducer(pulsar.ProducerOptions{
		Topic: "some-topic",
		Encryption: &pulsar.ProducerEncryptionInfo{
			KeyReader:                   keyReader,
			ProducerCryptoFailureAction: crypto.ProducerCryptoFailureActionFail,
			Keys:                        []string{"client-rsa.pem"},
		},
	})

	msgId, err := producer.Send(ctx, &pulsar.ProducerMessage{
		Key:   "abc-key",
		Value: true,
	})
```
Consumer code:
```
keyReader := crypto.NewFileKeyReader('PROVIDE PATH TO PUBLIC KEY', 'PROVIDE PATH TO PRIVATE KEY')
	options := pulsar.ConsumerOptions{
		Topic:            topic-name,
		Type:             pulsar.Shared,
		MessageChannel:              channel,
		SubscriptionInitialPosition: pulsar.SubscriptionPositionEarliest,
		Decryption: &pulsar.MessageDecryptionInfo{
			KeyReader:                   keyReader,
			ConsumerCryptoFailureAction: crypto.ConsumerCryptoFailureActionFail,
		},
	}
```

How to use schematype for BOOLEAN value? I checked the https://github.com/apache/pulsar-client-go/blob/master/pulsar/schema.go but it is unclear for boolean value. any help would be appreciated.
#### System configuration
**Pulsar version**: 3.0.x

dao-jun:
1. Pulsar-client-go not supports BOOLEAN schema yet

2. You didn't set a `Schema` for the producer and use the `Value` field, the `Value` field needs a `Schema` to encode it into `Payload([]byte)`. If the `Schema` not set, the message you sent will has an empty body. So you can't read the message payload when you consuming messages.
In this condition, you can
```
[...]

PR #1108, opened 2023-10-21
[feat][schema] Introduce BooleanSchema

dao-jun:
Fixes https://github.com/apache/pulsar-client-go/issues/1103

### Motivation

Introduce `BooleanSchema`

### Modifications

Add `BooleanSchema`

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (no)
  - The schema: (yes)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/schema.go
+++ b/pulsar/schema.go
@@ -101,6 +101,8 @@ func NewSchema(schemaType SchemaType, schemaData []byte, properties map[string]s
 		s = NewProtoSchema(schemaDef, properties)
 	case AVRO:
 		s = NewAvroSchema(schemaDef, properties)
+	case BOOLEAN:
+		s = NewBooleanSchema(properties)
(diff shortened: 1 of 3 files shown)


pulsar/schema.go
```diff
@@ -178,7 +180,7 @@ func (js *JSONSchema) Decode(data []byte, v interface{}) error {
 }
 
 func (js *JSONSchema) Validate(message []byte) error {
-	return js.Decode(message, nil)
+	return nil
```
tisonkun, line 183: why do we need these changes?

dao-jun: `Schema.Validate` method has never been called elsewhere before, but in actually, these methods will panic once they are called.
The second param cannot be nil.

PR #1119, opened 2023-10-24
[fix] Complete comment on ProducerInterceptor interface BeforeSend method

ojcm:
### Motivation

The comment against the `BeforeSend` method in the `ProducerInterceptor` interface is incomplete. Whilst it's fairly easy to infer the missing words, the comment is clearer when complete.

https://github.com/apache/pulsar-client-go/blob/af56e6031d8021ddcff188f94c4d744eba3b3211/pulsar/producer_interceptor.go#L21-L22

### Modifications

Completed the comment so it reads

```
// BeforeSend This is called before send the message to the brokers. This method is allowed to modify the
// message.
```

### Verifying this change

This is the first time I've come across these interceptors so I've verified that the `BeforeSend` method can modify the message: https://go.dev/play/p/UQ4Q1MNPOpf

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - If a feature is not applicable for documentation, explain why? It is the documentation.

### DIFF
--- a/pulsar/producer_interceptor.go
+++ b/pulsar/producer_interceptor.go
@@ -19,6 +19,7 @@ package pulsar
 
 type ProducerInterceptor interface {
 	// BeforeSend This is called before send the message to the brokers. This method is allowed to modify the
+	// message.
 	BeforeSend(producer Producer, message *ProducerMessage)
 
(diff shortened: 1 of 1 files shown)


tisonkun (approved):
Thank you!

PR #1133, opened 2023-11-16
[Issue 1132] Fix JSONSchema unmarshalling in TableView

### LINKED ISSUE #1132: [BUG] TableView with JSONSchema returns incorrect type in Get
#### Expected behavior

Using a TableView with a JSONSchema, the `Get` method should return a instance of the Go type provided via the `TableViewOptions.SchemaValueType` field.

#### Actual behavior

The type returned from the `Get` method is `map[string]interface{}` (and I presume it could also be `[]interface{}`, `string`, `float64`, `bool` or `nil` depending on the raw JSON).

#### Steps to reproduce

The following test reproduces the issue, where `exampleSchemaDef`, `testJSON`, `lookupURL` and `newTopicName` are defined in the `pulsar` package test files.
```
func TestTableViewJSONSchema(t *testing.T) {
	var (
		schema = NewJSONSchema(exampleSchemaDef, nil)
		key    = "testKey"
		value  = testJSON{ID: 1, Name: "Pulsar"}
	)

		URL: lookupURL,
	})

	assert.NoError(t, err)
	defer client.Close()

	topic := newTopicName()

		Topic:  topic,
		Schema: schema,
	})
	assert.NoError(t, err)
	defer producer.Close()

	_, err = producer.Send(context.Background(), &ProducerMessage{
		Key:   key,
		Value: value,
	})
	assert.NoError(t, err)

	tv, err := client.CreateTableView(TableViewOptions{
		Topic:           topic,
		Schema:          schema,
		SchemaValueType: reflect.TypeOf(value),
	})
	assert.NoError(t, err)
	defer tv.Close()

	gotValue := tv.Get(key)
	// The following checks fail because `gotValue` has type map[string]interface{}
	assert.IsType(t, value, gotValue)
	assert.Equal(t, value, gotValue)
}
```

#### System configuration
**Pulsar version**: 3.1.0

ojcm:
Fixes #1132 

### Motivation
Fix issue #1132 - using JSONSchema with TableView

### Modifications

- Set a concrete type in the `payload` variable before JSON-unmarshalling into that variable. This allows the JSON package to identify and use the type rather than seeing it as `interface{}`.
- Use `reflect.Indirect(payload).Interface()` when storing the payload and passing it to listeners to remove the pointer from `reflect.New`.
- Add test coverage for `TableView.Get` covering all supported schema types.
- Add test coverage for `TableView.ForEachAndListen` for JSONSchema.

Additional minor changes. They didn't seem worth their own MRs but I'm happy to split them out if that's better.
- Correct typo in comments on `TableView.ForEach` and `TableView.ForEachAndListen` interface methods.
- Correct `TableView.ForEachAndListen` comment to clarify that it continues to call the given action on future messages.
- Correct formatting directive (`%w` -> `%v`) in error log `tv.logger.Errorf("msg.GetSchemaValue() failed with %v; msg is %v", err, msg)`. (This indirectly calls `fmt.Sprintf` in logrus which doesn't support `%w`).

### Verifying this change

This change has some existing test coverage from `TestTableView`. 

This change has extended test coverage, see `TestTableViewSchemas` and `TestForEachAndListenJSONSchema`.

  - The public API: yes? This fixes broken behaviour in the API. It's theoretically possible that users could be relying on broken behaviour and are expecting the incorrect type from the TableView methods when using JSONSchema. In my opinion, this is unlikely and so shouldn't be considered a breaking change.
  - The schema: no
  - The wire protocol: no

### Documentation

  - If a feature is not applicable for documentation, explain why? not applicable

### DIFF
--- a/pulsar/table_view.go
+++ b/pulsar/table_view.go
@@ -65,12 +65,12 @@ type TableView interface {
 	// Keys returns a slice of the keys contained in this TableView.
 	Keys() []string
 
-	// ForEach performs the give action for each entry in this map until all entries have been processed or the action
+	// ForEach performs the given action for each entry in this map until all entries have been processed or the action
 	// returns an error.
 	ForEach(func(string, interface{}) error) error
 
-	// ForEachAndListen performs the give action for each entry in this map until all entries have been processed or
-	// the action returns an error.
+	// ForEachAndListen performs the given action for each entry in this map until all entries have been processed or
+	// the action returns an error.  The given action will then be performed on each new entry in this map.
 	ForEachAndListen(func(string, interface{}) error) error
 
 	// Close closes the table view and releases resources allocated.
--- a/pulsar/table_view_impl.go
+++ b/pulsar/table_view_impl.go
@@ -245,19 +245,18 @@ func (tv *TableViewImpl) handleMessage(msg Message) {
 	tv.dataMu.Lock()
 	defer tv.dataMu.Unlock()
 
-	var payload interface{}
+	payload := reflect.New(tv.options.SchemaValueType)
 	if len(msg.Payload()) == 0 {
 		delete(tv.data, msg.Key())
 	} else {
-		payload = reflect.Indirect(reflect.New(tv.options.SchemaValueType)).Interface()
-		if err := msg.GetSchemaValue(&payload); err != nil {
-			tv.logger.Errorf("msg.GetSchemaValue() failed with %w; msg is %v", err, msg)
+		if err := msg.GetSchemaValue(payload.Interface()); err != nil {
+			tv.logger.Errorf("msg.GetSchemaValue() failed with %v; msg is %v", err, msg)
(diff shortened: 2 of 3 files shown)


RobertIndie (approved):

RobertIndie:
> Additional minor changes. They didn't seem worth their own MRs but I'm happy to split them out if that's better.

They are minor changes. I'm OK with including them in one PR. Thanks for your fix.

ojcm:
I've made a couple of changes to the tests that failed the workflow:
- Replaced a helper function that used generics `pointer[T]` with a non-generic version `strPointer` as it was only helpful for strings anyway.
- Removed type assertions that copied locks with variants that don't. The lock copying wasn't an issue in this context but the linter can't know that.

ojcm:
One more change to replace `any`s with `interface{}`s to work with older Go versions.

PR #1136, opened 2023-11-21
feat: added missing fields on TopicStats

jffp113:
### Motivation

When using the admin api with golang there is some missing fields on topic stats

### Modifications

Added missing fields on topic stats

  - The public API: yes
  - The schema: yes
  - The wire protocol: no

### Documentation

  - Does this pull request introduce a new feature?  no
  - If yes, how is the feature documented?  PulsarDoc

### DIFF
--- a/pulsaradmin/pkg/utils/data.go
+++ b/pulsaradmin/pkg/utils/data.go
@@ -242,11 +242,16 @@ type TopicStats struct {
 }
 
 type PublisherStats struct {
-	ProducerID      int64             `json:"producerId"`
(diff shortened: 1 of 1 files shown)


RobertIndie (commented):
Could you add a test to verify that it could get these stats correctly?

geniusjoe:
I think we need to add `GetPartitionedStats` and `GetPartitionedStatsWithOption` interface test cases in `topic_test.go`. Since `make test` command will automatically deploy a pulsar broker in test period, it is quite easy to write test cases. And we can refer `TestPartitionStats` test case in the latest default branch.

image

PR #1141, opened 2023-12-05
fix: channel deadlock in regexp consumer

goncalo-rodrigues:
### Motivation

When using a regexp consumer, there's a race condition between producing and consuming new discovered topics. If the discover topic takes too long or the auto discovery period is too short, then multiple ticker.C messages may be processed in a row which will block on the subcsribe/unsubscribe channels as they only have a buffer size of 1. This will block new topics from being discovered forever.

### Modifications

Moved the consumers into their own goroutines and use an unbuffered channel. There's multiple ways to go about it but it's good practice to keep consumers and producers separate. Consumers are run until the channels they are consumed from are closed, which happens when the producer (monitor) returns.

### Verifying this change

- [ x ] Make sure that the change passes the CI checks.

  - *Added integration tests that fail when the patch is not there. There's 2-3 discover calls and then the monitor blocks indefinitely and the topics are not discovered. 

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - If a feature is not applicable for documentation, explain why? not a feature

### DIFF
--- a/pulsar/consumer_regex.go
+++ b/pulsar/consumer_regex.go
@@ -76,8 +76,8 @@ func newRegexConsumer(c *client, opts ConsumerOptions, tn *internal.TopicName, p
 		pattern:   pattern,
 
 		consumers:     make(map[string]Consumer),
-		subscribeCh:   make(chan []string, 1),
-		unsubscribeCh: make(chan []string, 1),
+		subscribeCh:   make(chan []string),
+		unsubscribeCh: make(chan []string),
 
 		closeCh: make(chan struct{}),
 
@@ -163,12 +163,11 @@ func (c *regexConsumer) Ack(msg Message) error {
 	return c.AckID(msg.ID())
 }
 
-func (c *regexConsumer) ReconsumeLater(msg Message, delay time.Duration) {
+func (c *regexConsumer) ReconsumeLater(_ Message, _ time.Duration) {
 	c.log.Warnf("regexp consumer not support ReconsumeLater yet.")
 }
 
-func (c *regexConsumer) ReconsumeLaterWithCustomProperties(msg Message, customProperties map[string]string,
-	delay time.Duration) {
+func (c *regexConsumer) ReconsumeLaterWithCustomProperties(_ Message, _ map[string]string, _ time.Duration) {
 	c.log.Warnf("regexp consumer not support ReconsumeLaterWithCustomProperties yet.")
 }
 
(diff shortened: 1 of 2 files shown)


pulsar/consumer_regex.go
```diff
@@ -320,6 +319,25 @@ func (c *regexConsumer) closed() bool {
...
+		}
+	}()
+
+	go func() {
+		for topics := range c.unsubscribeCh {
+			if len(topics) > 0 && !c.closed() {
+				c.unsubscribe(topics)
+			}
+		}
+	}()
```
RobertIndie, line 339: What about combining these two channels into one single channel?
Otherwise, under poor network conditions, where the process of subscribing or unsubscribing may take an long period, scenarios like the following might occur:

- A consumer is created.
- Topic A is created.
- The consumer discovers the topic and starts to subscribe to Topic A.
- Topic A is deleted.
- The consumer discovers the topic again, but the knownTopics list remains unchanged. Thus, the consumer doesn't unsubscribe from Topic A.
- Topic A is subscribed to, but this action fails.

This chain of events leads to the final outcome: Topic A should be unsubscribed from, but it isn't.

Another scenario that could result in data loss:
- A consumer is created and subscribes to Topic A.
- Topic A is deleted.
- The consumer discovers the topic and begins to unsubscribe from Topic A.
- Topic A is recreated.
- The consumer discovers the topic again, but the knownTopics list remains unchanged. As a result, the consumer doesn't resubscribe to Topic A.
- Topic A remains unsubscribed.
The final outcome in this case: Topic A is not subscribed to, and any subsequent messages in Topic A will be lost.

Merging the channels could simplify this logic and make it more robust. It would be more effective to have the monitoring of this topics list occur on a single thread. This approach is guaranteed in the Java client, preventing the aforementioned issues from occurring.

goncalo-rodrigues: That's a good point. I would go one step further and simplify this logic to not use channels but rather process subscriptions in the thread that processes the ticker. What do you think?

Theres no point to the channels as there's a single producer and a single consumer. And this already runs in a separate goroutine specific to the auto discovery of topics. If in the future we have additional producers then channels can be reintroduced with any new requirements in mind.

RobertIndie: Great. That makes sense to me.

goncalo-rodrigues: Perfect, done in the latest commit.

RobertIndie (approved):
LGTM. Thanks for your contribution.

PR #1165, opened 2024-01-31
[Producer] respect context cancellation in Flush

jayshrivastava:
### Motivation

The producer's `Flush` method does not respect context cancellation. If the caller's context get's cancelled, it will have to wait for the producer to finish flushing.

### Modifications

This change adds a `FlushWithCtx` method which takes a context and selects on two channels.

### Verifying this change

This change is already covered by existing tests. See code changes for relevant tests.

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - If a feature is not applicable for documentation, explain why? It's a very trivial change. It adds a common pattern in Go.

### DIFF
--- a/pulsar/producer.go
+++ b/pulsar/producer.go
@@ -237,9 +237,12 @@ type Producer interface {
 	// return the last sequence id published by this producer.
 	LastSequenceID() int64
 
-	// Flush all the messages buffered in the client and wait until all messages have been successfully
-	// persisted.
+	// Deprecated: Use `FlushWithCtx()` instead.
 	Flush() error
+
+	// Flush all the messages buffered in the client and wait until all messageshave been successfully
+	// persisted.
+	FlushWithCtx(ctx context.Context) error
 
 	// Close the producer and releases resources allocated
 	// No more writes will be accepted from this producer. Waits until all pending write request are persisted. In case
--- a/pulsar/producer_impl.go
+++ b/pulsar/producer_impl.go
@@ -334,11 +334,15 @@ func (p *producer) LastSequenceID() int64 {
 }
 
 func (p *producer) Flush() error {
+	return p.FlushWithCtx(context.Background())
+}
+
+func (p *producer) FlushWithCtx(ctx context.Context) error {
 	p.RLock()
 	defer p.RUnlock()
 
 	for _, pp := range p.producers {
-		if err := pp.Flush(); err != nil {
+		if err := pp.FlushWithCtx(ctx); err != nil {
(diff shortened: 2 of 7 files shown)


jayshrivastava:
@RobertIndie Would you mind taking a look at this please?

pulsar/producer.go
```diff
@@ -239,7 +239,7 @@ type Producer interface {
 
 	// Flush all the messages buffered in the client and wait until all messages have been successfully
 	// persisted.
-	Flush() error
+	Flush(ctx context.Context) error
```
merlimat, line 242: While I agree that we should have taking the context in this method, right now this would be a breaking API change. 

I'd say to add a new `FlushWithContext()` method and, perhaps, mark the other one as deprecated.

jayshrivastava: Just pushed an update. Lmk how it looks!

Btw, do you know if the library will support shutting down via context cancellation  in the future? There's some examples of where we don't do it:

https://github.com/apache/pulsar-client-go/blob/efd9806be9aec8a7ec0e602664bd7dd008822627/pulsar/producer_partition.go#L491

https://github.com/apache/pulsar-client-go/blob/efd9806be9aec8a7ec0e602664bd7dd008822627/pulsar/producer_partition.go#L1470-L1481

As long as things don't take too long to shut down, I suppose it's fine. Ideally, if the context is cancelled, you don't want to wait for a clean/graceful shutdown where we wait for the done channel. We want all the goroutines to terminate right where they are and immediately returned.

jayshrivastava: I would be happy to file an issue if there is not one already.

jayshrivastava: I can't implement `CloseWithCtx` such as in this PR because goroutines might leak (the call to close would terminate before the signal is given on the doneCh, indicating that things have shut down). I think it would be a bigger change.

pulsar/producer_impl.go
```diff
@@ -346,6 +346,19 @@ func (p *producer) Flush() error {
 	return nil
 }
 
+func (p *producer) FlushWithCtx(ctx context.Context) error {
```
merlimat, line 349: `Flush()` could just call `p.FlushWithCtx(context.Background())`

RobertIndie: +1

jayshrivastava: Done!

RobertIndie (approved):

merlimat (approved):

PR #1174, opened 2024-02-10
Enable accessing Message in TableView

petermnhull:
### Motivation

The current TableView implementation only stores the payload encoded to the SchemaValueType of each incoming message. There are many use cases where accessing the full metadata of the pulsar.Message type is needed, so this PR refactors the map storage of messages to be the pulsar.Message received from the consumer rather than the payload.

All the existing APIs are maintained by performing the payload encoding when the methods (e.g. Get) are called.

A single new public method, Message, is introduced to access the pulsar.Message for each key so that accompanying metadata for a payload can be easily retrieved.

### Modifications

- The TableView implementation `handleMessage` method now stores the pulsar.Message type and only encodes to the SchemaValueType for listeners.
- `Get` and `Entries` use the SchemaValueType when called to convert from pulsar.Message to the payload

### Verifying this change

This change added tests, and is covered by existing tests, and can be verified as follows:
- Running table_view_test.go which runs new and existing tests for all functionality.

  - Dependencies (does it add or upgrade a dependency): _No_
  - The public API: _Yes, new Message method on TableView interface. Existing API not impacted_
  - The schema: _No_
  - The default values of configurations: _No_
  - The wire protocol: _No_

### Documentation

  - Does this pull request introduce a new feature? _Yes_
  - If yes, how is the feature documented? _Code documentation, and happy to create new issue for other required documentation._


RobertIndie (commented):
This PR invokes new API changes. We need to draft a PIP for it: https://github.com/apache/pulsar/tree/master/pip

pulsar/table_view_impl.go
```diff
@@ -186,10 +203,23 @@ func (tv *TableViewImpl) Get(key string) interface{} {
...
+		}
 		data[k] = v
 	}
+
+	return data
+}
+
+func (tv *TableViewImpl) Messages() map[string]Message {
+	tv.dataMu.Lock()
+	defer tv.dataMu.Unlock()
```
merlimat, line 222: This is returning the same underlying map, allowing for race-conditions and modifications from the application. 

Note: the existing code is also wrong because it was returning `tv.data` instead of `data`

Issue #1177, opened 2024-02-20
Multi-topics reader support


RobertIndie:
**Is your feature request related to a problem? Please describe.**
Currently, there is an issue with the reader implementation. If the reader is creating, it won't get the topic metadata from the topic. The reader can only read messages from a single topic. If the topic is a partitioned topic, the reader won't know that and will try to create a non-partition topic with the same name. And it will lead to this issue: https://github.com/apache/pulsar/issues/22032

**Describe the solution you'd like**
- Support multi-topics reader

**Describe alternatives you've considered**
No

**Additional context**
No

PR #1178, opened 2024-02-20
[feat] Support partitioned topic reader

RobertIndie:
Master Issue: #1177

### Motivation

Currently, there is an issue with the reader implementation. If the reader is creating, it won't get the topic metadata from the topic. The reader can only read messages from a single topic. If the topic is a partitioned topic, the reader won't know that and will try to create a non-partition topic with the same name. And it will lead to this issue: https://github.com/apache/pulsar/issues/22032

### Modifications

- Support partitioned topic reader

### Verifying this change

This change added tests.

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -30,7 +30,6 @@ import (
 	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
 	"github.com/apache/pulsar-client-go/pulsar/log"
 	pkgerrors "github.com/pkg/errors"
-	uAtomic "go.uber.org/atomic"
 )
 
 const defaultNackRedeliveryDelay = 1 * time.Minute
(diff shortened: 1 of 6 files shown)


pulsar/consumer_impl.go
```diff
@@ -707,6 +709,34 @@ func (c *consumer) checkMsgIDPartition(msgID MessageID) error {
...
+
+	wg.Add(len(c.consumers))
+
+	var hasNext uAtomic.Bool
+	for _, pc := range c.consumers {
+		pc := pc
+		go func() {
+			defer wg.Done()
+			if pc.hasNext() {
+				hasNext.Store(true)
```
shibd, line 723: if a `pc` has the next message, we can stop the wait early?

RobertIndie: I had considered this before. However, I didn't implement it then because it required introducing a channel and context, which could add complexity and risks but without significant benefits. Anyway, I've just added this implementation and have carefully checked to ensure there won't be any panic.

shibd (approved):

RobertIndie:
This is also an important bug fix for the issue https://github.com/apache/pulsar/issues/22032, so I'm going to cherry-pick to branch-0.12

PR #1182, opened 2024-02-23
[fix] Fix Infinite Loop in Reader's `HasNext` Function

### LINKED ISSUE #1171: Reader HasNext() enters infinite loop
**Issue**
https://github.com/apache/pulsar-client-go/blob/f476814a9e1bac484e9589d4fb3f299066f54ddc/pulsar/reader_impl.go#L180

The code above logs errors infinitely until the HasNext() call is able to proceed.  If it never resolves, it just logs indefinitely.  For us, that resulted in flooding our logging system with unnecessary logs.

**Expected Behavior**. 
I would expect that on error, the loop would exit and return the error.   Maybe I am not understanding the intended behavior for this code.

**Pulsar Client Version**
v11 (code also present in v12)

RobertIndie:
Fixes #1171

### Motivation

If `getLastMessageId` continually fails, the reader.HasNext can get stuck in an infinite loop. Without any backoff, the reader would keep trying forever.

### Modifications

- Implemented a backoff policy for `getLastMessageID`.
- If HasNext fails, it now returns false.

#### Should the reader.HasNext returned `false` in case of failure?

Currently, the `HasNext` method doesn't report errors. However, failure is still possible. For instance, if `getLastMessageID` repeatedly fails and hits the retry limit. An option is to keep trying forever, but this would stall all user code. This isn't user-friendly, so I rejected this solution.

#### Couldn't utilize the BackOffPolicy in the Reader Options

The `HasNext` retry mechanism requires to use of `IsMaxBackoffReached` for the backoff. But it isn't exposed in the `BackOffPolicy` interface. Introducing a new method to the `BackOffPolicy` would introduce breaking changes for the user backoff implementation. So, I choose not to implement it. Before we do it, we need to refine the BackOffPolicy.

### Verifying this ch
This change added tests.

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -568,15 +568,31 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
 
 func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
-		pc.log.WithField("state", state).Error("Failed to redeliver closing or closed consumer")
-		return nil, errors.New("failed to redeliver closing or closed consumer")
+		pc.log.WithField("state", state).Error("Failed to getLastMessageID for the closing or closed consumer")
+		return nil, errors.New("failed to getLastMessageID for the closing or closed consumer")
 	}
-	req := &getLastMsgIDRequest{doneCh: make(chan struct{})}
-	pc.eventsCh <- req
+	backoff := &internal.DefaultBackoff{}
+	request := func() (*trackingMessageID, error) {
+		req := &getLastMsgIDRequest{doneCh: make(chan struct{})}
+		pc.eventsCh <- req
 
-	// wait for the request to complete
-	<-req.doneCh
-	return req.msgID, req.err
+		// wait for the request to complete
+		<-req.doneCh
+		return req.msgID, req.err
+	}
+	for {
+		msgID, err := request()
+		if err == nil {
+			return msgID, nil
+		}
+		pc.log.Info("RETRY !!!")
+		nextDelay := backoff.Next()
+		if backoff.IsMaxBackoffReached() {
+			return nil, fmt.Errorf("failed to getLastMessageID due to %w", err)
+		}
+		pc.log.WithError(err).Errorf("Failed to get last message id from broker, retrying in %v...", nextDelay)
+		time.Sleep(nextDelay)
+	}
 }
 
 func (pc *partitionConsumer) internalGetLastMessageID(req *getLastMsgIDRequest) {
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -594,6 +594,7 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 			return msgID, nil
 		}
 		if remainTime <= 0 {
+			pc.log.WithError(err).Error("Failed to getLastMessageID")
 			return nil, fmt.Errorf("failed to getLastMessageID due to %w", err)
 		}
 		nextDelay := backoff.Next()
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -573,7 +573,13 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 		pc.log.WithField("state", state).Error("Failed to getLastMessageID for the closing or closed consumer")
 		return nil, errors.New("failed to getLastMessageID for the closing or closed consumer")
 	}
-	backoff := &internal.DefaultBackoff{}
+	remainTime := pc.client.operationTimeout
+	var backoff internal.BackoffPolicy
+	if pc.options.backoffPolicy != nil {
+		backoff = pc.options.backoffPolicy
+	} else {
+		backoff = &internal.DefaultBackoff{}
+	}
 	request := func() (*trackingMessageID, error) {
 		req := &getLastMsgIDRequest{doneCh: make(chan struct{})}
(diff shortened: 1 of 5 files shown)


BewareMyPower (commented):
A better idea might be modifying the `readNext` API.

```golang
     HasNext() (bool, error)
```

According to the module version numbering,

> Automatic pseudo-version number
v0.x.x: Signals that the module is still in development and unstable. This release carries no backward compatibility or stability guarantees.

pulsar/reader_test.go
```diff
@@ -1023,3 +1024,58 @@ func createPartitionedTopic(topic string, n int) error {
...
+	client, err := NewClient(ClientOptions{
+		URL: serviceURL,
+	})
+	assert.Nil(t, err)
+	topic := newTopicName()
+	r, err := client.CreateReader(ReaderOptions{
+		Topic:          topic,
+		StartMessageID: EarliestMessageID(),
+		// Retry the connection 10 seconds after it's disconnected.
+		BackoffPolicy: newTestBackoffPolicy(10*time.Second, 10*time.Second),
```
BewareMyPower, line 1053: Maybe we can use a short timeout (3 seconds or 1 second?). 10 seconds are too long for a single test.

pulsar/consumer_partition.go
```diff
@@ -570,15 +570,30 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
...
 func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
-		pc.log.WithField("state", state).Error("Failed to redeliver closing or closed consumer")
-		return nil, errors.New("failed to redeliver closing or closed consumer")
+		pc.log.WithField("state", state).Error("Failed to getLastMessageID for the closing or closed consumer")
+		return nil, errors.New("failed to getLastMessageID for the closing or closed consumer")
 	}
-	req := &getLastMsgIDRequest{doneCh: make(chan struct{})}
-	pc.eventsCh <- req
+	backoff := &internal.DefaultBackoff{}
```
BewareMyPower, line 576: I'm wondering why a `DefaultBackoff{}` works in tests.

But theoretically we should use `pc.options.backoffPolicy` and add a `IsMaxBackoffReached` method to the `BackoffPolicy` interface.

RobertIndie: I've explained it in the PR description. Adding a new method to the interface requires code changes for the user implementation BackoffPolicy. All users should add the implementation of `IsMaxBackoffReached` for their own.

We could redesign it in the next major version or `1.0.0`.

BewareMyPower: But the `DefaultBackoff` object here might have a different policy with the user provided backoff.

pulsar/reader_test.go
```diff
@@ -1023,3 +1025,58 @@ func createPartitionedTopic(topic string, n int) error {
...
+		done <- true
+	}()
+
+	select {
+	case <-maxTimer.C:
+		t.Fatal("r.HasNext() blocked for more than 20s")
+	case <-done:
+
+		if minTimer.Stop() {
+			t.Fatal("r.HasNext() did not block for at least 10s")
```
BewareMyPower, line 1075:
```suggestion
			t.Fatal("r.HasNext() did not block for at least 1s")
```

pulsar/reader_test.go
```diff
@@ -1023,3 +1025,58 @@ func createPartitionedTopic(topic string, n int) error {
...
+		}
+		if !maxTimer.Stop() {
+			t.Fatal("r.HasNext() blocked for more than 20s")
```
BewareMyPower, line 1078:
```suggestion
			t.Fatal("r.HasNext() blocked for more than 2s")
```

pulsar/reader_test.go
```diff
@@ -1023,3 +1025,58 @@ func createPartitionedTopic(topic string, n int) error {
...
+
+	done := make(chan bool)
+	go func() {
+		assert.False(t, r.HasNext())
+		done <- true
+	}()
+
+	select {
+	case <-maxTimer.C:
+		t.Fatal("r.HasNext() blocked for more than 20s")
```
BewareMyPower, line 1071:
```suggestion
		t.Fatal("r.HasNext() blocked for more than 2s")
```

pulsar/consumer_partition.go
```diff
@@ -570,15 +570,30 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
...
+	}
+	for {
+		msgID, err := request()
+		if err == nil {
+			return msgID, nil
+		}
+		nextDelay := backoff.Next()
+		if backoff.IsMaxBackoffReached() {
+			return nil, fmt.Errorf("failed to getLastMessageID due to %w", err)
+		}
```
BewareMyPower, line 593: Then I found even after removing this check, the test could still pass. I found the loop breaks after `request()` returns an invalid message ID (`-1:-1:0`) and a nil `err`. Could you confirm it again?

RobertIndie:
@BewareMyPower After reevaluating, I've concluded that we shouldn't check for maxBackoff in the backoff policy. Similar to the Java client, the maximum timeout should be regulated by the ClientOptions.OperationTimeout.

I have also chosen a new way to simulate the failure of getting the last message ID. It would make the test more stable.

And also, I found that there are some issues with the BackoffPolicy. I tracked it in another issue: https://github.com/apache/pulsar-client-go/issues/1187.

BewareMyPower:
>  the maximum timeout should be regulated by the ClientOptions.OperationTimeout.

Yes, that's right.

BewareMyPower (approved):

shadygrove:
Thank you for the fix!

PR #1192, opened 2024-03-01
[Issue 1191][Consumer] allow override the Schema Creator by a consumer

adrianiacobghiula:
a consumer might want to use different implementation of a Schema

### Motivation

A consumer should be allowed to select / choose / override the default Schema. For example select to use https://github.com/hamba/avro and not github.com/linkedin/goavro/v2

### Modifications

Add options to 

### Verifying this change

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsaradmin/pkg/admin/admin.go
+++ b/pulsaradmin/pkg/admin/admin.go
@@ -112,3 +112,8 @@ func (c *pulsarClient) endpoint(componentPath string, parts ...string) string {
 		path.Join(escapedParts...),
 	)
 }
+
+// this function won't do any escaping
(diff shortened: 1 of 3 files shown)


RobertIndie (commented):
Introducing a new option would introduce complexity to the `ConsumerOptions.`
We could start a discussion to replace https://github.com/linkedin/goavro with https://github.com/hamba/avro. Could you share more motivation on this?
Would you like to initiate this discussion on the mailing list?

adrianiacobghiula:
I agree it increases complexity  :) On the other hand it offers the consumer/client the possibility to access all schema versions basically & create Schema for each of them. 
Yes. I would initiate a discussion on the mailing list :)

PR #1197, opened 2024-03-13
[fix] Reconnection logic and Backoff policy doesn't work correctly

### LINKED ISSUE #1187: [BUG] Reconnection logic and Backoff policy doesn't work correctly
I found that there are some issues with the BackoffPolicy:
1. We should use a new backoff instance for each use: https://github.com/apache/pulsar-client-go/blob/35ae05157b702804037446e1190a8f4e7557fdbf/pulsar/consumer_partition.go#L1656
2. In the reconnection logic, the checking for `IsMaxBackoffReached` doesn't respect the pc.options.backoffPolicy. https://github.com/apache/pulsar-client-go/blob/35ae05157b702804037446e1190a8f4e7557fdbf/pulsar/consumer_partition.go#L1700 
Actually, the `IsMaxBackoffReached` is incorrect. We should respect the operation timeout and check the whole retired time instead of the backoff delay time.
3. From the view of API design, we should move the BackoffPolicy out of the `internal` package.

crossoverJie:
Fixes #1187

### Modifications

- Move `backoff.go` to the `backoff` directory (because there are circular dependencies, they are not moved to the pulsar directory.)
- Create a new method for `BackOffPolicy` interface `IsMaxBackoffReached(delayReconnectTime, totalDelayReconnectTime time.Duration) bool`

This is a **breaking change** that modifies the package name and interface name.

Package: `internal`->`backoff`
Interface name: `BackoffPolicy`-> `Policy`

### Verifying this change

  - The public API: (yes)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation


RobertIndie (commented):
I'm concerned that such breaking changes could significantly impact users. Users would have to implement the new methods introduced by the backoff policy, even if they haven't encountered this issue. Perhaps we could consider including this change in the 1.0 version.

I have initiated a discussion of the Go client 1.0.0: https://lists.apache.org/thread/h022q6nrk23tz2k0qpxjf6j8m1o4qsrt

crossoverJie:
> Perhaps we could consider including this change in the 1.0 version.

Yes, I agree with you.

pulsar/consumer_partition.go
```diff
@@ -1673,11 +1680,8 @@ func (pc *partitionConsumer) reconnectToBroker() {
 			return
 		}
 
-		if pc.options.backoffPolicy == nil {
-			delayReconnectTime = defaultBackoff.Next()
-		} else {
-			delayReconnectTime = pc.options.backoffPolicy.Next()
-		}
+		delayReconnectTime = bo.Next()
+		totalDelayReconnectTime += delayReconnectTime
```
chinmayb, line 1684: We also need to reset the `delayReconnectTime` once connected after several retries in 1692.

nodece:
Any updates?

pulsar/consumer_partition.go
crossoverJie: After a successful reconnection at line 1699, the function will exit, and the `delayReconnectTime` variable will be initialized during the next reconnection. 

Therefore, there is no need to reset `delayReconnectTime` here.

crossoverJie:
The code conflicts have been resolved, but this PR involves changes to the public API. Do you recommend releasing it in version v0.13.0?

pulsar/backoff/backoff.go
```diff
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package internal
+package backoff
```
nodece, line 18: Could you move this pack to the `pulsar` directory and keep `BackoffPolicy`?

nodece, line 18: Could you move this pack to the `pulsar` directory and keep `BackoffPolicy`?

pulsar/backoff/backoff.go
```diff
@@ -26,10 +26,14 @@ func init() {
...
-// BackoffPolicy parameterize the following options in the reconnection logic to
+// Policy parameterize the following options in the reconnection logic to
 // allow users to customize the reconnection logic (minBackoff, maxBackoff and jitterPercentage)
-type BackoffPolicy interface {
+type Policy interface {
+	// Next returns the delay to wait before next retry
 	Next() time.Duration
+
+	// IsMaxBackoffReached evaluates if the max number of retries is reached
+	IsMaxBackoffReached(delayReconnectTime, totalDelayReconnectTime time.Duration) bool
```
nodece, line 36: `delayReconnectTime` and `totalDelayReconnectTime` are unnecessary, we can record this time in the backoff. 

Usually, we need to reset this backoff when some logic works fine, could you add the `Reset()` to this interface.

pulsar/consumer_partition.go
```diff
@@ -581,11 +583,11 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
...
 	}
 	remainTime := pc.client.operationTimeout
-	var backoff internal.BackoffPolicy
+	var bo backoff.Policy
 	if pc.options.backoffPolicy != nil {
-		backoff = pc.options.backoffPolicy
+		bo = pc.options.backoffPolicy
 	} else {
-		backoff = &internal.DefaultBackoff{}
+		bo = &backoff.DefaultBackoff{}
```
nodece, line 590:
```suggestion
		bo = NewDefaultBackoff()
```

pulsar/consumer_partition.go
```diff
@@ -1680,18 +1682,23 @@ func (pc *partitionConsumer) internalClose(req *closeRequest) {
...
 
-	var (
-		delayReconnectTime time.Duration
-		defaultBackoff     = internal.DefaultBackoff{}
-	)
+	var bo backoff.Policy
+	if pc.options.backoffPolicy != nil {
+		bo = pc.options.backoffPolicy
+	} else {
+		bo = &backoff.DefaultBackoff{}
```
nodece, line 1700:
```suggestion
		bo = NewDefaultBackoff()
```

pulsar/dlq_router.go
```diff
@@ -155,7 +156,7 @@ func (r *dlqRouter) getProducer(schema Schema) Producer {
 	}
 
 	// Retry to create producer indefinitely
-	backoff := &internal.DefaultBackoff{}
+	bo := &backoff.DefaultBackoff{}
```
nodece, line 159: Could you pass the backoff to the `dlqRouter`?

pulsar/producer_partition.go
```diff
@@ -448,17 +450,22 @@ func (p *partitionProducer) getOrCreateSchema(schemaInfo *SchemaInfo) (schemaVer
...
 
-	var (
-		delayReconnectTime time.Duration
-		defaultBackoff     = internal.DefaultBackoff{}
-	)
+	var bo backoff.Policy
+	if p.options.BackoffPolicy != nil {
+		bo = p.options.BackoffPolicy
+	} else {
+		bo = &backoff.DefaultBackoff{}
```
nodece, line 467:
```suggestion
		bo = NewDefaultBackoff()
```

pulsar/backoff/backoff.go
nodece: If you agree with this idea, you can move the backoff object to the global scope from the local scope, which can reduce the allocation of objects.

nodece (changes requested):

nodece:
> This is a breaking change that modifies the package name and interface name.

A breaking change that we add new method to the `Backoff` interface, not change package name.

In the user project, the user cannot import the `internal` package of the pulsar library, this is a limitation, please check https://go.dev/doc/modules/layout

pulsar/backoff/backoff.go
crossoverJie: This will lead to a circular dependency.

https://github.com/apache/pulsar-client-go/blob/master/pulsar/internal/http_client.go#L151

pulsar/backoff/backoff.go
crossoverJie: Done.

pulsar/dlq_router.go
crossoverJie: Done.

crossoverJie:
> Package: `internal`->`backoff`
> Interface name: `BackoffPolicy`-> `Policy`

Update:

`BackoffPolicy internal.BackoffPolicy` -> `BackOffPolicyFunc func() backoff.Policy`:

1. Because the Policy is concurrently called in the consumer, it has been modified to obtain a `policy` object each time using the `backoffPolicyFunc` function.

https://github.com/apache/pulsar-client-go/blob/1152cfc2e3c2024117ecff93a46d90e7eab0fa15/pulsar/consumer_partition.go#L1709-L1712


https://github.com/apache/pulsar-client-go/blob/1152cfc2e3c2024117ecff93a46d90e7eab0fa15/pulsar/consumer_partition.go#L607-L609


2. A new `Reset` function has been added to the Policy interface, which will be called after a successful reconnection. If a user's custom Policy requires the reuse of the same object, the data can be reset within this `Reset` function.

pulsar/producer_partition.go
```diff
@@ -473,11 +485,10 @@ func (p *partitionProducer) reconnectToBroker(connectionClosed *connectionClosed
 			delayReconnectTime = 0
 			assignedBrokerURL = connectionClosed.assignedBrokerURL
 			connectionClosed = nil // Only attempt once
-		} else if p.options.BackoffPolicy == nil {
-			delayReconnectTime = defaultBackoff.Next()
 		} else {
-			delayReconnectTime = p.options.BackoffPolicy.Next()
+			delayReconnectTime = bo.Next()
 		}
+		totalDelayReconnectTime += delayReconnectTime
```
nodece, line 491:
```suggestion
```

pulsar/producer_partition.go
```diff
@@ -448,17 +460,17 @@ func (p *partitionProducer) getOrCreateSchema(schemaInfo *SchemaInfo) (schemaVer
 }
 
 func (p *partitionProducer) reconnectToBroker(connectionClosed *connectionClosed) {
-	var maxRetry int
+	var (
+		maxRetry                                    int
+		delayReconnectTime, totalDelayReconnectTime time.Duration
```
nodece, line 465:
```suggestion
```

pulsar/producer_test.go
```diff
@@ -1273,11 +1275,13 @@ func TestProducerWithBackoffPolicy(t *testing.T) {
...
-	backoff := newTestBackoffPolicy(1*time.Second, 4*time.Second)
+	bo := newTestBackoffPolicy(1*time.Second, 4*time.Second)
 	_producer, err := client.CreateProducer(ProducerOptions{
-		Topic:         topicName,
-		SendTimeout:   2 * time.Second,
-		BackoffPolicy: backoff,
+		Topic:       topicName,
+		SendTimeout: 2 * time.Second,
+		BackOffPolicyFunc: func() backoff.Policy {
+			return bo
```
nodece, line 1283:
```suggestion
			return newTestBackoffPolicy(1*time.Second, 4*time.Second)
```

pulsar/dlq_router.go
nodece: DefaultBackoff will still be used?

nodece (approved):

pulsar/backoff/backoff.go
```diff
@@ -38,6 +45,13 @@ type DefaultBackoff struct {
 	backoff time.Duration
 }
 
+func NewDefaultBackoff() Policy {
+	return &DefaultBackoff{}
+}
+func NewDefaultBackoffWithBackOff(backoff time.Duration) Policy {
```
RobertIndie, line 51:
```suggestion
func NewDefaultBackoffWithInitialBackOff(backoff time.Duration) Policy {
```

Seems this name would be better?

pulsar/backoff/backoff.go
```diff
@@ -38,6 +45,13 @@ type DefaultBackoff struct {
...
+	return &DefaultBackoff{backoff: backoff}
```
RobertIndie, line 52:
```suggestion
	return &DefaultBackoff{backoff: backoff / 2}
```

I think we should treat this as the initial backoff time, right? 
If it's, then we need to first divide the backoff by 2. Because each time Next is called, the backoff is doubled first.

RobertIndie (commented):
Please help resolve the conflicts

RobertIndie (approved):

PR #1206, opened 2024-04-16
[improve] Use physical address information in connection pool key

dragosvictor:
### Motivation

Migrate https://github.com/apache/pulsar/pull/22085/ and (parts of) https://github.com/apache/pulsar-client-cpp/pull/411/ over to the Go client. Context for this idea [here](https://github.com/apache/pulsar/pull/22085/files#r1497008116).

Golang client support for blue-green migration needs the connection pool to differentiate between connections with the same logical address, but different physical addresses. Otherwise, the wrong connection might be used by the client, in effect pointing to the old cluster, instead of the new one.

### Modifications

The connection pool maintains a map of connections, keyed by their logical address and a random connection id. This PR proposes including the physical address in the key also, therefore allowing the upper layer to differentiate between connections with identical logical addresses, but different physical addresses. 

In addition to this change, the test setup had to be fixed to address breakages in `TestRetryWithMultipleHosts` and `TestReaderWithMultiHosts`. All tests in the repository are using a local standalone setup currently. This unusual configuration has broker lookup operations reply with flag `proxyThroughServiceUrl=true` ([ref](https://github.com/apache/pulsar/blob/e7c2a75473b545134a3b292ae0e87a79d65cb756/pulsar-broker/src/main/java/org/apache/pulsar/broker/lookup/TopicLookupBase.java#L369)). This in turn has the Go lookup service attempt a name resolution of the configured service addresses ([ref](https://github.com/apache/pulsar-client-go/blob/c3e94e243a730ae22d59bf9d330c4539733b7eef/pulsar/internal/lookup_service.go#L124)). The resolver picks addresses in round-robin mode. Because these two tests use a correct (reachable) broker address _and_ an unreachable address, the resolved address can end up pointing to the unreachable address. The connection pool is then corrupted with a logically invalid entry, causing the tests to fail:

| Logical Address | Physical Address | Notes |
| --------------- | ---------------- | ----- |
| reachable-broker | reachable-broker | Valid |
| unreachable-broker | unreachable-broker | Valid, but currently unusable |
| reachable-broker | unreachable-broker | *Invalid entry* |

To address the issue:
- Switch the test setup to a more common cluster configuration. File `integration-tests/clustered/docker-compose.yml` instructs how this setup should look like.
- Migrate the tests to separate files and test suites. New test files `pulsar/client_impl_clustered_test.go` and `pulsar/reader_clustered_test.go` contain Go tag `clustered`, allowing them to be ignored during the standalone test runs by virtue of the Go build process.
- Add script `run-ci-clustered.sh`, specifying the "clustered" tests to run.
- Changes in the `Makefile` add targets `make test_clustered` `make test_standalone` to run the respective test suites independently, while allowing `make test` to run all the tests, as before.
- `Dockerfile` and `run-ci.sh` are modified to run the Go build process in the container build, such that it does not need to be run again in the new `run-ci-clustered.sh` script. The image is locally consumed by the tests only and is not published, so there is no risk of contaminating users.

### Verifying this change

  - Dependencies (does it add or upgrade a dependency): _No_
  - The public API: _No_
  - The schema: _No_
  - The default values of configurations: _No_
  - The wire protocol: _No_

### Documentation

  - Does this pull request introduce a new feature? _No_
  - If yes, how is the feature documented? _Not applicable, not a feature_
  - If a feature is not applicable for documentation, explain why? _Not applicable, not a feature_
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation _Not applicable, not a feature_

### DIFF
--- a/pulsar/reader_clustered_test.go
+++ b/pulsar/reader_clustered_test.go
@@ -35,7 +35,7 @@ func TestReaderClusteredTestSuite(t *testing.T) {
 	suite.Run(t, new(ReaderClusteredTestSuite))
 }
 
-func (suite *ReaderClusteredTestSuite) TestRetryWithMultipleHosts() {
+func (suite *ReaderClusteredTestSuite) TestReaderWithMultipleHosts() {
 	req := suite.Require()
 
 	// Multi hosts included an unreached port and the actual port for verify retry logic
--- a/pulsar/reader_clustered_test.go
+++ b/pulsar/reader_clustered_test.go
@@ -1,0 +1,87 @@
+// //go:build clustered
+
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package pulsar
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type readerClusteredTestSuite struct {
+	suite.Suite
+}
+
+func ReaderClusteredTestSuite(t *testing.T) {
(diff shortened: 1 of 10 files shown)


pulsar/reader_clustered_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+type ReaderClusteredTestSuite struct {
+	suite.Suite
+}
+
+func TestReaderClusteredTestSuite(t *testing.T) {
+	suite.Run(t, new(ReaderClusteredTestSuite))
+}
+
+func (suite *ReaderClusteredTestSuite) TestRetryWithMultipleHosts() {
```
heesung-sn, line 38: Shouldn't the name be TestReader*?

dragosvictor: Good catch, fixed!

RobertIndie (approved):

PR #1208, opened 2024-04-18
[improve] PIP-307: Use assigned broker URL hints during broker reconnection

dragosvictor:

Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation


pulsar/consumer_partition.go
```diff
@@ -1358,10 +1358,13 @@ func createEncryptionContext(msgMeta *pb.MessageMetadata) *EncryptionContext {
 	return &encCtx
 }
 
-func (pc *partitionConsumer) ConnectionClosed() {
+func (pc *partitionConsumer) ConnectionClosed(closeConsumer *pb.CommandCloseConsumer) {
 	// Trigger reconnection in the consumer goroutine
 	pc.log.Debug("connection closed and send to connectClosedCh")
-	pc.connectClosedCh <- connectionClosed{}
+	pc.connectClosedCh <- &connectionClosed{
```
heesung-sn, line 1364: Can we pass one of them only by checking if tls is enabled or not here? I think we better pass one arg to the downstream funcs to simplify. (same for CommandCloseProducer)

pulsar/producer_partition.go
```diff
@@ -221,14 +221,25 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 	return p, nil
 }
 
-func (p *partitionProducer) grabCnx() error {
-	lr, err := p.client.lookupService.Lookup(p.topic)
+func (p *partitionProducer) lookupTopic(brokerURL, brokerURLTLS string) (*internal.LookupResult, error) {
+	if len(brokerURL) == 0 && len(brokerURLTLS) == 0 {
```
heesung-sn, line 225: can we move this logic to `lookupService.GetBrokerAddress` to reuse? (Then maybe we need to pass additionally pass topicName to GetBrokerAddress)

pulsar/internal/lookup_service.go
```diff
@@ -284,12 +291,24 @@ type httpLookupService struct {
 	metrics             *Metrics
 }
 
-func (h *httpLookupService) getBrokerAddress(ld *httpLookupData) (logicalAddress *url.URL,
+func (h *httpLookupService) GetBrokerAddress(brokerServiceURL, brokerServiceURLTls string,
```
heesung-sn, line 294: As I commented, can we pass one of these(brokerServiceURL, brokerServiceURLTls)?

pulsar/consumer_partition.go
dragosvictor: Done!

pulsar/producer_partition.go
dragosvictor: Done!

pulsar/internal/lookup_service.go
dragosvictor: Done!

heesung-sn (approved):

pulsar/internal/pulsar_proto/PulsarApi.pb.go
```diff
@@ -18,8 +18,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.1
-// 	protoc        v3.21.9
+// 	protoc-gen-go v1.31.0
+// 	protoc        v4.24.3
```
BewareMyPower, line 22: Could you use the same `protoc` to generate the `PulsarApi.pb.go`? Not sure if the protobuf introduces any breaking change between 3 and 4.

We'd better update the instruction here

https://github.com/apache/pulsar-client-go/blob/b4d45cd360599b52e3f160db9d39d619889c556a/pulsar/internal/pulsar_proto/README.md?plain=1#L4

to

```bash
go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.1
```

And to keep using `protoc` 3.21, here is an example on macOS:

```bash
brew install protobuf@21
/opt/homebrew/opt/protobuf@21/bin/protoc --go_out=. --go_opt=paths=source_relative PulsarApi.proto
```

pulsar/internal/lookup_service.go
```diff
@@ -30,8 +30,9 @@ import (
 
 // LookupResult encapsulates a struct for lookup a request, containing two parts: LogicalAddr, PhysicalAddr.
 type LookupResult struct {
-	LogicalAddr  *url.URL
-	PhysicalAddr *url.URL
+	IsProxyThroughServiceURL bool
```
BewareMyPower, line 33: It seems `IsProxyThroughServiceURL` is never used?

pulsar/producer_partition.go
```diff
@@ -373,10 +390,14 @@ func (p *partitionProducer) GetBuffer() internal.Buffer {
 	return b
 }
 
-func (p *partitionProducer) ConnectionClosed() {
+func (p *partitionProducer) ConnectionClosed(closeProducer *pb.CommandCloseProducer) {
```
BewareMyPower, line 393: I see `ConnectionClosed(nil)` could be called in `connection.go`, should you perform the null check here?

pulsar/internal/pulsar_proto/PulsarApi.pb.go
dragosvictor: Done! Note that it really only changed the version comments, everything else stayed the same: https://github.com/apache/pulsar-client-go/pull/1208/commits/880d732ed54d774b539d09897f273e24e5b7608f.

pulsar/producer_partition.go
dragosvictor: Good catch, fixed!

pulsar/internal/lookup_service.go
dragosvictor: You're right, it's never used outside of `lookupService.GetBrokerAddress`, where it is retrieved and consumed at the same time. Removed, thanks!

pulsar/internal/pulsar_proto/PulsarApi.pb.go
BewareMyPower: Yeah, when I regenerated it locally, I also found only the version comments were changed.

pulsar/producer_partition.go
```diff
@@ -429,12 +453,22 @@ func (p *partitionProducer) reconnectToBroker() {
 			return
 		}
 
-		if p.options.BackoffPolicy == nil {
+		var assignedBrokerURL string
+
+		if connectionClosed != nil && connectionClosed.HasURL() {
+			delayReconnectTime = 0
+			assignedBrokerURL = connectionClosed.assignedBrokerURL
+			connectionClosed = nil // Only attempt once
```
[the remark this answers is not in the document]
dragosvictor: Assigning the value to nil makes it resume the established reconnection logic if the initial connection attempt fails with the given assigned URL. I tried to capture this idea in the comment.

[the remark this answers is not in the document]
BewareMyPower: Sorry I didn't get it. Could you explain it with more details?

> if the initial connection attempt fails with the given assigned URL

Did you mean `p.grabCnx(assignedBrokerURL)` failed? If so, `reconnectToBroker` would just return and would be called when the next time `ConnectionClosed` is called.

How can assigning a local variable (`connectionClosed`) to nil just resume the established reconnection?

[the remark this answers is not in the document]
dragosvictor: There is a for loop wrapping around the reconnection attempts:

```go
	for maxRetry != 0 {
			// Producer is already closing
			p.log.Info("producer state not ready, exit reconnect")
			return
		}

		var assignedBrokerURL string

		if connectionClosed != nil && connectionClosed.HasURL() {
			delayReconnectTime = 0
		} else if p.options.BackoffPolicy == nil {
			delayReconnectTime = defaultBackoff.Next()
		} else {
			delayReconnectTime = p.options.BackoffPolicy.Next()
		}
	...
```

Setting the variable to nil guarantees we will only attempt connection to the assigned broker once. If that fails, for whatever reason, we revert back to performing topic lookups in order to locate the broker serving the topic.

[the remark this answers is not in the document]
dragosvictor: My understanding here is that if `grabCnx` fails, the loop handles the error by increasing the reconnect delay and trying up to a configurable number of times, before giving up.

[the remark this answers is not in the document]
BewareMyPower: Oh I see. I was so blind that I missed the `for maxRetry != 0` loop before. (BTW, the loop is too long)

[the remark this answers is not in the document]
BewareMyPower: How about moving the `assignedBrokerURL` out of the for loop?

```go
	var assignedBrokerURL string
	if connectionClosed != nil && connectionClosed.HasURL() {
	}

	for maxRetry != 0 {
		// ...
		if p.options.BackoffPolicy == nil {
			delayReconnectTime = defaultBackoff.Next()
		} else {
			delayReconnectTime = p.options.BackoffPolicy.Next()
		}
```

When `connectionClosed` is nil, `assignedBrokerURL` will be an empty string and `grabCnx(assignedBrokerURL)` will be equivalent with `grabCnx("")`.

The code will look more clear and less code changes are needed.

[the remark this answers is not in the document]
dragosvictor: `assignedBrokerUrl` being set to empty is what signals to `grabCnx` to execute the lookup further on ([ref](https://github.com/apache/pulsar-client-go/pull/1208/files#diff-8adb09af2175be5751840343e14df2220a2154ca4bc8de21157ed112be25b6b9R224-R239)), so it cannot be fixed outside of the loop and never changed subsequently.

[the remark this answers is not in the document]
BewareMyPower: I see, when the client failed to connect the assigned broker URL, it needs to call `grabCnx("")` for the lookup.

It makes sense to me. Thanks!

BewareMyPower (approved):

PR #1211, opened 2024-05-05
fix: fix print message id without batch index

Chenxulin97:

Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

*I found that different payload had the same message id*

### Modifications

*append batch index when call string method of MessageID*

### Verifying this change

  - Dependencies (does it add or upgrade a dependency): (**yes** / no)
  - The public API: (**yes** / no)
  - The schema: (**yes** / no / don't know)
  - The default values of configurations: (**yes** / no)
  - The wire protocol: (**yes** / no)

### Documentation

### DIFF
--- a/pulsar/impl_message.go
+++ b/pulsar/impl_message.go
@@ -184,7 +184,7 @@ func (id *messageID) BatchSize() int32 {
 }
 
 func (id *messageID) String() string {
-	return fmt.Sprintf("%d:%d:%d", id.ledgerID, id.entryID, id.partitionIdx)
+	return fmt.Sprintf("%d:%d:%d:%d", id.ledgerID, id.entryID, id.partitionIdx, id.batchIdx)
 }
 
 func deserializeMessageID(data []byte) (MessageID, error) {


pulsar/impl_message.go
```diff
@@ -184,7 +184,7 @@ func (id *messageID) BatchSize() int32 {
 }
 
 func (id *messageID) String() string {
-	return fmt.Sprintf("%d:%d:%d", id.ledgerID, id.entryID, id.partitionIdx)
+	return fmt.Sprintf("%d:%d:%d:%d", id.ledgerID, id.entryID, id.partitionIdx, id.batchIdx)
```
shoothzj, line 187: if batchIndex = -1, do we still need to print it?

RobertIndie: Let's keep it consistent with the Java client. Don't print the batch index when it's -1.

RobertIndie (commented):
Please add the unit test case for it.

pulsar/impl_message.go
Chenxulin97: ok

shoothzj:
@RobertIndie, do we really need a unit test case for the `String()` method? In my opinion, simply testing the `String()` method has less value.

RobertIndie:
> do we really need a unit test case for the `String()` method? In my opinion, simply testing the `String()` method has less value.

I think it's worth adding these tests. This is a public method exposed to users, and we need to ensure the correctness of the returned content. Moreover, other tests also rely on this String's output. We can see that this change has caused failures in other tests in the CI.

shoothzj:
> > do we really need a unit test case for the `String()` method? In my opinion, simply testing the `String()` method has less value.
> 
> I think it's worth adding these tests. This is a public method exposed to users, and we need to ensure the correctness of the returned content. Moreover, other tests also rely on this String's output. We can see that this change has caused failures in other tests in the CI.

I understand the importance of writing test cases, and I won't insist on not writing them. However, I'd like to note that this part of the code logic is quite clear and straightforward.

RobertIndie:
> However, I'd like to note that this part of the code logic is quite clear and straightforward.

We write tests not because the code is simple or complex, but to ensure the functionality behaves correctly. Even for simple code, I believe it's best to write tests to cover it. This is exactly why the metric of test coverage is important.

shoothzj:
@RobertIndie Fair enough, thanks for your explanation. I won't block this review.

geniusjoe:
I think the unit test case ran fail is because `TestGetMessagesByID` function in `pulsaradmin/pkg/admin/subscription_test.go` use message `id.String()` as a map key to check if current messages are in a same batch. 
When we change `id.String()` from `fmt.Sprintf("%d:%d:%d", id.ledgerID, id.entryID, id.partitionIdx)`
	to `fmt.Sprintf("%d:%d:%d:%d", id.ledgerID, id.entryID, id.partitionIdx, id.batchIdx)`, map key changed, so that the test case cannot distinguish messages in the same batch.

```
	for i := 0; i <= numberMessages; i++ {
		producer.SendAsync(ctx, &pulsar.ProducerMessage{
		}, func(id pulsar.MessageID, message *pulsar.ProducerMessage, err error) {
			assert.Nil(t, err)
			messageIDMap[id.String()]++
			wg.Done()
		})
	}
```

If we need to fix test case, we can replace `messageIDMap[id.String()]` to `messageIDMap[fmt.Sprintf("%d:%d:%d", id.ledgerID, id.entryID, id.partitionIdx)]`, but I think this pr is an incompatibility change, so it should be greater to add change log in documentation.

Chenxulin97:
It seems more works than I expect, I have no time to work on this.

PR #1221, opened 2024-05-22
[Improve] PIP-313 Add GetLastMessageIDs API

crossoverJie:
### Motivation
To keep consistent with the Java client.

Releted PR: https://github.com/apache/pulsar/pull/20040

### Modifications

- Add `GetLastMessageIDs`api for consumer.

### Verifying this change

  - The public API: (yes)
  - The schema: (no)
  - The wire protocol: (yes)

### Documentation

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -281,7 +281,7 @@ type Consumer interface {
 	// GetLastMessageIDs get all the last message id of the topics the consumer subscribed.
 	//
 	// The list of MessageID instances of all the topics that the consumer subscribed
-	GetLastMessageIDs() ([]MessageID, error)
+	GetLastMessageIDs() ([]*TopicMessageID, error)
 
 	// Receive a single message.
 	// This calls blocks until a message is available.
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -278,6 +278,11 @@ type Consumer interface {
 	// where more than one consumer are currently connected.
 	UnsubscribeForce() error
 
+	// GetLastMessageIDs get all the last message id of the topics the consumer subscribed.
+	//
+	// The list of MessageID instances of all the topics that the consumer subscribed
+	GetLastMessageIDs() ([]MessageID, error)
+
 	// Receive a single message.
 	// This calls blocks until a message is available.
 	Receive(context.Context) (Message, error)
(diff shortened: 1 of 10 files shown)


pulsar/consumer.go
```diff
@@ -278,6 +278,11 @@ type Consumer interface {
 	// where more than one consumer are currently connected.
 	UnsubscribeForce() error
 
+	// GetLastMessageIDs get all the last message id of the topics the consumer subscribed.
+	//
+	// The list of MessageID instances of all the topics that the consumer subscribed
+	GetLastMessageIDs() ([]MessageID, error)
```
RobertIndie, line 284: We need to implement a TopicMesasgeID and return TopicMessageIDs here. Otherwise, the users can't know which  messages ids belong to which topic.

crossoverJie: I was careless, done.

pulsar/message.go
```diff
@@ -166,6 +166,12 @@ type MessageID interface {
 	String() string
 }
 
+// TopicMessageID defined the correspondence between topic and MessageID.
+type TopicMessageID struct {
```
RobertIndie, line 170: Is it better to make it as an interface? And I think it should also implement the `MessageID` interface.

crossoverJie: Good suggestion, keep the external API consistent.

pulsar/consumer_multitopic_test.go
```diff
@@ -137,3 +137,69 @@ func TestMultiTopicConsumerForceUnsubscribe(t *testing.T) {
...
+		p.Close()
+		if err != nil {
+			assert.Nil(t, err)
+		}
+	}
+
+	topicMessageIDs, err := consumer.GetLastMessageIDs()
+	assert.Nil(t, err)
+	assert.Equal(t, topic1Partition+topic2Partition+topic3Partition, len(topicMessageIDs))
+	for _, id := range topicMessageIDs {
```
RobertIndie, line 195: If the topicMessageIDs contains the wrong topic name, the test would be bypassed.
We need to assert that the topicMessageID covers all the topics.

Same for other test cases.

RobertIndie (commented):
The implementation looks good to me. 👍 
Left one comment for the test.

pulsar/consumer_multitopic_test.go
RobertIndie: A better approach is to use a map to store `topic1`, `topic2` and `topic3` and their corresponding messages count. So that we can assert if the `id` here matches one of them.

RobertIndie (approved):

crossoverJie:
@RobertIndie The integration test failed, maybe we can try it again, it passes in my own repository.

https://github.com/crossoverJie/pulsar-client-go/actions/runs/9253419225/job/25463657165

PR #1225, opened 2024-05-30
[Issue 1223] Support ZeroQueueConsumer

### LINKED ISSUE #1223: [Question]Zero value of ReceiverQueueSize
For php client was discussed [#26](https://github.com/ikilobyte/pulsar-client-php/issues/26)

According to implementation
```go
if options.ReceiverQueueSize <= 0 {
	options.ReceiverQueueSize = 1000
}
```
According to [recommendations](https://pulsar.apache.org/docs/next/cookbooks-message-queue/#client-configuration-changes):
"...set the consumers' receiver queue size very low (potentially even to **_0_** if necessary)..."

So according to docs, I need set this value to 0, but go client sets it to 1000.

Please clarify

crossoverJie:
Fixes #1223

### Motivation

Support ZeroQueueConsumer, refer to Java [ZeroQueueConsumerImpl](https://github.com/apache/pulsar/blob/8c50a6c2e91c81dbf187ce5e66cb39e2758a741e/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java#L42)

### Modifications

- The consumer add a new optional parameter `EnableZeroQueueConsumer`
- Add a new `zeroQueueConsumer`

### Verifying this change

  - The public API: (yes)
  - The schema: (no)
  - The default values of configurations: (yes)
  - The wire protocol: (no)

### Documentation


crossoverJie:
@RobertIndie PTAL, thanks.

pulsar/consumer_impl.go
```diff
@@ -236,7 +240,19 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
...
+	partitions, err := client.TopicPartitions(topic)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(partitions) > 1 && options.EnableZeroQueueConsumer {
+		return nil, pkgerrors.New("ZeroQueueConsumer is not supported for partitioned topics")
+	}
+
+	if len(partitions) == 1 && options.EnableZeroQueueConsumer {
```
RobertIndie, line 253: We should also prevent creating a zero queue consumer for a partitioned topic that has only one partition.

pulsar/consumer_zero_queue.go
```diff
@@ -0,0 +1,328 @@
...
+		decryption:                  zc.options.Decryption,
+		ackWithResponse:             zc.options.AckWithResponse,
+		maxPendingChunkedMessage:    zc.options.MaxPendingChunkedMessage,
+		expireTimeOfIncompleteChunk: zc.options.ExpireTimeOfIncompleteChunk,
+		autoAckIncompleteChunk:      zc.options.AutoAckIncompleteChunk,
+		consumerEventListener:       zc.options.EventListener,
+		enableBatchIndexAck:         zc.options.EnableBatchIndexAcknowledgment,
+		ackGroupingOptions:          zc.options.AckGroupingOptions,
+		autoReceiverQueueSize:       zc.options.EnableAutoScaledReceiverQueueSize,
+	}
```
RobertIndie, line 107: Could we abstract this logic so that we can reuse the code? https://github.com/apache/pulsar-client-go/blob/fc15995c1a180a151aca58c478bc40843120783e/pulsar/consumer_impl.go#L390-L420

pulsar/consumer_zero_queue.go
```diff
@@ -0,0 +1,328 @@
...
+			return cm.Message, nil
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+	}
+
+}
+
+func (z *zeroQueueConsumer) Chan() <-chan ConsumerMessage {
+	z.log.Warnf("zeroQueueConsumer cannot support Chan method")
```
RobertIndie, line 174: Better to use error level log here.

What about panicking here?

pulsar/consumer_zero_queue_test.go
```diff
@@ -0,0 +1,443 @@
...
+	})
+
+	assert.Nil(t, err)
+	defer client.Close()
+
+	topic := newTopicName()
+	ctx := context.Background()
+
+	// create consumer
+	consumer, err := client.Subscribe(ConsumerOptions{
```
RobertIndie, line 46: Can we check the type for the consumer?

pulsar/consumer_zero_queue.go
crossoverJie: Thanks, done.

RobertIndie (approved):

PR #1228, opened 2024-06-25
fix dynamic config

labuladong:
If users use a custom dynamic configuration, it's better not to do any escaping for the config value. Let the user fill it out manually.

The original implementation will escape the value twice, the first time is here:

https://github.com/apache/pulsar-client-go/blob/a1ce5e6d48fd00d07d45958a5c7482751f0cbb68/pulsaradmin/pkg/admin/brokers.go#L123-L123

the second time is here:

https://github.com/apache/pulsar-client-go/blob/a1ce5e6d48fd00d07d45958a5c7482751f0cbb68/pulsaradmin/pkg/admin/admin.go#L107-L109

### DIFF
--- a/pulsaradmin/pkg/admin/admin.go
+++ b/pulsaradmin/pkg/admin/admin.go
@@ -112,3 +112,8 @@ func (c *pulsarClient) endpoint(componentPath string, parts ...string) string {
 		path.Join(escapedParts...),
 	)
 }
+
+// this function won't do any escaping
+func (c *pulsarClient) endpointWithFullPath(componentPath string, fullPath string) string {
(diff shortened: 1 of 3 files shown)


pulsaradmin/pkg/admin/admin.go
```diff
@@ -112,3 +112,8 @@ func (c *pulsarClient) endpoint(componentPath string, parts ...string) string {
 		path.Join(escapedParts...),
 	)
 }
+
+// this function won't do any escaping
+func (c *pulsarClient) endpointWithFullPath(componentPath string, fullPath string) string {
```
Demogorgon314, line 117: Can we add some tests to cover this fix?

crossoverJie (approved):
LGTM

Demogorgon314 (approved):
Good catch!

Issue #1191, opened 2024-03-01
allow override the Schema Creator by a consumer


adrianiacobghiula:
**Is your feature request related to a problem? Please describe.**
A consumer should be allowed to select / choose / override the default Schema. For example select to use https://github.com/hamba/avro and not github.com/linkedin/goavro/v2 

**Describe the solution you'd like**
Allow the consumer in Options to set a different function as SchemaCreator

RobertIndie:
Can you highlight the benefits of using https://github.com/hamba/avro? Given that https://github.com/linkedin/goavro has yet to see updates in 2 years, it might be worthwhile considering replacing it with https://github.com/hamba/avro if we have compelling reasons.

adrianiacobghiula:
@RobertIndie  thanks for looking at the PR 
I choose to use hamba because of avro to golang struct generation -> see avrogen.
Easier handling of "nullable" union. by having a field as pointer. 
benchmark shows to be faster than linkedin.
There is a minus that hopefully with later versions can be covered: recursive structs not not covered.

PR #1230, opened 2024-06-26
[refactor][schema] replace linkedin/goavro/v2 with hamba/avro/v2

adrianiacobghiula:
### Motivation

Allow direct deserialization in go struct, allow generation of go struct using avro schema <- support offered by hamba/avro
Improved support in hamba avro project, while linkedin has limited support since 2022 

### Modifications

Replaced linkedin/goavro/v2 with hamba/avro/v2

### Verifying this change

  - The public API: (no)
  - The schema: ( no )
  - The wire protocol: (no)

### Documentation


adrianiacobghiula:
Starting from #1191

README.md
```diff
@@ -38,7 +38,7 @@ CGo-based library.
 
 ## Requirements
 
-- Go 1.18+
+- Go 1.20+
```
crossoverJie, line 41: This is a breaking change, maybe we need more discussion.

adrianiacobghiula: this is from the required upgrade of github.com/klauspost/compress v1.17.9

nodece: Usually, we want to be compatible with lower versions of Golang, but some libraries force us to upgrade to higher versions.
 
Using 1.20 is in accordance with Golang's release policy: https://go.dev/doc/devel/release

We will release 0.13.0, and I think we can make this change.

pulsar/schema.go
```diff
@@ -125,20 +126,19 @@ func NewSchema(schemaType SchemaType, schemaData []byte, properties map[string]s
 }
 
 type AvroCodec struct {
-	Codec *goavro.Codec
+	Codec avro.Schema
 }
 
-func NewSchemaDefinition(schema *goavro.Codec) *AvroCodec {
-	schemaDef := &AvroCodec{
+func NewSchemaDefinition(schema avro.Schema) *AvroCodec {
```
nodece, line 132: I worry that this method will be called on the outside(users).

adrianiacobghiula: Should this function not be exported ?

nodece: This parameter has been changed to `avro.Schema`.

README.md
RobertIndie: Yes. According to the go release policy:

> Each major Go release is supported until there are two newer major releases.

The currently supported go version is 1.22 and 1.21. We are OK to upgrade the lowest supported go version here.

pulsar/schema.go
```diff
@@ -124,21 +125,10 @@ func NewSchema(schemaType SchemaType, schemaData []byte, properties map[string]s
 	return
 }
 
-type AvroCodec struct {
-	Codec *goavro.Codec
-}
-
-func NewSchemaDefinition(schema *goavro.Codec) *AvroCodec {
```
RobertIndie, line 131: This is a breaking change. Could we avoid it?

Is it possible to allow users to continue using goavro?

adrianiacobghiula: we could, I guess :) -> but that would mean that there is dependency on both linkedin/goavro & hamba/avro 
using hamba internally and just have goavro.

personally I see no benefit of a function that does ... nothing

adrianiacobghiula: @RobertIndie keeping `NewSchemaDefinition` function & `AvroCodec` struct with added Deprecation

nodece:
Rethinking recent changes,  due to you changing some method parameters, it appears that there is a breaking change, which resulted in me making an incorrect review.

In the old design, the codec was exported in the schema, and which was created by NewSchemaDefinition(should be internal).

Usually, the users use the following methods to create the avro schema:
```go
NewAvroSchema()
NewAvroSchemaWithValidation()
```

So that we can remove NewSchemaDefinition and AvroCodec.

The next version is the minor version, we can make this changes.

BTW, newXXSchema method returns an implement, not an interface, we also need to improve that in the future.

Is this a fair reason?

adrianiacobghiula:
@nodece fair & done

nodece (approved):

pulsar/schema.go
```diff
@@ -124,21 +124,10 @@ func NewSchema(schemaType SchemaType, schemaData []byte, properties map[string]s
...
```
RobertIndie, line 131: Even though the client is still in the development version (0.x.y), it is widely used in production. We should avoid introducing breaking changes in the v0.x version.

We may still need to support goavro here but marked it with deprecation. We can make these breaking changes and remove it in v1.0.

RobertIndie: Ok. Now I'm +1.

RobertIndie (approved):

PR #1237, opened 2024-07-05
Fix transaction coordinator client cann't reconnect to the broker

RobertIndie:

Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/transaction_coordinator_client.go
+++ b/pulsar/transaction_coordinator_client.go
@@ -19,7 +19,10 @@ package pulsar
 
 import (
 	"context"
+	"github.com/pkg/errors"
+	uAtomic "go.uber.org/atomic"
 	"strconv"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pulsar/transaction_test.go
```diff
@@ -539,3 +540,56 @@ func TestAckChunkMessage(t *testing.T) {
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(1 * time.Second):
+				connections := cli.cnxPool.GetConnections()
+				for _, conn := range connections {
+					conn.Close()
+				}
```
BewareMyPower, line 575: The only difference with `ConnectionPool.Close` is it does not remove the connection from the pool. Could you call `cli.cnxPool.Close()` here instead to avoid adding a `GetConnections()` method just for tests?

pulsar/transaction_coordinator_client.go
```diff
@@ -20,23 +20,330 @@ package pulsar
...
+		RequestId:     proto.Uint64(requestID),
+		TcId:          proto.Uint64(nextTcID),
+		TxnTtlSeconds: proto.Uint64(uint64(op.timeout.Milliseconds())),
+	}
+	res, err := t.tc.client.rpcClient.RequestOnCnx(t.getConn(), requestID, pb.BaseCommand_NEW_TXN, cmdNewTxn)
+	if t.checkRetriableError(err, op) {
+		return
+	}
+	defer close(op.done)
+	defer t.tc.semaphore.Release()
```
shibd, line 228: If my understand right.

This semaphore will be released 2 times for the same operation.

https://github.com/apache/pulsar-client-go/pull/1237/files#diff-d923c7a298876f490ae5b4b04c712deae3da2afa49cc5161e1c37e70e9822f73R394

Other method have the same issue.

pulsar/transaction_coordinator_client.go
```diff
@@ -20,23 +20,330 @@ package pulsar
...
+	// Request
+	timeout time.Duration
+
+	// Response
+	done  chan any
+	err   error
+	txnID *TxnID
+}
+
+func (t *transactionHandler) newTransaction(op *newTxnOp) {
```
shibd, line 215: We should consider renaming these types of methods to distinguish the external ones.

For example:

- **internalXX**
- **retriableXX**

pulsar/transaction_coordinator_client.go
RobertIndie: Good catch. Yes. I will fix them.

pulsar/transaction_coordinator_client.go
RobertIndie: They are implemented under different structs. It can be distinguished by the struct.

shibd (approved):

BewareMyPower (changes requested):
This PR might be over complicated.

IIUC, the root cause is that when `RequestOnCnx` returns an error that indicates the connection is closed, it just propagate to the caller side. Here is a simple fix (without any backoff mechanism and timeout):

```go
func (tc *transactionCoordinatorClient) sendRequest(cnxIndex uint64, requestID uint64, cmdType pb.BaseCommand_Type, message proto.Message) error {
        for {
                res, err := tc.client.rpcClient.RequestOnCnx(tc.cons[cnxIndex], requestID, cmdType, message)
                if err != nil {
                        if err.Error() != "connection closed" {
                                return err
                        }
                        if err = tc.grabConn(cnxIndex); err != nil {
                                return err
                        }
                        continue
                }
                if res.Response.EndTxnResponse.Error != nil {
                        return getErrorFromServerError(res.Response.EndTxnResponse.Error)
                }
                return nil
        }
}
```

Then replace the `RequestOnCnx` call like:

```go
        err := tc.sendRequest(id.MostSigBits, requestID, pb.BaseCommand_END_TXN, cmdEndTxn)
        tc.semaphore.Release()
        if err != nil {
                return err
        }
        return nil
```

The fix above will bring much less code changes and the code will look more simple.

pulsar/error.go
```diff
@@ -121,6 +121,15 @@ const (
...
 	ProducerFenced
+	// MaxConcurrentOperationsReached indicates that the maximum number of concurrent operations
+	// has been reached. This means that no additional operations can be started until some
+	// of the current operations complete.
+	MaxConcurrentOperationsReached
+	// TransactionCoordinatorNotEnabled indicates that the transaction coordinator is not enabled.
+	// This error is returned when an operation that requires the transaction coordinator is attempted
+	// but the transaction coordinator feature is not enabled in the system or the transaction coordinator
+	// has not initialized
+	TransactionCoordinatorNotEnabled
```
BewareMyPower, line 132: These new error codes and related changes are also unrelated to this PR. You'd better open a new PR for it.

That's also a disadvantage of such a huge PR because unrelated changes can easily hide in the changes.

RobertIndie:
> The fix above will bring much less code changes and the code will look more simple.

This implementation misses some key details. For example, it lacks a backoff mechanism and connection timeout logic as you said. Also, what happens if many goroutines call `grabConn` at the same time? We might end up with multiple goroutines trying to handle reconnections simultaneously. This doesn't seem like a good approach. Although we could use locks or other methods to manage concurrent executions, there are still issues. What if there's a connection timeout and we need to notify the user? Plus, how do we inform other goroutines to stop attempting reconnections in that case?  It seems we need complex and less intuitive logic to address these issues.

IIUC, the main simplification in this implementation involves transfering the XXOp and the event loop logic, right? The key difference appears to be the use of the event loop model. However, reducing this complexity doesn't seem very significant to me. Both producers and consumers already use the event loop for handling requests.
https://github.com/apache/pulsar-client-go/blob/b0111a2dc473317e4b41ab4af09e42b09a72384c/pulsar/consumer_partition.go#L1578-L1582

The Java implementation of the TransactionMetaStoreHandler also uses something like event loop, similar to Go:
[TransactionMetaStoreHandler Java code](https://github.com/apache/pulsar/blob/83b86abcb74595d7e8aa31b238a7dbb19a04dde2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TransactionMetaStoreHandler.java#L220-L226)

Moreover, without the event loop, it seems challenging to implement asynchronous methods like `XXXAsync`. For example, [Producer.SendAsync](https://github.com/apache/pulsar-client-go/blob/2a28e21c59d005515e118fed5bf8f333d6699e39/pulsar/producer.go#L230). Although we don't have this feature yet, it would be beneficial to make these operations truly asynchronous in the future. The Java client already supports this feature:
[Async support in Java client](https://github.com/apache/pulsar/blob/4e5c0bcc2b44c33a966287b86c2c235be249dc51/pulsar-client/src/main/java/org/apache/pulsar/client/impl/transaction/TransactionCoordinatorClientImpl.java#L176)

This PR mainly adopts the eventloop patterns of the the existing producer and consumer. Maybe it's better to start a separate discussion for this.

pulsar/error.go
RobertIndie: This was introduced while I was fixing this issue. I agree to move it to a separate PR.

BewareMyPower:
I see, currently the `Context` object is also not used so we need an event loop to handle it.

pulsar/transaction_coordinator_client.go
```diff
@@ -20,23 +20,322 @@ package pulsar
...
+}
+
+type addPublishPartitionOp struct {
+	// Request
+	id         *TxnID
+	partitions []string
+
+	// Response
+	done chan any
+	err  error
```
BewareMyPower, line 241: Why not use `chan error` as the response? Then you can replace `close(op.done)` with 

```go
	if err != nil {
		op.errorCh <- err
	} else if res.Response.AddPartitionToTxnResponse.Error != nil {
		op.errorCh <- getErrorFromServerError(res.Response.AddPartitionToTxnResponse.Error)
	} else {
		op.errorCh <- nil
	}
```

and you just need to return the value from the channel:

```go
return <-op.errorCh 
```

The same question is applied for other `xxxOp` structs.

pulsar/transaction_coordinator_client.go
```diff
@@ -20,23 +20,322 @@ package pulsar
...
+				t.newTransaction(r)
+			case *addPublishPartitionOp:
+				t.addPublishPartitionToTxn(r)
+			case *addSubscriptionOp:
+				t.addSubscriptionToTxn(r)
+			case *endTxnOp:
+				t.endTxn(r)
+			}
+		case <-t.connectClosedCh:
+			t.log.Infof("Transaction handler %d will reconnect to the transaction coordinator", t.partition)
```
BewareMyPower, line 143: This log seems redundant because you already log "Reconnect to broker" in `reconnectToBroker`. You can add the partition info into that log and remove this line.

```suggestion
```

pulsar/transaction_coordinator_client.go
```diff
@@ -20,23 +20,322 @@ package pulsar
...
+		op.err = getErrorFromServerError(res.Response.EndTxnResponse.Error)
+	}
+}
+
+func (t *transactionHandler) close() {
+	if t.getState() != txnHandlerReady {
+		return
+	}
+	close(t.closeCh)
+	t.setState(txnHandlerClosed)
```
BewareMyPower, line 336:
```suggestion
	if t.state.CAS(txnHandlerReady, txnHandlerClosed) {
		close(t.closeCh)
	}
```

Use a thread safe implementation. (then you can remove `setState`)

BewareMyPower (approved):

PR #1210, opened 2024-05-05
[feat] PIP-188 Support blue-green migration

heesung-sn:
Master Issue: https://github.com/apache/pulsar/issues/16551

### Motivation

- Support blue-green migration pip-188 for golang client

### Modifications

- added blue-green client logic 
- register the producer instance in the producers map before sending produce creation command. This is required since broker could send topic migration command in the middle of creating the producer.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

### Does this pull request potentially affect one of the following parts:

*If `yes` was chosen, please highlight the changes*

  - Dependencies (does it add or upgrade a dependency): (yes / no)
  - The public API: (yes / no)
  - The schema: (yes / no / don't know)
  - The default values of configurations: (yes / no)
  - The wire protocol: (yes / no)

### Documentation

  - Does this pull request introduce a new feature? (yes / no)
  - If yes, how is the feature documented? (not applicable / docs / GoDocs / not documented)
  - If a feature is not applicable for documentation, explain why?
  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

PR #1243, opened 2024-07-11
fix: fix producer connection

nodece:
### Motivation

https://github.com/apache/pulsar-client-go/pull/1210 breaks the producer feature.

When the `MaxConnectionsPerBroker` is greater than 1, the client will create multiple connections, the producer cannot use the correct connect to send the message to the broker:

```
make build
./bin/pulsar-perf produce test01 -r 100 --profile  --max-connections 30 --profile --service-url pulsar://localhost:6650 -q 100 --batching-time 10

INFO[16:40:23.991] Connecting to broker                          remote_addr="pulsar://localhost:6650"
INFO[16:40:23.994] TCP connection established                    local_addr="[::1]:59230" remote_addr="pulsar://localhost:6650"
INFO[16:40:23.996] Connection is ready                           local_addr="[::1]:59230" remote_addr="pulsar://localhost:6650"
INFO[16:40:24.002] Connecting to broker                          remote_addr="pulsar://localhost:6650"
INFO[16:40:24.003] TCP connection established                    local_addr="[::1]:59231" remote_addr="pulsar://localhost:6650"
INFO[16:40:24.007] Connection is ready                           local_addr="[::1]:59231" remote_addr="pulsar://localhost:6650"
INFO[16:40:24.009] Connecting to broker                          remote_addr="pulsar://localhost:6650"
INFO[16:40:24.010] TCP connection established                    local_addr="[::1]:59232" remote_addr="pulsar://localhost:6650"
INFO[16:40:24.012] Connection is ready                           local_addr="[::1]:59232" remote_addr="pulsar://localhost:6650"
INFO[16:40:24.012] Connecting to broker                          remote_addr="pulsar://localhost:6650"
INFO[16:40:24.012] TCP connection established                    local_addr="[::1]:59233" remote_addr="pulsar://localhost:6650"
INFO[16:40:24.015] Connection is ready                           local_addr="[::1]:59233" remote_addr="pulsar://localhost:6650"
INFO[16:40:24.150] Connected producer                            cnx="[::1]:59233 -> [::1]:6650" epoch=0 topic="persistent://public/default/test01"
INFO[16:40:24.150] Created producer                              cnx="[::1]:59232 -> [::1]:6650" producerID=1 producer_name=standalone-37-3 topic="persistent://public/default/test01"
WARN[16:40:24.153] Connection was closed                         cnx="[::1]:59232 -> [::1]:6650" producerID=1 producer_name=standalone-37-3 topic="persistent://public/default/test01"
INFO[16:40:24.153] runEventsLoop will reconnect in producer      producerID=1 producer_name=standalone-37-3 topic="persistent://public/default/test01"
````

You can notice that there are two connections,  the producer was created by the `[::1]:59233` connection, not `[::1]:59232`.
```
INFO[16:40:24.150] Connected producer                            cnx="[::1]:59233 -> [::1]:6650" epoch=0 topic="persistent://public/default/test01"
INFO[16:40:24.150] Created producer                              cnx="[::1]:59232 -> [::1]:6650" producerID=1 producer_name=standalone-37-3 topic="persistent://public/default/test01"
```

When a message was sent to the broke by `[::1]:59232`, this connection will be closed, because the connection didn't create the producer. 

### Modifications

- Fix get connection error, if err is not nil, we must to return err, not ignore.
- Use `p.client.rpcClient.RequestOnCnx` instead of ` p.client.rpcClient.Request` to create the producer, when create fails, we need to remove the connection listener.
- The `rpcClient.Request` method should call the `rpcClient.RequestOnCnx`, and we need to handle the `roducerAccessModeWaitForExclusive` case.

### Verifying this change

Added test.

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -301,7 +301,7 @@ func (p *partitionProducer) grabCnx(assignedBrokerURL string) error {
 		p.log.WithError(err).Error("Failed to create producer at send PRODUCER request")
 		if errors.Is(err, internal.ErrRequestTimeOut) {
 			id := p.client.rpcClient.NewRequestID()
-			_, _ = p.client.rpcClient.RequestOnCnx(res.Cnx, id, pb.BaseCommand_CLOSE_PRODUCER,
+			_, _ = p.client.rpcClient.RequestOnCnx(cnx, id, pb.BaseCommand_CLOSE_PRODUCER,
 				&pb.CommandCloseProducer{
 					ProducerId: &p.producerID,
 					RequestId:  &id,
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -292,12 +292,12 @@ func (p *partitionProducer) grabCnx(assignedBrokerURL string) error {
 		}
 	}
 
-	res, err := p.client.rpcClient.Request(lr.LogicalAddr, lr.PhysicalAddr, id, pb.BaseCommand_PRODUCER, cmdProducer)
+	res, err := p.client.rpcClient.RequestOnCnx(cnx, id, pb.BaseCommand_PRODUCER, cmdProducer)
 	if err != nil {
 		p.log.WithError(err).Error("Failed to create producer at send PRODUCER request")
 		if errors.Is(err, internal.ErrRequestTimeOut) {
 			id := p.client.rpcClient.NewRequestID()
-			_, _ = p.client.rpcClient.Request(lr.LogicalAddr, lr.PhysicalAddr, id, pb.BaseCommand_CLOSE_PRODUCER,
+			_, _ = p.client.rpcClient.RequestOnCnx(res.Cnx, id, pb.BaseCommand_CLOSE_PRODUCER,
 				&pb.CommandCloseProducer{
 					ProducerId: &p.producerID,
 					RequestId:  &id,
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -294,6 +294,7 @@ func (p *partitionProducer) grabCnx(assignedBrokerURL string) error {
 
 	res, err := p.client.rpcClient.RequestOnCnx(cnx, id, pb.BaseCommand_PRODUCER, cmdProducer)
 	if err != nil {
+		p._getConn().UnregisterListener(p.producerID)
 		p.log.WithError(err).Error("Failed to create producer at send PRODUCER request")
 		if errors.Is(err, internal.ErrRequestTimeOut) {
 			id := p.client.rpcClient.NewRequestID()
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -284,12 +284,15 @@ func (p *partitionProducer) grabCnx(assignedBrokerURL string) error {
 
 	cnx, err := p.client.cnxPool.GetConnection(lr.LogicalAddr, lr.PhysicalAddr)
 	// registering the producer first in case broker sends commands in the middle
-	if err == nil {
-		p._setConn(cnx)
-		err = p._getConn().RegisterListener(p.producerID, p)
-		if err != nil {
-			p.log.WithError(err).Errorf("Failed to register listener: {%d}", p.producerID)
-		}
+	if err != nil {
+		p.log.Error("Failed to get connection")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pulsar/producer_test.go [hunk 1]
```diff
@@ -2474,3 +2476,33 @@ func TestDisableReplication(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, []string{"__local__"}, msgMetadata.GetReplicateTo())
 }
+
+func TestProducerWithMaxConnectionsPerBroker(t *testing.T) {
```
RobertIndie, line 2480: The change looks good to me. But how this test case reproduce the issues?

crossoverJie:
```go
func TestSimpleProducer(t *testing.T) {
		URL: serviceURL,
		MaxConnectionsPerBroker: 2,
	})
	assert.NoError(t, err)
	defer client.Close()

		Topic: newTopicName(),
	})

	assert.NoError(t, err)
	assert.NotNil(t, producer)
	defer producer.Close()

	for i := 0; i < 10; i++ {
		ID, err := producer.Send(context.Background(), &ProducerMessage{
			Payload: []byte("hello"),
		})

		assert.NoError(t, err)
		assert.NotNil(t, ID)
	}

	
}
```
Under this test scenario, the issue can be triggered as long as the `MaxConnectionsPerBroker` is greater than 1.

Because the connection obtained here:
https://github.com/apache/pulsar-client-go/blob/50dce7ea7c2d7aff64c65ccbc2ef696a25cc9afc/pulsar/producer_partition.go#L285-L289

is not the same as the connection obtained there:
https://github.com/apache/pulsar-client-go/blob/50dce7ea7c2d7aff64c65ccbc2ef696a25cc9afc/pulsar/producer_partition.go#L295


This will lead to the following exception:

```log
time="2024-07-12T09:52:25+08:00" level=error msg="Failed to create producer at send PRODUCER request" error="server error: ProducerBusy: org.apache.pulsar.broker.service.BrokerServiceException$NamingException: Producer with name 'standalone-6-9' is already connected to topic 'persistent://public/default/my-topic-348398000'" producerID=1 producer_name=standalone-6-9 topic="persistent://public/default/my-topic-348398000"
time="2024-07-12T09:52:25+08:00" level=error msg="Failed to create producer at reconnect" error="server error: ProducerBusy: org.apache.pulsar.broker.service.BrokerServiceException$NamingException: Producer with name 'standalone-6-9' is already connected to topic 'persistent://public/default/my-topic-348398000'" producerID=1 producer_name=standalone-6-9 topic="persistent://public/default/my-topic-348398000"
```

crossoverJie (approved):
LGTM, but the CI has failed.

pulsar/producer_test.go
RobertIndie: Thanks. I get it now.

RobertIndie (approved):

RobertIndie:
This is a regression bug fix. I will include it into the 0.13.0

pulsar/producer_partition.go
```diff
@@ -292,12 +292,12 @@ func (p *partitionProducer) grabCnx(assignedBrokerURL string) error {
 		}
 	}
 
-	res, err := p.client.rpcClient.Request(lr.LogicalAddr, lr.PhysicalAddr, id, pb.BaseCommand_PRODUCER, cmdProducer)
+	res, err := p.client.rpcClient.RequestOnCnx(cnx, id, pb.BaseCommand_PRODUCER, cmdProducer)
```
heesung-sn, line 295: Dont we need to handle the case when cnx is null, here?

crossoverJie: Perhaps we should add error handle in line 286, so we don't have to check for the `cnx` is nil.

nodece: When the err is nil, the cnx should be non-null. This is a convention.

nodece: Fixed.

nodece:
When the `test_extensible_load_manager` and `blue-green migration test` fail, it is difficult for us to debug the code, because this test runs on the container, not host.

crossoverJie (approved):
👍Good job.

RobertIndie (approved):

PR #1244, opened 2024-07-11
[Fix] PulsarCtl 1266  Oauth2 Client credentials flow use scopes from the keyfile as well

Nikolajls:

Fixes https://github.com/streamnative/pulsarctl/issues/1266 which uses the code being touched here.

### Motivation

As the issue shows when using Pulsarctl which works with a context as configuration style, Oauth2 is used under the hood.

However the library does not expose any way to inject the context configuration `scope` value, and then relies on reading that from the `keyFile` for `client_credentials flow`.

However that is not being utilized in the current code as the scope value is not read from from the file.

To allow the usage of oauth2 with a keyfile in Pulsarctl 3+ which it doesn't right now

### Modifications

Alters so reading the keyFile for client credentials also returns the `scope` value.

The value is then split on spaces, and added to a temporary slice. 
After that the value of any additionalScopes that was already in the options is then added to the temp slice as well.
In the end the additionalScopes property on the options is set to the temp slice containing values from previous additionalScopes and the ones from the keyFile.

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/oauth2/client_credentials_provider_test.go
+++ b/oauth2/client_credentials_provider_test.go
@@ -34,12 +34,14 @@ func TestNewClientCredentialsProviderFromKeyFile(t *testing.T) {
 	ClientSecret := "CLIENT_SECRET"
 	ClientEmail := "CLIENT_EMAIL"
 	IssuerURL := "ISSUER_URL"
+	Scope := "SCOPE"
 	keyFile := &KeyFile{
 		Type:         oauthType,
 		ClientID:     clientID,
 		ClientSecret: ClientSecret,
 		ClientEmail:  ClientEmail,
 		IssuerURL:    IssuerURL,
+		Scope:        Scope,
 	}
 
 	b, err := json.Marshal(keyFile)
--- a/oauth2/client_credentials_flow.go
+++ b/oauth2/client_credentials_flow.go
@@ -80,19 +80,6 @@ func NewDefaultClientCredentialsFlow(options ClientCredentialsFlowOptions) (*Cli
 	}
 
 	tokenRetriever := NewTokenRetriever(&http.Client{})
-
-	// Merge the scopes of the options AdditionalScopes with the scopes read from the keyFile config
-	var scopesToAdd []string
-	if len(options.AdditionalScopes) > 0 {
-		scopesToAdd = append(scopesToAdd, options.AdditionalScopes...)
-	}
-
-	if keyFile.Scope != "" {
-		scopesSplit := strings.Split(keyFile.Scope, " ")
-		scopesToAdd = append(scopesToAdd, scopesSplit...)
-	}
-	options.AdditionalScopes = scopesToAdd
-
 	return newClientCredentialsFlow(
 		options,
 		keyFile,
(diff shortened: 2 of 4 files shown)


oauth2/client_credentials_provider_test.go
```diff
@@ -34,12 +34,14 @@ func TestNewClientCredentialsProviderFromKeyFile(t *testing.T) {
...
 	ClientEmail := "CLIENT_EMAIL"
 	IssuerURL := "ISSUER_URL"
+	Scope := "SCOPE"
 	keyFile := &KeyFile{
 		Type:         oauthType,
 		ClientID:     clientID,
 		ClientSecret: ClientSecret,
 		ClientEmail:  ClientEmail,
 		IssuerURL:    IssuerURL,
+		Scope:        Scope,
```
RobertIndie, line 44: Could you help verify that the scope has been added to the `AdditionalScopes` successfully?

Nikolajls: I've pushed a change to the file client_credentials_flow_test.go that should verify the logic of adding the value of scope from keyfile to the `AdditionalScopes` change.

Nikolajls: I've reworked the code a bit so it now in the actual Token request, that way any usage of it has the logic.
My earlier implementation would have caused only the ones create earlier it would only happen for those created by the method NewDefaultClientCredentialsFlow and not newClientCredentialsFlow which is used in the provider test file.

A change to the client_credentials_flow_test.go was added that should verify, I however cannot run the test locally so currently only when it runs in the CI pipelines I can verify.

RobertIndie: Good work! The test works fine. Only some lint issues.

RobertIndie (approved):

Nikolajls:
@RobertIndie  Just pushed a commit that should fix the linting issue, and thank you!

RobertIndie:
> @RobertIndie Just pushed a commit that should fix the linting issue, and thank you!

@Nikolajls There are still lint issues. BTW, you can run `make lint` to check for lint issues locally.

Nikolajls:
> > @RobertIndie Just pushed a commit that should fix the linting issue, and thank you!
> 
> @Nikolajls There are still lint issues. BTW, you can run `make lint` to check for lint issues locally.

Im on a windows machine which makes the tooling a lot harder to replicate, I've tried installing the golintci-cli in WSL and currently in the process of that. 

I'll try the make lint in WSL instead now :)

Nikolajls:
@RobertIndie 

So the linting succeeded this time.

Do you know what the process is when this is merged in on how the updated go package is applied to the Pulsarctl repo that has the issue I'm trying to solve?

RobertIndie:
> @RobertIndie
> 
> So the linting succeeded this time.
> 
> Do you know what the process is when this is merged in on how the updated go package is applied to the Pulsarctl repo that has the issue I'm trying to solve?

I'm planning to release 0.13.1 next week which will contain this fix. And I'm handling the go client package upgrade for the pulsarctl.

PR #1249, opened 2024-07-12
fix: producer close was blocked

### LINKED ISSUE #1190: [BUG] The process will block if close client or producer when pulsar server is down.
#### Expected behavior

Tell us what should happen

When calling the client or producer close method, it should fail and return.

#### Actual behavior

Tell us what happens instead

Instead of failing and returning, the close method keeps blocking if the pulsar cluster is down.

#### Steps to reproduce

1. Deploy a standalone pulsar server.
2. Connect a go client to the server.
3. Shut down the server.
4. Close the client.
5. Observe that the process blocks.

#### System configuration
**Pulsar version**: pulsar-client-go v0.11.0

nodece:
Closes #1190

### Motivation

When the producer keeps reconnecting to the broker in the event loop, at the same time, the user calls `producer.Close`, the producer cannot be closed, because the event loop is executing the reconnect request.

### Modifications

- Add a context to control the reconnect and send logic, and stop execution when the context is done.
- Upgrade go from 1.20 to 1.21, because `testcontainers` requires go 1.21.

### Verifying this change

Added test.


gunli:
Hi @nodece , 

I think calling `p.internalClose(cp)` is not a good idea to fix this bug, 'cause it breaks the thread/goroutine model of the SDK(the CSP model: DO NOT COMMUNICATE BY SHARING MEMORY; INSTEAD, SHARE MEMORY BY COMMUNICATING.). The purpose of the cmd channel is to isolate different goroutines, but now we just bypass it.

The root cause of this bug is that the event loop is stuck in an infinite loop in reconnectToBroker(), I think infinite loop in a case of select{case...} is not a good design, we can run the infinite loop in a seperate goroutine, like this:
```go
case connectionClosed := <-p.connectClosedCh:
			p.log.Info("runEventsLoop will reconnect in producer")
			// Start a new goroutine for reconnecting to broker
			go p.reconnectToBroker(connectionClosed)
```

when the producer is closed, it will quit at:
```go
			// Producer is already closing
			p.log.Info("producer state not ready, exit reconnect")
			return
		}
```

nodece:
`Reconnect` and `Close` are different things.

The reconnect request should block the send and flush requests.

The close request shouldn't be added to the event loop, otherwise, the close request will be blocked when reconnecting.

gunli:
> The reconnect request should bloc

OK, since `The reconnect request should block the send and flush requests.` and `The close request shouldn't be added to the event loop`, we need to make sure that the `runEventsLoop` completely stopped after we call `internalClose()` (to be precise, it should be after calling p.casProducerState(producerReady, producerClosing)), Or when the state is set to producerClosing, the reconnecting case of the event loop will return and other cases will begin to run, these case may have race condition with `internalClose()`, which will lead to panic or some other bugs, it may not happen now, but it is hard to say that it won't happen in the future when we extend the code of event loop cases or the logic of internalClose().

nodece:
@gunli Good catch, we should improve the event loop, once the producer is closed, the event loop should be stopped.

RobertIndie:
> we need to make sure that the runEventsLoop completely stopped after we call internalClose()

@gunli  +1

nodece:
@gunli @RobertIndie The `runEventsLoop` can be stopped, when the producer is closed.

Event:
- reconnect request
This method already checks the producer state: https://github.com/apache/pulsar-client-go/blob/v0.13.0/pulsar/producer_partition.go#L468

- send and close request
Close methods will close the `dataChan` and `cmdChan`, the event loop can check this close status:
  - https://github.com/apache/pulsar-client-go/blob/v0.13.0/pulsar/producer_partition.go#L547
  - https://github.com/apache/pulsar-client-go/blob/v0.13.0/pulsar/producer_partition.go#L553

gunli:
> @gunli @RobertIndie The `runEventsLoop` can be stopped, when the producer is closed.
> 
> Event:
> 
> * reconnect request
>   This method already checks the producer state: https://github.com/apache/pulsar-client-go/blob/v0.13.0/pulsar/producer_partition.go#L468
> * send and close request
>   Close methods will close the `dataChan` and `cmdChan`, the event loop can check this close status:
>   
>   * https://github.com/apache/pulsar-client-go/blob/v0.13.0/pulsar/producer_partition.go#L547
>   * https://github.com/apache/pulsar-client-go/blob/v0.13.0/pulsar/producer_partition.go#L553

But the behavior is unpredictable as they are run in different goroutines, after `doClose()` update the `state` to `closing`, `reconnecting()` will return immediately, but now `dataChan` and `cmdChan` are not closed yet, neither the batchFlushTicker, the event loop will coninue to run, it can still run into `internalSend()` and `internalFlushCurrentBatch()`.

nodece:
Updated.

gunli:
Good job, but I think it can be better, my idea is:
1. add mutex and cancelfunc to producer
```go
type partitionProducer struct {

	reconnectMutex     sync.Mutex
	reconnectCancelFun context.CancelFunc
}
```
2. when reconnecting, pass a context to it so that we can cancel it outside
```go
case connectionClosed := <-p.connectClosedCh:
			p.log.Info("runEventsLoop will reconnect in producer")
			ctx, cancel := context.WithCancel(context.Background())
			p.reconnectMutex.Lock()
			p.reconnectCancelFun = cancel
			p.reconnectMutex.Unlock()
			p.reconnectToBroker(ctx, connectionClosed)
			cancel()
```
```go
func (p *partitionProducer) reconnectToBroker(ctx context.Context, connectionClosed *connectionClosed) {
	var maxRetry int
	if p.options.MaxReconnectToBroker == nil {
		maxRetry = -1
	} else {
		maxRetry = int(*p.options.MaxReconnectToBroker)
	}

	var (
		delayReconnectTime time.Duration
		defaultBackoff     = internal.DefaultBackoff{}
	)

	for maxRetry != 0 {
		select {
		case <-ctx.Done():
			return // quit
		default:

		}
			// Producer is already closing
			p.log.Info("producer state not ready, exit reconnect")
			return
		}

		var assignedBrokerURL string

		if connectionClosed != nil && connectionClosed.HasURL() {
			delayReconnectTime = 0
		} else if p.options.BackoffPolicy == nil {
			delayReconnectTime = defaultBackoff.Next()
		} else {
			delayReconnectTime = p.options.BackoffPolicy.Next()
		}

		p.log.WithFields(log.Fields{
			"assignedBrokerURL":  assignedBrokerURL,
			"delayReconnectTime": delayReconnectTime,
		}).Info("Reconnecting to broker")
		time.Sleep(delayReconnectTime)

		// double check
			// Producer is already closing
			p.log.Info("producer state not ready, exit reconnect")
			return
		}

		atomic.AddUint64(&p.epoch, 1)
		err := p.grabCnx(assignedBrokerURL)
		if err == nil {
			// Successfully reconnected
			p.log.WithField("cnx", p._getConn().ID()).Info("Reconnected producer to broker")
			return
		}
		p.log.WithError(err).Error("Failed to create producer at reconnect")
		errMsg := err.Error()
		if strings.Contains(errMsg, errMsgTopicNotFound) {
			// when topic is deleted, we should give up reconnection.
			p.log.Warn("Topic not found, stop reconnecting, close the producer")
			p.doClose(joinErrors(ErrTopicNotfound, err))
			break
		}

		if strings.Contains(errMsg, errMsgTopicTerminated) {
			p.log.Warn("Topic was terminated, failing pending messages, stop reconnecting, close the producer")
			p.doClose(joinErrors(ErrTopicTerminated, err))
			break
		}

		if strings.Contains(errMsg, errMsgProducerBlockedQuotaExceededException) {
			p.log.Warn("Producer was blocked by quota exceed exception, failing pending messages, stop reconnecting")
			p.failPendingMessages(joinErrors(ErrProducerBlockedQuotaExceeded, err))
			break
		}

		if strings.Contains(errMsg, errMsgProducerFenced) {
			p.log.Warn("Producer was fenced, failing pending messages, stop reconnecting")
			p.doClose(joinErrors(ErrProducerFenced, err))
			break
		}

		if maxRetry > 0 {
			maxRetry--
		}
		p.metrics.ProducersReconnectFailure.Inc()
		if maxRetry == 0 || defaultBackoff.IsMaxBackoffReached() {
			p.metrics.ProducersReconnectMaxRetry.Inc()
		}
	}
}
```
3. when closing, check if we are in reconnecting, if so, cancel it and wait for closing done
```go
func (p *partitionProducer) Close() {
		// Producer is closing
		return
	}

	cp := &closeProducer{doneCh: make(chan struct{})}
	p.cmdChan <- cp

	// wait for close producer request to complete
	ctx := context.Background()
	for {
		select {
		case <-cp.doneCh: // close done
			return
		case <-ctx.Done(): // app exit
			return
		default:
			// if we are stuck in reconnecting, cancel it
			p.reconnectMutex.Lock()
			if p.reconnectCancelFun != nil {
				p.reconnectCancelFun()
			}
			p.reconnectMutex.Unlock()
			time.Sleep(10 * time.Millisecond)
		}
	}
}
```

In this way, we can make sure
1. that the closing and reconnecting procedures are run in the same goroutine with no race condition;
2. the main goroutine and the event loop goroutine just share the cancel func
3. when closing, we still can wait for closing done even if we are stuck in reconnecting.

pulsar/producer_partition.go
```diff
@@ -464,77 +469,82 @@ func (p *partitionProducer) reconnectToBroker(connectionClosed *connectionClosed
 	)
 
 	for maxRetry != 0 {
-		if p.getProducerState() != producerReady {
-			// Producer is already closing
-			p.log.Info("producer state not ready, exit reconnect")
+		select {
+		case <-p.ctx.Done():
```
gunli, line 473: the default case is too long, may be below is better
select {
case <-p.ctx.Done()
    return
default:
}

nodece: Done.

crossoverJie (approved):
LGTM.

gunli (approved):
LGTM

gunli:
This PR may be related to issue #1190

PR #1250, opened 2024-07-13
ci: validate multiple version builds and use golangci-lint-action

nodece:
### Motivation

Improve CI.

### Modifications

- build and test uses the `go-version: [ '1.20', '1.21', '1.22' ]` matrix.
- lint uses the go 1.20(golangci-lint@1.51.2 doesn't work on go 1.21 or later version) version and `golangci/golangci-lint-action@v6`, the lint result will be reported on the pr.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v5
         with:
-          go-version: 'stable'
+          go-version: '1.20'
       - name: Check license header
         run: docker run --rm -v $(pwd):/github/workspace ghcr.io/korandoru/hawkeye-native:v3 check
       - name: Run golangci-lint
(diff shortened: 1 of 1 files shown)


tisonkun (approved):

.github/workflows/ci.yml
```diff
@@ -20,25 +20,35 @@ on: [pull_request]
...
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.20'
       - name: Check license header
         run: docker run --rm -v $(pwd):/github/workspace ghcr.io/korandoru/hawkeye-native:v3 check
       - name: Run golangci-lint
-        run: make lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.51.2
```
tisonkun, line 45: Why this change? We still provide the Makefile target. This unaligned may silently cause CI logic and dev logic different ..

Both download binaries so that time consumption should be similiar.

nodece: golangci/golangci-lint-action can report lint result to the ci summary.

When we update the golangci-lint, we must update the workflow and makefile files to keep same version.

PR #1251, opened 2024-07-15
fix pulsar admin revoke subscription permission endpoint

NoFacePeace:
### Motivation

The endpoint of RevokeSubPermission does't contain `subscription`

wrong endpoint: /{property}/{namespace}/permissions/subscription/{subscription}/{role}

correct endpoint: /{property}/{namespace}/permissions/{subscription}/{role}

https://github.com/apache/pulsar/blob/88ebe785dbdab239104981453a9bd0e4a7e896d3/pulsar-broker/src/main/java/org/apache/pulsar/broker/admin/v2/Namespaces.java#L366

### Modifications

delete subscription


nodece:
Could you add a test for this PR?

PR #1255, opened 2024-07-23
fix: Connection use minConnMaxIdleTime

crossoverJie:
### Motivation

Keep consistent with [Java client](https://github.com/apache/pulsar/blob/6b2938223cf45a9298f9d40ab6ae108bea9a5a6d/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionPool.java#L152-L156).

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/client_impl.go
+++ b/pulsar/client_impl.go
@@ -66,8 +66,10 @@ func newClient(options ClientOptions) (Client, error) {
 	if connectionMaxIdleTime == 0 {
 		connectionMaxIdleTime = defaultConnMaxIdleTime
 	} else if connectionMaxIdleTime > 0 && connectionMaxIdleTime < minConnMaxIdleTime {
-		return nil, newError(InvalidConfiguration, fmt.Sprintf("Connection max idle time should be at least %f "+
-			"seconds", minConnMaxIdleTime.Seconds()))
+		logger.Warnf("Connection idle detect interval seconds default same as max idle seconds, but max idle"+
+			" seconds less than %d, to avoid checking"+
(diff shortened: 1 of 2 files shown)


pulsar/client_impl.go
```diff
@@ -66,8 +66,10 @@ func newClient(options ClientOptions) (Client, error) {
 	if connectionMaxIdleTime == 0 {
 		connectionMaxIdleTime = defaultConnMaxIdleTime
 	} else if connectionMaxIdleTime > 0 && connectionMaxIdleTime < minConnMaxIdleTime {
-		return nil, newError(InvalidConfiguration, fmt.Sprintf("Connection max idle time should be at least %f "+
-			"seconds", minConnMaxIdleTime.Seconds()))
+		logger.Warnf("Connection idle detect interval seconds default same as max idle seconds, but max idle"+
+			" seconds less than %d, to avoid checking"+
+			" connection status too much, use default value : %d", minConnMaxIdleTime.Minutes(), minConnMaxIdleTime.Minutes())
+		connectionMaxIdleTime = minConnMaxIdleTime
```
RobertIndie, line 72: This is a misconfiguration. And IMO, we should throw an error for it instead of the warning. Otherwise, the client's behavior may differ from what users expect.
Are there any cases where this change is needed?

crossoverJie: Thank you for your explanatition, it as is might be a good choice as weel.

PR #1256, opened 2024-07-25
improve: use ctx and timer instead sleep

nodece:
### Motivation

We are using `time.Sleep` and `for` to execute the retry, which cannot be interrupted, this PR will use context and timer to improve this behavior.

The next improvement idea supports passing the context to interrupt the retry, when closing the producer or consumer, we need to do that.

### Modifications

- Add the `Retry` method to execute the try, and use `Retry` instead of `time.Sleep` and `for`.

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -623,12 +623,9 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 		return req.msgID, req.err
 	}
 
-	opFn := func() (*trackingMessageID, error) {
-		return request()
-	}
 	ctx, cancel := context.WithTimeout(context.Background(), pc.client.operationTimeout)
 	defer cancel()
-	res, err := internal.Retry(ctx, opFn, func(err error) time.Duration {
+	res, err := internal.Retry(ctx, request, func(err error) time.Duration {
 		nextDelay := bo.Next()
 		pc.log.WithError(err).Errorf("Failed to get last message id from broker, retrying in %v...", nextDelay)
 		return nextDelay
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"container/list"
+	"context"
 	"encoding/hex"
 	"fmt"
 	"math"
@@ -612,7 +613,6 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 		pc.log.WithField("state", state).Error("Failed to getLastMessageID for the closing or closed consumer")
 		return nil, errors.New("failed to getLastMessageID for the closing or closed consumer")
 	}
-	remainTime := pc.client.operationTimeout
 	bo := pc.backoffPolicyFunc()
 	request := func() (*trackingMessageID, error) {
 		req := &getLastMsgIDRequest{doneCh: make(chan struct{})}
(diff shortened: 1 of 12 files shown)


RobertIndie (commented):
Overall looks good to me. Left one comment.

pulsar/consumer_partition.go
```diff
@@ -595,23 +596,28 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
...
-		if remainTime <= 0 {
-			pc.log.WithError(err).Error("Failed to getLastMessageID")
-			return nil, fmt.Errorf("failed to getLastMessageID due to %w", err)
-		}
-		nextDelay := backoff.Next()
-		if nextDelay > remainTime {
-			nextDelay = remainTime
-		}
-		remainTime -= nextDelay
-		pc.log.WithError(err).Errorf("Failed to get last message id from broker, retrying in %v...", nextDelay)
```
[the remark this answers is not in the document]
nodece: @RobertIndie Done.

pulsar/internal/retry.go
```diff
@@ -0,0 +1,62 @@
...
+		res, err = op()
+		if err == nil {
+			return res, nil
+		}
+
+		duration := nextDuration(err)
+		if timer == nil {
+			timer = time.NewTimer(duration)
+		} else {
+			timer.Reset(duration)
```
RobertIndie, line 53: According to the [go doc of the timer](https://pkg.go.dev/time#Timer.Reset):

> If a program has not yet received a value from t.C, however, the timer must be stopped and—if Stop reports that the timer expired before being stopped—the channel explicitly drained

We should stop the timer and drain the channel before resetting the timer.

nodece: > // If a program has already received a value from t.C, the timer is known
> // to have expired and the channel drained, so t.Reset can be used directly.

The `timer.C` has been consumed, so we can use `timer.Reset` directly.

pulsar/consumer_test.go
```diff
@@ -3691,6 +3691,9 @@ func TestConsumerSeekByTimeOnPartitionedTopic(t *testing.T) {
 	err = consumer.SeekByTime(currentTimestamp.Add(-retentionTimeInSecond))
 	assert.Nil(t, err)
 
+	// Seek command disconnects the consumer, so we need to wait for reconnection.
+	<-time.After(5 * time.Second)
```
RobertIndie, line 3695: What happens if we don't add `time.After` here? We should allow the user to invoke `consumer.Receive` immediately after the seek operation.

pulsar/internal/retry.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+type OpFn[T any] func() (T, error)
+
+// Retry the given operation until the returned error is nil or the context is done.
+func Retry[T any](ctx context.Context, op OpFn[T], nextDuration func(error) time.Duration) (T, error) {
```
crossoverJie, line 29: This function is with the internal package, Why not rename it `retry`.

pulsar/consumer_test.go
nodece: CI will fail.

There are races:
```
2024-07-31T04:53:49.4735743Z time="2024-07-31T04:36:16Z" level=info msg="Reconnected consumer to broker" consumerID=5 name=itqer subscription=my-sub topic="persistent://public/default/TestSeekByTimeOnPartitionedTopic-partition-3"
2024-07-31T04:53:49.4736472Z time="2024-07-31T04:40:19Z" level=warning msg="Failed to write on connection" error="write tcp 127.0.0.1:57350->127.0.0.1:6650: use of closed network connection" local_addr="127.0.0.1:57350" remote_addr="pulsar://localhost:6650"
2024-07-31T04:53:49.4737067Z time="2024-07-31T04:40:26Z" level=warning msg="Failed to write on connection" error="write tcp 127.0.0.1:52750->127.0.0.1:6650: use of closed network connection" local_addr="127.0.0.1:52750" remote_addr="pulsar://localhost:6650"
2024-07-31T04:53:49.4737665Z time="2024-07-31T04:40:26Z" level=warning msg="Failed to write on connection" error="write tcp 127.0.0.1:52750->127.0.0.1:6650: use of closed network connection" local_addr="127.0.0.1:52750" remote_addr="pulsar://localhost:6650"
```

Because the reconnect and ack requests of the consumer in the different goroutine.

In the past, I guess that CI is passed because of sleep.

pulsar/internal/retry.go
nodece: The pulsar package need to use the internal package, so we need to use `Retry` instead of `retry`.

nodece:
Let me fix the consumer seek test.

pulsar/consumer_test.go
RobertIndie: Which sleep are you referring to?

nodece: Reconnect to broker.

nodece:
Seek test will be fixed by #1265.

nodece:
@RobertIndie @crossoverJie Could you have a chance to review?

pulsar/consumer_partition.go
```diff
@@ -622,23 +622,23 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
...
-			pc.log.WithError(err).Error("Failed to getLastMessageID")
-			return nil, fmt.Errorf("failed to getLastMessageID due to %w", err)
-		}
+
+	opFn := func() (*trackingMessageID, error) {
+		return request()
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), pc.client.operationTimeout)
+	defer cancel()
+	res, err := internal.Retry(ctx, opFn, func(err error) time.Duration {
```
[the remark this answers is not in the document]
nodece: Good catch!

crossoverJie (approved):
LGTM

PR #1265, opened 2024-08-01
fix: seek race

nodece:
### Motivation

Broker's seek logic: 
1. Close the consumer, not connection.
2. Move the read position of the cursor.
3. Return response to the client.

Seek issue:
1. Seek and reconnect requests in the different goroutines, which should be executed serially.
2. Message loss after seek operation based on time: the partitioned topic has multiple consumer partitions, when all seek request is done, the parent consumer will clean up the receive queue, and at the same time, the consumer partition re-reads the message asynchronous. when a message is received, the consumer partition will push the message to the parent consumer, there is a concurrency issue with message cleaning and message push.

### Modifications

- Use the same goroutine to perform the seek and reconnect.
- The parent consumer pauses receiving messages from all consumer partitions before performing the seek operation and then cleans up the message queue, when the seek is done on the consumer partition, continues to push the message to the parent consumer.

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -1587,22 +1587,16 @@ func (pc *partitionConsumer) runEventsLoop() {
 	}()
 	pc.log.Debug("get into runEventsLoop")
 
-	go func() {
-		for {
-			select {
-			case <-pc.closeCh:
-				pc.log.Info("close consumer, exit reconnect")
-				return
-			case connectionClosed := <-pc.connectClosedCh:
-				pc.log.Debug("runEventsLoop will reconnect")
-				pc.reconnectToBroker(connectionClosed)
-			}
-		}
-	}()
-
 	for {
-		for i := range pc.eventsCh {
-			switch v := i.(type) {
+		select {
+		case <-pc.closeCh:
+			pc.log.Info("close consumer, exit reconnect")
+			return
+		case connectionClosed := <-pc.connectClosedCh:
+			pc.log.Debug("runEventsLoop will reconnect")
+			pc.reconnectToBroker(connectionClosed)
+		case event := <-pc.eventsCh:
+			switch v := event.(type) {
 			case *ackRequest:
 				pc.internalAck(v)
 			case *ackWithTxnRequest:
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -1463,6 +1463,8 @@ func (pc *partitionConsumer) dispatcher() {
 				}
 				pc.metrics.PrefetchedMessages.Dec()
 				pc.metrics.PrefetchedBytes.Sub(float64(len(queueMsg.payLoad)))
+			} else {
+				pc.log.Debug("skip dispatching messages when seeking")
 			}
 		} else {
 			queueCh = pc.queueCh
@@ -1500,6 +1502,7 @@ func (pc *partitionConsumer) dispatcher() {
 			if !ok {
 				return
 			}
+			pc.log.Debug("received dispatcherSeekingControlCh, set isSeek to true")
 			pc.isSeeking.Store(true)
 
 		case msg, ok := <-queueCh:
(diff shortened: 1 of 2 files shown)


pulsar/consumer_partition.go
```diff
@@ -1594,6 +1607,13 @@ func (pc *partitionConsumer) runEventsLoop() {
 				pc.log.Info("close consumer, exit reconnect")
 				return
 			case connectionClosed := <-pc.connectClosedCh:
+				if pc.seeking.Load() {
+					go func() {
+						<-time.After(100 * time.Millisecond)
+						pc.connectClosedCh <- connectionClosed
```
RobertIndie, line 1613: Does the Java consumer have the same issue?

nodece: Java client has been fixed this issue

RobertIndie: Ok. I see that the Java client will wait a while before reconnecting to the broker:
https://github.com/apache/pulsar/blob/83b86abcb74595d7e8aa31b238a7dbb19a04dde2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java#L200

And it's the same for the go client: 
https://github.com/apache/pulsar-client-go/blob/682bf5fde149754c1648fbffccd39c2c51f17551/pulsar/consumer_partition.go#L1719

So it doesn't seem the go client should have this issue, right?

nodece: Fixed by this PR.

nodece:
Could you review this PR again? @RobertIndie

pulsar/consumer_partition.go
```diff
@@ -1440,17 +1452,18 @@ func (pc *partitionConsumer) dispatcher() {
...
 			nextMessageSize = queueMsg.size()
 
-			if pc.dlq.shouldSendToDlq(&nextMessage) {
-				// pass the message to the DLQ router
-				pc.metrics.DlqCounter.Inc()
-				messageCh = pc.dlq.Chan()
-			} else {
-				// pass the message to application channel
-				messageCh = pc.messageCh
+			if !pc.isSeeking.Load() {
```
crossoverJie, line 1455: Why not use channel for communication here? but use memory.

pulsar/consumer_partition.go
```diff
@@ -1440,17 +1452,18 @@ func (pc *partitionConsumer) dispatcher() {
...
+				if pc.dlq.shouldSendToDlq(&nextMessage) {
+					// pass the message to the DLQ router
+					pc.metrics.DlqCounter.Inc()
+					messageCh = pc.dlq.Chan()
+				} else {
+					// pass the message to application channel
+					messageCh = pc.messageCh
+				}
+				pc.metrics.PrefetchedMessages.Dec()
+				pc.metrics.PrefetchedBytes.Sub(float64(len(queueMsg.payLoad)))
```
crossoverJie, line 1465: If messages can no longer be distributed, it is recommended to print a warning log to facilitate user troubleshooting.

pulsar/consumer_partition.go
```diff
@@ -185,6 +186,16 @@ type partitionConsumer struct {
 
 	redirectedClusterURI string
 	backoffPolicyFunc    func() backoff.Policy
+
+	dispatcherSeekingControlCh chan bool
```
crossoverJie, line 190: Can `dispatcherSeekingControlCh chan struct{}`  be used instead?

pulsar/consumer_partition.go
nodece: The channel is used to add or consume once. There is a loop, so we need to use `atomic.Bool` instead `chan`.

pulsar/consumer_partition.go
nodece: I added the debug log, which can help user to debug the seek.

pulsar/consumer_partition.go
```diff
@@ -1680,6 +1696,9 @@ func (pc *partitionConsumer) internalClose(req *closeRequest) {
 }
 
 func (pc *partitionConsumer) reconnectToBroker(connectionClosed *connectionClosed) {
+	if pc.isSeeking.CompareAndSwap(true, false) {
```
crossoverJie, line 1699: Wouldn't it be better to move to line 1747?
https://github.com/nodece/pulsar-client-go/blob/1f043f814406bc604314353467485eeeb9217e5e/pulsar/consumer_partition.go#L1745-L1750

nodece: Once the connection is established, the broker will push the message to the client, so we need to reset the seek flag before the request reconnects.

crossoverJie: thank you for your explanation

crossoverJie (approved):
LGTM

PR #1273, opened 2024-08-28
[Issue 1272][connection] Attempt to avoid deadlock during reconnection

### LINKED ISSUE #1272: Deadlock in connection pool
#### Expected behavior

The client should properly handle a cluster that is not ready yet and retry until it gets healthy again.

#### Actual behavior

When the client is attempting to reconnect or to create new producers, consumers or readers but the service is not ready, it has a chance to block the connection pool.

#### Steps to reproduce

`ConnectionClosed` callbacks have a chance of blocking the connection pool because the `GetConnection` of the pool may close a connection when the state has changed, which happens when the cluster is not ready. In our case, we were observing a lot of closing because right after getting a connection to the broker, it was closing due to ServiceNotReady since too many bookies were down.

#### System configuration
**Pulsar version**: 3.0.5
**Pulsar client**: 13.1

Gilthoniel:
Fixes #1272 

### Motivation

Producers and consumers register themselves to the connection to be notified when it gets closed. Even though the callback push the events in a channel, it can get stuck and the connection pool is locked which prevents any other caller to get a connection.

### Modifications

This PR makes sure that the callback never blocks.

### Verifying this change

none

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -332,7 +332,7 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 		startMessageID:       atomicMessageID{msgID: options.startMessageID},
 		connectedCh:          make(chan struct{}),
 		messageCh:            messageCh,
-		connectClosedCh:      make(chan *connectionClosed, 10),
+		connectClosedCh:      make(chan *connectionClosed, 1),
 		closeCh:              make(chan struct{}),
 		clearQueueCh:         make(chan func(id *trackingMessageID)),
 		compressionProviders: sync.Map{},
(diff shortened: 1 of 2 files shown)


nodece (approved):

nodece:
We can use a goroutine to call the `ConnectionClosed` method in the connection pool.

What do you think of that?

nodece:
If the broker is not ready, the client will retry to connect to the broker. Once the broker is ready, the client can have an available connection to create the producer/consumer, when the producer/consumer is created, and then the connection pool listens to the producer/consumer. 

I don't know why you can receive several dozens or hundreds of callback.


If your network environment is unstable? or if the broker is changing its status?

> What is your concern with this fix ?

You are dropping the reconnect request.

Gilthoniel:
It happens during a restart of everything so it's very unstable yes, brokers are probably changing their status.

I understand your worry but here it ensures that all reconnects will be handled and it only drops duplicate. Also in normal situation, as you said, you get only one reconnect.

lhotari:
Great contributions @Gilthoniel! Thank you

RobertIndie (approved):

PR #1275, opened 2024-08-29
[Issue 1274][connection] Refactor connection concurrency model

### LINKED ISSUE #1274: Race during reconnect
#### Expected behavior

After reconnecting to the broker, a consumer or a producer should register itself for callbacks.

#### Actual behavior

After reconnecting to the broker, a consumer attempts to add itself before the connection is closed then attempts to lock the list of handlers alongside dozen of other consumers. The connection gets closed at that moment because the service is not ready and wins the race against the lock, removes the listeners and closes. Then the consumer adds the listener and considers itself ready.

This can happen because a mutex has no guarantee to the order routines will acquire it.

#### Steps to reproduce

I could observe this behavior in a cluster considered not ready as it was restarting. After the application reconnected, one reader was considered ok and was waiting for messages while it was actually not registered in the topic.

#### System configuration
**Pulsar version**: v3.0.5
**Pulsar client**: v13.1

Gilthoniel:
Fixes #1274 

### Motivation

A race can happen while a consumer or a producer is registering itself to the connection to later receive closing notification.

### Modifications

This pull request changes the concurrency model of the connection to properly lock the different actors between the closing procedure and the registration of handlers and listeners. After this, a caller will properly check the state of the connection and aborts if closed, and it will be certain that it will not be registered in concurrency with a closing.

### Verifying this change

none

### Documentation

none

### DIFF
--- a/pulsar/internal/connection.go
+++ b/pulsar/internal/connection.go
@@ -35,8 +35,6 @@ import (
 
 	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
 	"github.com/apache/pulsar-client-go/pulsar/log"
-
-	ua "go.uber.org/atomic"
 )
 
 const (
@@ -87,7 +85,7 @@ type Connection interface {
 	ID() string
 	GetMaxMessageSize() int32
 	Close()
-	WaitForClose() <-chan interface{}
+	WaitForClose() <-chan struct{}
 	IsProxied() bool
 }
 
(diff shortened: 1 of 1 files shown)


nodece (approved):

lhotari (approved):
LGTM, great work @Gilthoniel

crossoverJie (approved):

frankjkelly:
@Gilthoniel Thanks for this fix. Was this issue introduced in `0.13.1` or was does it predate that and if so do you know how far back it goes? 

CC: @crossoverJie @nodece

PR #1280, opened 2024-09-04
[fix] Only decompress the payload if it's not empty

stepanbujnak:
### Motivation

The message payload is optional and in some cases only message properties are sent. In this case, the message decompression would fail.

### Modifications

Only do the decompression if the payload is not empty.

### Verifying this change

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - If a feature is not applicable for documentation, explain why? it's just a bugfix


nodece (approved):

crossoverJie (approved):
LGTM

PR #1289, opened 2024-09-24
[pulsaradmin] Add GetAllSchemas command

calindima:
Didn't create an issue but I can if that's necessary, otherwise I'll write the details in this PR.

### Motivation

We actively use `pulsarctl` which has a dependency on this library. We have a use case where we want to interogate the SchemaRegistry to retrieve a history of schema versions and observe schema evolution for topics with higher retention (e.g. infinite retention, >6mo retention etc.).

### Modifications

The PR only contains additions such as:
- Added `GetAllSchemas` method to the `Schema` interface
- Added the implementation for the `GetAllSchemas` method
- Added a model for the Pulsar Admin API response
- Added a utility method to convert from this model to a list of `SchemaInfoWithVersion`

### Verifying this change

  - *Added an integration test that uploads a schema on a topic and then asserts the length of all schemas is 1*

  - **The public API: (yes)**
  - **The schema: (don't know)**
  - The wire protocol: (no)

### Documentation

  - If yes, how is the feature documented? (doc comment)

Just let me know if there's any documentation I should update - I only added a godoc comment to the new method

### DIFF
--- a/pulsaradmin/pkg/admin/schema.go
+++ b/pulsaradmin/pkg/admin/schema.go
@@ -144,7 +144,6 @@ func (s *schemas) GetAllSchemas(topic string) ([]*utils.SchemaInfoWithVersion, e
 
 	err = s.pulsar.Client.Get(endpoint, &response)
 	if err != nil {
-		fmt.Println("err:", err.Error())
 		return nil, err
 	}
 
--- a/pulsaradmin/pkg/admin/schema.go
+++ b/pulsaradmin/pkg/admin/schema.go
@@ -35,6 +35,9 @@ type Schema interface {
 	// GetSchemaInfoByVersion retrieves the schema of a topic at a given <tt>version</tt>
 	GetSchemaInfoByVersion(topic string, version int64) (*utils.SchemaInfo, error)
 
+	// GetAllSchemas retrieves the latest schema of a topic
+	GetAllSchemas(topic string) ([]*utils.SchemaInfoWithVersion, error)
+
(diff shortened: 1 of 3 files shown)


crossoverJie:
> Thanks for running the workflows @crossoverJie. I fixed the linter error but the failed integration test is not related to anything changed in this PR as far as I can tell.

Yes, I have approved it. Thank you for your contribution

crossoverJie (approved):

calindima:
Thanks for the quick feedback. Any idea in which release will this be included?

pulsaradmin/pkg/admin/schema.go
```diff
@@ -130,6 +133,25 @@ func (s *schemas) GetSchemaInfoByVersion(topic string, version int64) (*utils.Sc
...
+	if err != nil {
+		return nil, err
+	}
+	var response utils.GetAllSchemasResponse
+	endpoint := s.pulsar.endpoint(s.basePath, topicName.GetTenant(), topicName.GetNamespace(),
+		topicName.GetLocalName(), "schemas")
+
+	err = s.pulsar.Client.Get(endpoint, &response)
+	if err != nil {
+		fmt.Println("err:", err.Error())
```
nodece, line 147: I don't think you should print the error here.

calindima: Removed. It came together with copy pasting `GetSchemaInfoWithVersion`.

nodece (approved):

PR #1293, opened 2024-09-28
[improve] let tlsoptions include std tlsconfig

dream-kzx:
Motivation
    I propose to incorporate the standard library's tlsconfig into tlsoptions to accommodate specific TLS configuration requirements that are not currently supported. There are scenarios, particularly in private cloud environments, where domain names may not be available, necessitating the ability to skip domain validation or implement custom validation logic.

    By integrating the tlsconfig from the standard library into tlsoptions, we can ensure compatibility while extending the functionality to meet these specialized needs. This enhancement will provide users with greater flexibility in configuring their TLS settings, making it more aligned with the way tlsconfig is passed in Go.
   
    Of course this is for discussion only and does not cover all the code


shoothzj:
https://lists.apache.org/thread/985zqvlgv1kggqsnxvwbmpsm2r45mwk0

pulsar/client.go
```diff
@@ -129,6 +129,8 @@ type ClientOptions struct {
 	// TLSMaxVersion contains the maximum TLS version that is acceptable. See tls.Config MaxVersion for more information.
 	TLSMaxVersion uint16
 
+	StdTlsConfig *tls.Config
```
nodece, line 132: Or `TLSConfig *tls.Config`?

dream-kzx: ok, now I'd rather talk about the next two proposals, which one do you prefer.

dream-kzx: Has been changed to TLSConfig

nodece (approved):

shoothzj (approved):

Issue #1142, opened 2023-12-06
Expose Error for user


WoWsj:
**Is your feature request related to a problem? Please describe.**
Hello, I'm currently trying to use pulser-client-go, and it seems very useful and easy. 

However, producer can occur error and it is really ok. But, in some cases, I want to do some specific behavior based on its error content.

But, this package doesn't expose Error's field and doesn't offer any method or easy way to compare errors. So, I need to rewrite errors and write some comparing logic for my business logic.

I think it is not good pattern. If it is ok, please expose errors.

**Describe the solution you'd like**

Expose errors, and let users be able to use error.

tisonkun:
Thanks for reporting this case @WoWsj!

I agree that currently almost all errors pulsar-client-go returns are opaque and we may expose the error in the public API or define some common categories (is_fatal? is_retriable?).

Can you share a bit more details what methods you want to handle its errors and what errors you want handle specifically?

Besides, I remember that @flowchartsman ever had a similar idea to expose client errors API in https://github.com/apache/pulsar-client-go/issues/1075. May you can share the design thoughts here to find a consensus.

Gleiphir2769:
> Expose errors, and let users be able to use error.
+1. I think at least it should allow users to use the std library to compare errors (errors.Is/As).

Gleiphir2769:
I have searched all errors pulsar-client-go returned with key words `fmt.Errorf` and `errors.New`. Unfortunataly, there are almost no existing Error struct that can be exposed (e.g. errConnectionClosed). In most cases, anonymous errors are returned directly.

<img width="897" alt="image" src="https://github.com/apache/pulsar-client-go/assets/50270213/e1712d16-dbef-48f3-966a-911b9382e980">

<img width="897" alt="image" src="https://github.com/apache/pulsar-client-go/assets/50270213/15ed85b2-d721-45b5-a51d-691624d967bc">



> sorting out all the errors 

I think it's too difficult to sorting out ALL errors. The better way is to sort out errors relative producer/consumer public API. What do you think? @tisonkun

WoWsj:
> Still, user cases that need to handle concrete errors specially help us formalize the API.

I am trying to use `pulser-client-go` for streaming process application. In some cases, some errors can be handled by just printing errors(like format issues or something like just informational error), but other cases which needs specific action like recreating client needs to be handled by application side. 

However, when I checked the code,
1. Basically, errors are not exposed to users
[...]

PR #1301, opened 2024-10-29
Support acknowledging a list of message IDs

BewareMyPower:
### Motivation

For `Shared` and `Key_Shared` subscriptions, if some messages failed to acknowledge, these messages would never be dispatched to the consumer until restarted. If the number of unacknowledged messages reached the threshold, the broker would never dispatch messages anymore. However, the consumer does not have a chance to check which messages failed to acknowledged.

Even if this case was not hit, if the consumer restarted after consuming many messages, the old unacknowledged messages would be delivered again, which is very confusing and might affect the business logic.

Therefore, we can only enable `AckWithResponse` to know which messages failed to acknowledge. Unfortunately, currently the Go SDK only supports acknowledging single messages. It harms the performance significantly.

To solve this solution, this PR adds an API to acknowledge a list of messages.

```golang
	// AckIDList the consumption of a list of messages, identified by their MessageIDs
	// Returns a map of MessageID to error, the keys are the MessageIDs that failed to be acknowledged
	AckIDList([]MessageID) map[MessageID]error
```

Users can save the failed message IDs and add them again in the next `AckIDList` call.

### Modifications

- Add an `AckIDList` API and reuse the logic of `ack_grouping_tracker.go` to convert user provided message IDs to the message IDs in the ACK requests
- Add the request id to the ACK request in `internalAckList` and wait for the response error via a error channel
- Add `TestAckIDList` to verify the case that a message ID list has message IDs of non-batched messages, whole batched messages or partial batched messages because the behaviors are different if the batch index ACK is enabled
- Add `TestMultiTopicAckIDList` to verify the multi-topics case, including regex subscription.

  - The public API: (**yes**)

### Documentation

  - Does this pull request introduce a new feature? (**yes**)

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -329,8 +329,8 @@ type Consumer interface {
 	//
 	// If the consumer has subscribed multiple topics or partitions, since each topic is associated with an ACK request,
 	// `AckError` will be returned if there are errors because some message IDs might succeed while others might fail.
-	// If you're sure that only 1 message is subscribed, all message IDs should be re-acknowledged when it returns an
-	// error. Otherwise, you should cast the error to `AckError` and handle it accordingly.
+	// If you're sure that only 1 message is subscribed, you should treat all message IDs as failed when it returns an
+	// error. Otherwise, you should cast the error to `AckError` and check the error for each message ID.
 	AckIDList([]MessageID) error
 
 	// AckWithTxn the consumption of a single message with a transaction
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -307,6 +307,11 @@ type Consumer interface {
 	// AckID the consumption of a single message, identified by its MessageID
 	AckID(MessageID) error
 
+	// AckIDList the consumption of a list of messages, identified by their MessageID
+	// NOTE: this API will acknowledge the MessageID even if it represents a single message in the batch,
+	//   i.e. the enableBatchIndexAck option will be treated as true when calling this method
+	AckIDList([]MessageID) map[MessageID]error
+
 	// AckWithTxn the consumption of a single message with a transaction
 	AckWithTxn(Message, Transaction) error
 
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -268,20 +268,19 @@ type ConsumerOptions struct {
 	startMessageID *trackingMessageID
 }
 
-type TopicAckError struct {
-	Err    error
-	MsgIDs []MessageID
-}
-
-// This error could only be returned when calling `AckIDList` on a consumer that subscribes multiple topics
-// The key is the topic name, and the value is the ACK error and the failed message ID in this topic.
-type AckError map[string]*TopicAckError
+// This error is returned when `AckIDList` failed and `AckWithResponse` is true.
+// It only contains the valid message IDs that failed to be acknowledged in the `AckIDList` call.
+// For those invalid message IDs, users should ignore them and not acknowledge them again.
+type AckError map[MessageID]error
 
 func (e AckError) Error() string {
 	builder := strings.Builder{}
-	for topic, topicAckError := range e {
-		builder.WriteString(fmt.Sprintf("topic: %s, error: %s, messageIDs: %v\n",
-			topic, topicAckError.Err, topicAckError.MsgIDs))
+	errorMap := make(map[string][]MessageID)
+	for id, err := range e {
+		errorMap[err.Error()] = append(errorMap[err.Error()], id)
+	}
+	for err, msgIDs := range errorMap {
+		builder.WriteString(fmt.Sprintf("error: %s, failed message IDs: %v\n", err, msgIDs))
 	}
 	return builder.String()
 }
@@ -335,14 +334,8 @@ type Consumer interface {
 	// This method should be used when `AckWithResponse` is true. Otherwise, it will be equivalent with calling
 	// `AckID` on each message ID in the list.
 	//
-	// If there are some invalid message IDs in the list, this method won't return an error for them, only error logs
-	// will be printed because retrying acknowledging them will not make a difference. Message IDs retrieved from
-	// Message objects are always valid.
-	//
-	// If the consumer has subscribed multiple topics or partitions, since each topic is associated with an ACK request,
-	// `AckError` will be returned if there are errors because some message IDs might succeed while others might fail.
-	// If you're sure that only 1 message is subscribed, you should treat all message IDs as failed when it returns an
-	// error. Otherwise, you should cast the error to `AckError` and check the error for each message ID.
+	// When `AckWithResponse` is true, the returned error could be an `AckError` which contains the failed message ID
+	// and the corresponding error.
 	AckIDList([]MessageID) error
 
 	// AckWithTxn the consumption of a single message with a transaction
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -20,6 +20,7 @@ package pulsar
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/apache/pulsar-client-go/pulsar/backoff"
@@ -267,12 +268,22 @@ type ConsumerOptions struct {
 	startMessageID *trackingMessageID
 }
 
+type TopicAckError struct {
+	Err    error
+	MsgIDs []MessageID
+}
+
 // This error could only be returned when calling `AckIDList` on a consumer that subscribes multiple topics
-// The key in the map represents the error for the corresponding message ID list.
-type AckError map[error][]MessageID
+// The key is the topic name, and the value is the ACK error and the failed message ID in this topic.
+type AckError map[string]*TopicAckError
 
 func (e AckError) Error() string {
-	return fmt.Sprintf("%v", map[error][]MessageID(e))
+	builder := strings.Builder{}
+	for topic, topicAckError := range e {
+		builder.WriteString(fmt.Sprintf("topic: %s, error: %s, messageIDs: %v\n",
+			topic, topicAckError.Err, topicAckError.MsgIDs))
+	}
+	return builder.String()
 }
 
 // Consumer is an interface that abstracts behavior of Pulsar's consumer
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -307,9 +307,8 @@ type Consumer interface {
 	// AckID the consumption of a single message, identified by its MessageID
 	AckID(MessageID) error
 
-	// AckIDList the consumption of a list of messages, identified by their MessageID
-	// NOTE: this API will acknowledge the MessageID even if it represents a single message in the batch,
-	//   i.e. the enableBatchIndexAck option will be treated as true when calling this method
+	// AckIDList the consumption of a list of messages, identified by their MessageIDs
+	// Returns a map of MessageID to error, the keys are the MessageIDs that failed to be acknowledged
 	AckIDList([]MessageID) map[MessageID]error
 
 	// AckWithTxn the consumption of a single message with a transaction
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -309,6 +309,10 @@ type Consumer interface {
 
 	// AckIDList the consumption of a list of messages, identified by their MessageIDs
 	// Returns a map of MessageID to error, the keys are the MessageIDs that failed to be acknowledged
+	// NOTE: When EnableBatchIndexAcknowledgment is false, if a message ID represents a message in the batch,
+	// it will not be actually acknowledged by broker until all messages in that batch are acknowledged via
+	// the AckID or AckIDList method.
+	// However, in this case, no error will be returned for that message ID even if AckWithResponse is true.
 	AckIDList([]MessageID) map[MessageID]error
 
 	// AckWithTxn the consumption of a single message with a transaction
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -19,6 +19,7 @@ package pulsar
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/apache/pulsar-client-go/pulsar/backoff"
@@ -266,6 +267,13 @@ type ConsumerOptions struct {
 	startMessageID *trackingMessageID
 }
 
+// This error could only be returned when calling `AckIDList` on a consumer that subscribes multiple topics
+type AckError map[MessageID]error
+
+func (e AckError) Error() string {
+	return fmt.Sprintf("%v", map[MessageID]error(e))
+}
+
 // Consumer is an interface that abstracts behavior of Pulsar's consumer
 type Consumer interface {
 	// Subscription get a subscription for the consumer
@@ -305,15 +313,25 @@ type Consumer interface {
 	Ack(Message) error
 
 	// AckID the consumption of a single message, identified by its MessageID
+	// When `EnableBatchIndexAcknowledgment` is false, if a message ID represents a message in the batch,
+	// it will not be actually acknowledged by broker until all messages in that batch are acknowledged via
+	// `AckID` or `AckIDList`.
 	AckID(MessageID) error
 
 	// AckIDList the consumption of a list of messages, identified by their MessageIDs
-	// Returns a map of MessageID to error, the keys are the MessageIDs that failed to be acknowledged
-	// NOTE: When EnableBatchIndexAcknowledgment is false, if a message ID represents a message in the batch,
-	// it will not be actually acknowledged by broker until all messages in that batch are acknowledged via
-	// the AckID or AckIDList method.
-	// However, in this case, no error will be returned for that message ID even if AckWithResponse is true.
-	AckIDList([]MessageID) map[MessageID]error
+	//
+	// This method should be used when `AckWithResponse` is true. Otherwise, it will be equivalent with calling
+	// `AckID` on each message ID in the list.
+	//
+	// If there are some invalid message IDs in the list, this method won't return an error for them, only error logs
+	// will be printed because retrying acknowledging them will not make a difference. Message IDs retrieved from
+	// Message objects are always valid.
+	//
+	// If the consumer has subscribed multiple topics or partitions, since each topic is associated with an ACK request,
+	// `AckError` will be returned if there are errors because some message IDs might succeed while others might fail.
+	// If you're sure that only 1 message is subscribed, all message IDs should be re-acknowledged when it returns an
+	// error. Otherwise, you should cast the error to `AckError` and handle it accordingly.
+	AckIDList([]MessageID) error
 
 	// AckWithTxn the consumption of a single message with a transaction
 	AckWithTxn(Message, Transaction) error
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -268,10 +268,11 @@ type ConsumerOptions struct {
 }
 
 // This error could only be returned when calling `AckIDList` on a consumer that subscribes multiple topics
-type AckError map[MessageID]error
+// The key in the map represents the error for the corresponding message ID list.
+type AckError map[error][]MessageID
 
 func (e AckError) Error() string {
-	return fmt.Sprintf("%v", map[MessageID]error(e))
+	return fmt.Sprintf("%v", map[error][]MessageID(e))
 }
 
 // Consumer is an interface that abstracts behavior of Pulsar's consumer
--- a/pulsar/consumer_multitopic.go
+++ b/pulsar/consumer_multitopic.go
@@ -170,7 +170,7 @@ func (c *multiTopicConsumer) AckID(msgID MessageID) error {
 func (c *multiTopicConsumer) AckIDList(msgIDs []MessageID) error {
 	return ackIDListFromMultiTopics(c.log, msgIDs, func(msgID MessageID) (acker, error) {
 		if !checkMessageIDType(msgID) {
-			return nil, errors.New("invalid message id type %T")
+			return nil, fmt.Errorf("invalid message id type %T", msgID)
 		}
 		if mid := toTrackingMessageID(msgID); mid != nil && mid.consumer != nil {
 			return mid.consumer, nil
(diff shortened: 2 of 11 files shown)


nodece:
This feature is good to me, I have some questions:

1. Why does the `AckIDList` not support ack the incomplete batch message, when the users use that, it is confusing. 
2. Why not refactor the ` ackIDCommon(msgID MessageID, withResponse bool, txn Transaction)`, we can use `trackingMessageID` to track the batch message. Right now , the `ackIDCommon` and `AckIDList` have the duplication code.

BewareMyPower:
### The 1st question

What's confusing is the tracking message ID itself. The existing `AckID` method's semantic is wrong. Let's look at the following example with `AckID`:

```golang
func TestMyAck(t *testing.T) {
		URL: lookupURL,
	})
	assert.Nil(t, err)
	defer client.Close()

	topic := fmt.Sprintf("test-my-ack-%v", time.Now().Nanosecond())
	createConsumer := func() Consumer {
			Topic:                       topic,
			SubscriptionName:            "my-sub",
			SubscriptionInitialPosition: SubscriptionPositionEarliest,
			Type:                        Shared,
			AckWithResponse:             true,
		})
		assert.Nil(t, err)
		return consumer
	}
	consumer := createConsumer()
	sendMessages(t, client, topic, 0, 2, true) // send 0 and 1 in the same batch
	msgs := receiveMessages(t, consumer, 2)
	for i := 0; i < 2; i++ {
		fmt.Println("Received message: ", string(msgs[i].Payload()), msgs[i].ID())
	}

	if err := consumer.AckID(msgs[0].ID()); err != nil {
		fmt.Println("Ack message 0 failed: ", err.Error())
	} else {
		fmt.Println("Ack message 0 success")
	}
	consumer.Close()

	consumer = createConsumer()
	msgs = receiveMessages(t, consumer, 1)
	fmt.Println("Received message: ", string(msgs[0].Payload()), msgs[0].ID())
}
```

Outputs:

```
Received message:  msg-0 9:0:0
Received message:  msg-1 9:0:0
Ack message 0 success
Received message:  msg-0 9:0:0
```

From the perspective from user side:
- "msg-0"'s message ID has been acknowledged successfully
- "msg-0" was delivered again after restarting the consumer

Actually this API implements `AckWithResponse` with correct semantics. Replacing the `AckID` call with:

```golang
	for msgID, err := range consumer.AckIDList([]MessageID{msgs[0].ID()}) {
		fmt.Println("Failed to acknowledge ", msgID, err.Error())
	}
```

The outputs will be:

```
Received message:  msg-0 10:0:0
Received message:  msg-1 10:0:0
Failed to acknowledge  10:0:0 incomplete batch
Received message:  msg-0 10:0:0
```

P.S. we should fix the `String()` method of message ID.

Users should add the failed message ID to the next message ID list passed to `AckIDList`. Acknowledging the same message ID 

IMO, we should make batch index ACK enabled by default for both client side and server side. The current default behavior is really confusing.

### The 2nd question

`ackIDCommon` is an abstraction for acknowledgment on a single message ID. When `AckWithResponse` is true, we cannot reuse this method because we should avoid sending N requests.

nodece:
@BewareMyPower 

1. The 1st answer

Your explanation is clear, I agree with you:

> IMO, we should make batch index ACK enabled by default for both client side and server side. The current default behavior is really confusing.

Default to batch index ACK is disabled. When disabled, the users must ack all batch messages by `AckID(id)`,  and then the client sends the ack request to the broker. 

> Received message:  msg-0 9:0:0
> Received message:  msg-1 9:0:0
> Ack message 0 success
> Received message:  msg-0 9:0:0

You only ack the first message, the client doesn't send the ack request to the broker, so you still receive the first message after restart the consumer.

I remember the Java client has the same behavior, if I am wrong, please let me know.

2. The 2nd answer.

> `ackIDCommon` is an abstraction for acknowledgment on a single message ID. When `AckWithResponse` is true, we cannot reuse this method because we should avoid sending N requests.

You can refactor the `ackIDCommon`, because you ignore the ackTracer, so it is confusing. I hope we still use the ackTracer to track all(non-batch, batch, chunk) messages, and we can keep the same semantics with the `AckID`.

You can also disscus this issue on the dev mailing list.

BewareMyPower:
>  the client doesn't send the ack request to the broker

There is no way to let the client know whether the ack request was sent. That's the point. In addition, users should never care if the ack was sent. They only care if the ack succeeded.

> I remember the Java client has the same behavior

Yeah, but this behavior is wrong. I don't want to blame the author but the error handling is really not taken care of. Happy path is always easy to write.

> because you ignore the ackTracer, so it is confusing.

Oh I see. Let me think a better way to reuse the code.

> You can also disscus this issue on the dev mailing list.

I plan to write a PIP to explain these things, including why to enable batch index ACK by default, the mess of acknowledge APIs' semantics.

BewareMyPower:
@nodece I reused the tracking message ID's tracker and used the semantics that ACK on an incomplete message ID in the batch does not return an error. Please check the latest tests. As for reusing the `ackIDCommon` method. It's still hard. There is not much to reuse except for the `ack` method call on the tracking message ID. The `ack` and `ackWithTxn` methods of `unAckChunksTracker` are also a mess.

BewareMyPower:
@nodece I reused the tracking message ID's tracker and used the semantics that ACK on an incomplete message ID in the batch does not return an error. Please check the latest tests. As for reusing the `ackIDCommon` method. It's still hard. There is not much to reuse except for the `ack` method call on the tracking message ID. The `ack` and `ackWithTxn` methods of `unAckChunksTracker` are also a mess.

pulsar/consumer.go
```diff
@@ -307,6 +307,14 @@ type Consumer interface {
 	// AckID the consumption of a single message, identified by its MessageID
 	AckID(MessageID) error
 
+	// AckIDList the consumption of a list of messages, identified by their MessageIDs
+	// Returns a map of MessageID to error, the keys are the MessageIDs that failed to be acknowledged
+	// NOTE: When EnableBatchIndexAcknowledgment is false, if a message ID represents a message in the batch,
+	// it will not be actually acknowledged by broker until all messages in that batch are acknowledged via
+	// the AckID or AckIDList method.
+	// However, in this case, no error will be returned for that message ID even if AckWithResponse is true.
+	AckIDList([]MessageID) map[MessageID]error
```
RobertIndie, line 316: Can we return the `error` directly to simplify this API?
And we can also define the error type like this to let users get more specified errors for each msg ids:
```go
type AckIDListError map[MessageID]error

func (e AckIDListError) Error() string {
    return "some message IDs failed to be acknowledged"
}
```

This way, the user can process the error like this:
```go
if err != nil {
        if ackErr, ok := err.(AckIDListError); ok {
	        for id, e := range ackErr {
		        fmt.Printf("Message ID %d: %s\n", id, e)
	        }
        }
}
```

BewareMyPower: Users should not ignore any failed message IDs, that's why I return a map error. I don't want to users to call it like:

```golang
if err := consumer.AckIDList(ids); err != nil {
        fmt.Println(err)
}
```

BewareMyPower: If it returns `error` directly, in which case should users not convert it to the `AckIDListError`?

RobertIndie: > Users should not ignore any failed message IDs, that's why I return a map error. I don't want users to call it like:

Yes. Users could handle it. And users should not call it like that.

Instead, I gave the example in my above comment:
```
if err != nil {
        if ackErr, ok := err.(AckIDListError); ok {
	        for id, e := range ackErr {
		        fmt.Printf("Message ID %d: %s\n", id, e)
	        }
        }
}
```
The users could check the failed messages ids my converting the error to `AckIDListError`.

> If it returns error directly, in which case should users not convert it to the AckIDListError?

So, if the error is nil, then the user doesn't need to conver it.

nodece: I think `map[MessageID]error` is better. Usually, I don't check the error type.

BewareMyPower: @nodece I just had a quick sync with @RobertIndie 

With the current implementation, the only case that only partial of the message ID list failed is: there are some invalid message IDs.

Therefore, we can just return a trivial `error` to tell users that all these message ID failed.

```golang
if err := consumer.AckIDList(ids); err != nil {
    if errorMap, ok := err.(AckIDListError); ok {
        for msgID, err := range errorMap {
            // Add msgID to the message ID list to ack again
        }
    } else {
        // Add the whole ids to the message ID list to retry
    }
}
```

WDYT?

RobertIndie: What about returnning an error that are not specified to the message ID level like the `ErrConsumerClosed`? It doens't make sense to make it as the message ID error.
Actually, checking the type of error should be a common practice. We have discussed this before: https://github.com/apache/pulsar-client-go/issues/1142. It is better to expose the error type to users so they can easily identify the error.

BewareMyPower: Or should we just fail the whole ids if there is any invalid message ID?

I designed the `map[MessageID]error` as the return value type because I want to handle the case that partial messages in the batch are acknowledged. But since now, they are treated as successful result, it might not be meaningful to return a map now.

pulsar/consumer_test.go
```diff
@@ -4691,3 +4691,115 @@ func TestPartitionConsumerGetLastMessageIDs(t *testing.T) {
...
+	producer.Flush()
+}
+
+func receiveMessages(t *testing.T, consumer Consumer, numMessages int) []Message {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	msgs := make([]Message, 0)
+	for i := 0; i < numMessages; i++ {
+		if msg, err := consumer.Receive(ctx); err == nil {
+			fmt.Println("Received message: ", string(msg.Payload()))
```
nodece, line 4796:
```suggestion
			t.Log("Received message: ", string(msg.Payload()))
```

pulsar/consumer_test.go
```diff
@@ -4691,3 +4691,115 @@ func TestPartitionConsumerGetLastMessageIDs(t *testing.T) {
...
+			msgs = append(msgs, msg)
+		} else {
+			fmt.Printf("Failed to receive message: %v", err)
```
nodece, line 4799:
```suggestion
			t.Logf("Failed to receive message: %v", err)
```

pulsar/consumer_test.go
```diff
@@ -4691,3 +4691,115 @@ func TestPartitionConsumerGetLastMessageIDs(t *testing.T) {
...
+	ctx := context.Background()
+	for i := 0; i < numMessages; i++ {
+		msg := &ProducerMessage{Payload: []byte(fmt.Sprintf("msg-%d", startIndex+i))}
+		if batching {
+			producer.SendAsync(ctx, msg, func(_ MessageID, _ *ProducerMessage, _ error) {})
+		} else {
+			producer.Send(ctx, msg)
+		}
+	}
+	producer.Flush()
```
nodece, line 4787: Miss error checks.

pulsar/consumer_test.go
```diff
@@ -4691,3 +4691,115 @@ func TestPartitionConsumerGetLastMessageIDs(t *testing.T) {
...
+	assert.Nil(t, err)
+	defer producer.Close()
+
+	ctx := context.Background()
+	for i := 0; i < numMessages; i++ {
+		msg := &ProducerMessage{Payload: []byte(fmt.Sprintf("msg-%d", startIndex+i))}
+		if batching {
+			producer.SendAsync(ctx, msg, func(_ MessageID, _ *ProducerMessage, _ error) {})
+		} else {
+			producer.Send(ctx, msg)
```
nodece, line 4784: Miss error checks.

pulsar/consumer_test.go
```diff
@@ -4691,3 +4691,115 @@ func TestPartitionConsumerGetLastMessageIDs(t *testing.T) {
...
+		AckWithResponse:                true,
+	})
+	assert.Nil(t, err)
+	return consumer
+}
+
+func sendMessages(t *testing.T, client Client, topic string, startIndex int, numMessages int, batching bool) {
+	producer, err := client.CreateProducer(ProducerOptions{
+		Topic:           topic,
+		DisableBatching: !batching,
```
nodece, line 4773:
```suggestion
		DisableBatching: !batching,
		BatchingMaxPublishDelay: 1 * time.Hour
```

Avoid the Avoid sending immediately.

pulsar/consumer_partition.go
```diff
@@ -683,6 +688,83 @@ func (pc *partitionConsumer) AckID(msgID MessageID) error {
...
+		} else if msgID.BatchIdx() >= 0 && msgID.BatchSize() > 0 &&
+			msgID.BatchIdx() >= msgID.BatchSize() {
+			errorMap[msgID] = fmt.Errorf("invalid batch index %v (size: %v)", msgID.BatchIdx(), msgID.BatchSize())
+		} else {
+			switch convertedMsgID := msgID.(type) {
+			case *trackingMessageID:
+				if convertedMsgID.ack() {
+					pendingAcks[newPosition(msgID)] = nil
+				} else {
+					incompleteTrackingIDs = append(incompleteTrackingIDs, convertedMsgID)
```
nodece, line 720:
```suggestion
					if pc.options.enableBatchIndexAck {
						pendingAcks[newPosition(msgID)] = convertedMsgID.tracker.getAckBitSet()
					}
```

pulsar/consumer_partition.go
```diff
@@ -683,6 +688,83 @@ func (pc *partitionConsumer) AckID(msgID MessageID) error {
...
+			}
+		}
+		return errorMap
+	}
+
+	chunkedMsgIDs := make([]*chunkMessageID, 0) // we need to remove them after acknowledging
+	pendingAcks := make(map[position]*bitset.BitSet)
+
+	// They might be complete after the whole for loop
+	incompleteTrackingIDs := make([]*trackingMessageID, 0)
```
nodece, line 706:
```suggestion
```

pulsar/consumer_partition.go
```diff
@@ -683,6 +688,83 @@ func (pc *partitionConsumer) AckID(msgID MessageID) error {
...
+	}
+
+	if pc.options.enableBatchIndexAck {
+		for _, trackingID := range incompleteTrackingIDs {
+			position := newPosition(trackingID)
+			if _, found := pendingAcks[position]; !found {
+				pendingAcks[position] = trackingID.tracker.getAckBitSet()
+			}
+		}
+	}
```
nodece, line 742:
```suggestion
```
Move this logic to `case *trackingMessageID` section.

pulsar/consumer.go
crossoverJie: > Can we return the `error` directly to simplify this API? And we can also define the error type like this to let users get more specified errors for each msg ids:
> 
> ```go
> type AckIDListError map[MessageID]error
> 
> func (e AckIDListError) Error() string {
>     return "some message IDs failed to be acknowledged"
> }
> ```
> 
> This way, the user can process the error like this:
> 
> ```go
> if err != nil {
>         if ackErr, ok := err.(AckIDListError); ok {
> 	        for id, e := range ackErr {
> 		        fmt.Printf("Message ID %d: %s\n", id, e)
> 	        }
>         }
> }
> ```

I guess many users will habitually directly judge `err != nil` and then print out the err information.

Maybe this would be better(or somthine about more infomations):

```go
type AckIDListError map[MessageID]error

func (e AckIDListError) Error() string {
	ids := make([]MessageID, 0)
	for id, err := range e {
		if err != nil{
			ids = append(ids, id)
		}
	}
	return fmt.Sprintf("some message IDs %s failed to be acknowledged", ids)
}
```

pulsar/consumer_partition.go
```diff
@@ -683,6 +688,83 @@ func (pc *partitionConsumer) AckID(msgID MessageID) error {
...
+				chunkedMsgIDs = append(chunkedMsgIDs, convertedMsgID)
+			case *messageID:
+				pendingAcks[newPosition(msgID)] = nil
+			default:
+				errorMap[msgID] = fmt.Errorf("invalid message id type %T", msgID)
+			}
+		}
+	}
+
+	if pc.options.enableBatchIndexAck {
```
shibd, line 735: A non-blocking comment for this PR:

If the `msgId` is a batch, can we record this ID in `ackGroupTracking` even when `batchIndexAck` is not enabled?

This way, even if a batch ID is not acknowledged, we can filter it out the next time the same message is received, avoiding duplicate delivery to the user.

This scenario references the issue: https://github.com/apache/pulsar/issues/23436

pulsar/consumer_test.go
BewareMyPower: This log was added for debugging, I'd like to remove it (because it could be used in new tests that many messages are sent)

pulsar/consumer_partition.go
BewareMyPower: It is a separated issue while this PR might be huge. Could you open an issue in this repo to track?

BewareMyPower:
@RobertIndie @nodece @crossoverJie @shibd Thanks for your reviews. The `AckIDList` API is updated now.

For the simple consumer case that only 1 topic is subscribed, since only 1 ACK request will be sent, these message IDs either all fail or all succeed. The only exceptional case is that invalid message IDs are passed. However, in this case, there is no way to handle these message IDs because acknowledging them will always fail. Hence I just return a trivial `error` here. 

However, if the consumer subscribes multiple topics, there might be multiple `partitionConsumer` objects here, each of them has a separated events channel (`eventsCh`) so multiple ACK requests could be sent. Therefore, an `AckError` will be returned, which contains a map that maps the failed message ID to its error.

See the API documents and the tests for details. PTAL again.

pulsar/consumer_multitopic.go
```diff
@@ -167,6 +167,47 @@ func (c *multiTopicConsumer) AckID(msgID MessageID) error {
 	return mid.consumer.AckID(msgID)
 }
 
+func (c *multiTopicConsumer) AckIDList(msgIDs []MessageID) error {
+	return ackIDListFromMultiTopics(c.log, msgIDs, func(msgID MessageID) (acker, error) {
+		if !checkMessageIDType(msgID) {
+			return nil, errors.New("invalid message id type %T")
```
shibd, line 173:
```suggestion
			return nil, errors.New("invalid message id type %T"， msgId)
```

pulsar/consumer.go
```diff
@@ -305,8 +325,26 @@ type Consumer interface {
...
+	// This method should be used when `AckWithResponse` is true. Otherwise, it will be equivalent with calling
+	// `AckID` on each message ID in the list.
+	//
+	// If there are some invalid message IDs in the list, this method won't return an error for them, only error logs
+	// will be printed because retrying acknowledging them will not make a difference. Message IDs retrieved from
+	// Message objects are always valid.
+	//
+	// If the consumer has subscribed multiple topics or partitions, since each topic is associated with an ACK request,
+	// `AckError` will be returned if there are errors because some message IDs might succeed while others might fail.
+	// If you're sure that only 1 message is subscribed, you should treat all message IDs as failed when it returns an
```
shibd, line 344: > If you're sure that only 1 message is subscribed, you should treat all message IDs as failed when it returns an error

Why is this logic needed? We could keep it simple and return AckError even if only one topic has an error.

pulsar/consumer_multitopic.go
```diff
@@ -167,6 +167,47 @@ func (c *multiTopicConsumer) AckID(msgID MessageID) error {
...
+	})
+}
+
+func ackIDListFromMultiTopics(log log.Logger, msgIDs []MessageID, findConsumer func(MessageID) (acker, error)) error {
+	consumerToMsgIDs := make(map[acker][]MessageID)
+	for _, msgID := range msgIDs {
+		if consumer, err := findConsumer(msgID); err == nil {
+			consumerToMsgIDs[consumer] = append(consumerToMsgIDs[consumer], msgID)
+		} else {
+			log.Warnf("Can not find consumer for %v", msgID)
```
shibd, line 188: Do we need to include this type of error in `AckError` as well?

pulsar/consumer_multitopic.go
```diff
@@ -167,6 +167,47 @@ func (c *multiTopicConsumer) AckID(msgID MessageID) error {
...
+			ackError[consumer.Topic()] = &TopicAckError{
+				Err:    err,
+				MsgIDs: ids,
+			}
+		}
+	}
+	if len(ackError) == 0 {
+		return nil
+	} else if len(ackError) == 1 {
+		for _, topicAckError := range ackError {
```
shibd, line 204: As I commented on the API, maybe we could keep it simple and consistently return `AckError`.

pulsar/consumer_multitopic.go
shibd: https://github.com/apache/pulsar-client-go/blob/6ffc7d283fbdb43e11d16bd643a7f28544cd9d1c/pulsar/consumer_impl.go#L564-L571

For example, if checkMsgIDPartition return an error, we need to tell the user.

shibd (commented):
Left some small comment

pulsar/consumer_multitopic.go
RobertIndie: +1.

pulsar/consumer_multitopic.go
BewareMyPower: But users can do nothing to handle invalid message IDs except for excluding them from the next `AckIDList` call.

It will make users hard to handle the error.

pulsar/consumer_multitopic.go
BewareMyPower: I simplified the `AckError` with a `map[MessageID]error` now. But I don't include invalid message IDs in this map. PTAL again @shibd @RobertIndie

pulsar/consumer_multitopic.go
shibd: Ok, Make sense to me. The user can from warn log to get this.

shibd (approved):
/LGTM

pulsar/consumer.go
BewareMyPower: I adopted this solution eventually, the difference is that I included the error message abd grouped the message IDs with the same error message.

RobertIndie (approved):

PR #1303, opened 2024-10-30
[improve] PIP-368: Support lookup based on the lookup properties

crossoverJie:
### Motivation

Support pip: https://github.com/apache/pulsar/pull/23223/

### Modifications

- Upgrade Pulsar image to 4.0.0
- Upgrade golint
- Client support LookupProperties

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes)
  - The schema: (no)
  - The wire protocol: (yes)

### Documentation


BewareMyPower:
Could you use a separated PR to fix conflicts after upgrading Pulsar to 4.0.0 like https://github.com/apache/pulsar-client-go/pull/1302?

BTW, I found the lint task in the Makefile is repeated, we use `golangci/golangci-lint-action@v6` in CI, where `make lint` is never used.

crossoverJie:
> BTW, I found the lint task in the Makefile is repeated, we use `golangci/golangci-lint-action@v6` in CI, where `make lint` is never used.

You are right, `make lint` can be used locally.

BewareMyPower (commented):
Could you use a separated PR to fix conflicts after upgrading Pulsar to 4.0.0?

In addition, we'd better use latest image by default.

crossoverJie:
> Could you use a separated PR to fix conflicts after upgrading Pulsar to 4.0.0?

Sorry, I didn't fully understand what you meant😂.

This PR has upgraded the image to 4.0.0. do you mean to create a separate PR for this part? I think the contents of this PR are not complicated, and the code review should be completed quickly.

BewareMyPower:
> This PR has upgraded the image to 4.0.0. do you mean to create a separate PR for this part?

Yes. Including the tool versions upgrade in a code PR is not a good practice.

crossoverJie:
> Yes. Including the tool versions upgrade in a code PR is not a good practice.

PTAL. https://github.com/apache/pulsar-client-go/pull/1304

pulsar/consumer_test.go
```diff
@@ -4691,3 +4691,66 @@ func TestPartitionConsumerGetLastMessageIDs(t *testing.T) {
 	}
 
 }
+
+func TestLookupConsumer(t *testing.T) {
+	client, err := NewClient(ClientOptions{
```
BewareMyPower, line 4696: What does this test try to verify? It's a trivial end-to-end test. The message properties are never related to the lookup properties.

BewareMyPower: Since the lookup properties rely on the customized lookup logic, I think it's okay to not add any test.

crossoverJie: Yes, I agree with you, this is a basic functional test of duplication to ensure that the  `LookupProperties` do not affect the existing functionality.

BewareMyPower: If so, we can just configure values in `ProducerMessage`. Configuring key and properties might make users think if lookup properties will affect them.

nodece (commented):
LGTM, could you add a test for verify the lookup properties in the `pulsar/internal/lookup_service_test.go`?

crossoverJie:
> LGTM, could you add a test for verify the lookup properties in the `pulsar/internal/lookup_service_test.go`?

Relevant code has been added https://github.com/apache/pulsar-client-go/pull/1303/files#diff-eaa8996b10b63f4ccd290c3af5dff5397e3911ab7cd228e20ecc99f0d69aeff4.

BewareMyPower (approved):

PR #1305, opened 2024-11-16
Improve the AckIDList performance when there are many topics subscribed

BewareMyPower:
### Motivation

Currently when a consumer subscribes multiple topic-partitions and `AckWithResponse` is true, the `AckIDList` method will iterate over all internal consumers **sequentially**. It harms the performance especially there are many internal consumers. For example, if the connection of an internal consumer was stuck by some reason, message IDs from other consumer would be blocked for the operation timeout.

### Modifications

In `ackIDListFromMultiTopics`, call `consumer.AckIDList` in goroutines and use a channel to receive all errors from these calls.

Add `TestMultiTopicAckIDListTimeout`, which sets a dummy connection instance whose `SendRequest` never completes the callback, to verify the `AckIDList` call will not take much more time than the operation timeout to complete. Without this improvement, it will take more than 5 times of the operation timeout to fail.

### DIFF
--- a/pulsar/consumer_multitopic.go
+++ b/pulsar/consumer_multitopic.go
@@ -190,29 +190,23 @@ func ackIDListFromMultiTopics(log log.Logger, msgIDs []MessageID, findConsumer f
 	}
 
 	subErrCh := make(chan error, len(consumerToMsgIDs))
-	errCh := make(chan error)
-	go func() {
-		ackError := AckError{}
-		for i := 0; i < len(consumerToMsgIDs); i++ {
-			err := <-subErrCh
-			if topicAckError, ok := err.(AckError); ok {
-				for id, err := range topicAckError {
-					ackError[id] = err
-				}
-			}
-		}
-		if len(ackError) == 0 {
-			errCh <- nil
-		} else {
-			errCh <- ackError
-		}
-	}()
 	for consumer, ids := range consumerToMsgIDs {
 		go func() {
 			subErrCh <- consumer.AckIDList(ids)
 		}()
 	}
-	return <-errCh
+	ackError := AckError{}
+	for i := 0; i < len(consumerToMsgIDs); i++ {
+		if topicAckError, ok := (<-subErrCh).(AckError); ok {
+			for id, err := range topicAckError {
+				ackError[id] = err
+			}
+		}
+	}
+	if len(ackError) == 0 {
+		return nil
+	}
+	return ackError
 }
 
 // AckWithTxn the consumption of a single message with a transaction
--- a/pulsar/consumer_multitopic.go
+++ b/pulsar/consumer_multitopic.go
@@ -189,25 +189,30 @@ func ackIDListFromMultiTopics(log log.Logger, msgIDs []MessageID, findConsumer f
 		}
 	}
 
-	ackError := AckError{}
-	for consumer, ids := range consumerToMsgIDs {
-		if err := consumer.AckIDList(ids); err != nil {
-			if topicAckError := err.(AckError); topicAckError != nil {
+	subErrCh := make(chan error, len(consumerToMsgIDs))
+	errCh := make(chan error)
+	go func() {
+		ackError := AckError{}
+		for i := 0; i < len(consumerToMsgIDs); i++ {
+			err := <-subErrCh
+			if topicAckError, ok := err.(AckError); ok {
 				for id, err := range topicAckError {
-					ackError[id] = err
-				}
-			} else {
-				// It should not reach here
-				for _, id := range ids {
 					ackError[id] = err
 				}
 			}
 		}
+		if len(ackError) == 0 {
+			errCh <- nil
+		} else {
+			errCh <- ackError
+		}
+	}()
+	for consumer, ids := range consumerToMsgIDs {
+		go func() {
+			subErrCh <- consumer.AckIDList(ids)
+		}()
 	}
-	if len(ackError) == 0 {
-		return nil
-	}
-	return ackError
+	return <-errCh
 }
 
 // AckWithTxn the consumption of a single message with a transaction
--- a/pulsar/consumer_multitopic_test.go
+++ b/pulsar/consumer_multitopic_test.go
@@ -18,6 +18,7 @@
 package pulsar
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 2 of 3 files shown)


geniusjoe:
I missed PR #1301 and I wonder does this `AckIDList([]MessageID) map[MessageID]error` API is an alternative to Java `void acknowledge(List<MessageId> messageIdList) throws PulsarClientException` API? Are there any differences  when I use them?

BewareMyPower:
> I wonder does this AckIDList([]MessageID) map[MessageID]error API is an alternative to Java void acknowledge(List<MessageId> messageIdList) throws PulsarClientException API? 

Not an alternative. The Go API still returns an `error`, not `map[MessageID]error`. But the error can be cast to an `AckError` (`map[MessageID]error`) so you can just retry the failed message IDs rather than the whole message IDs. While the Java API `void acknowledge(List<MessageId> messageIdList) throws PulsarClientException` does not define such an exception that you can retrieve failed message IDs and corresponding errors from it.

geniusjoe:
> Not an alternative. The Go API still returns an `error`, not `map[MessageID]error`. But the error can be cast to an `AckError` (`map[MessageID]error`) so you can just retry the failed message IDs rather than the whole message IDs. While the Java API `void acknowledge(List<MessageId> messageIdList) throws PulsarClientException` does not define such an exception that you can retrieve failed message IDs and corresponding errors from it.

So that if I don't need to check each message exception, I can just regard AckError as a normal exception? I think using `error` instead of `map[MessageID]error` is more consistent with Java API.

BewareMyPower:
> So that if I don't need to check each message exception, I can just regard AckError as a normal exception?

Yes. You can still write the following code:

```go
	if err := consumer.AckIDList(msgIDs); err != nil {
		fmt.Errorf("Failed to ack messages: %v", err)
	}
```

The design in Go client is more like to add the following exception to the Java client and document that this exception could be thrown by `acknowledge`.

```java
    public static class AckException extends PulsarClientException {
        
        private final Map<MessageId, PulsarClientException> exceptionMap;

        public AckException(Map<MessageId, PulsarClientException> exceptionMap) {
            super(toString(exceptionMap));
            this.exceptionMap = exceptionMap;
        }
        
        private static String toString(Map<MessageId, PulsarClientException> exceptionMap) {
            return ""; // TODO
        }
    }
```

geniusjoe:
> Yes. You can still write the following code:
> 
> ```go
> 	if err := consumer.AckIDList(msgIDs); err != nil {
> 		fmt.Errorf("Failed to ack messages: %v", err)
> 	}
> ```
> 
> The design in Go client is more like to add the following exception to the Java client and document that this exception could be thrown by `acknowledge`.
> 
> ```java
>     public static class AckException extends PulsarClientException {
>         
>         private final Map<MessageId, PulsarClientException> exceptionMap;
> 
>         public AckException(Map<MessageId, PulsarClientException> exceptionMap) {
>             super(toString(exceptionMap));
>             this.exceptionMap = exceptionMap;
>         }
>         
>         private static String toString(Map<MessageId, PulsarClientException> exceptionMap) {
>             return ""; // TODO
>         }
>     }
> ```

I got it, thank you very much.

pulsar/consumer_multitopic_test.go
```diff
@@ -317,3 +318,102 @@ func runMultiTopicAckIDList(t *testing.T, regex bool) {
...
+		assert.True(t, ok)
+		conn = pc._getConn()
+		pc._setConn(dummyConnection{})
+	}
+
+	start := time.Now()
+	err = consumer.AckIDList(msgIDs)
+	elapsed := time.Since(start)
+	t.Logf("AckIDList takes %v ms", elapsed)
+	assert.True(t, elapsed < 4*time.Second && elapsed >= 3*time.Second)
```
nodece, line 405:
```suggestion
	assert.True(t, elapsed < 5*time.Second && elapsed >= 3*time.Second)
```
Increase test time to avoid the flaky test.

pulsar/consumer_multitopic.go
```diff
@@ -189,25 +189,30 @@ func ackIDListFromMultiTopics(log log.Logger, msgIDs []MessageID, findConsumer f
...
 		}
+		if len(ackError) == 0 {
+			errCh <- nil
+		} else {
+			errCh <- ackError
+		}
+	}()
+	for consumer, ids := range consumerToMsgIDs {
+		go func() {
+			subErrCh <- consumer.AckIDList(ids)
```
crossoverJie, line 212:
```go
	for consumer, ids := range consumerToMsgIDs {
		consumer := consumer
		ids := ids
		go func() {
			subErrCh <- consumer.AckIDList(ids)
		}()
	}
```
Necessary to make a variable copy? Although this issue was fixed in [1.22](https://go.dev/blog/loopvar-preview), would it trigger this issue if the user is using a version less than 1.22?

https://github.com/apache/pulsar-client-go/blob/35076ac09138b5fa11a508ccc0ef8cd1b1993347/pulsar/consumer_impl.go#L779

BewareMyPower: It's already 1.22: https://github.com/apache/pulsar-client-go/blob/92c6e284cfcd1eb921116da6482b79e5500e03ae/go.mod#L3

pulsar/consumer_multitopic_test.go
```diff
@@ -317,3 +318,102 @@ func runMultiTopicAckIDList(t *testing.T, regex bool) {
...
+	if ackError, ok := err.(AckError); ok {
```
RobertIndie, line 406:
```suggestion
	var ackError AckError
	if errors.As(err, &ackError) {
```

It's better to use `errors.As` here so that we can also handle the warpped error.

pulsar/consumer_multitopic.go
```diff
@@ -189,25 +189,30 @@ func ackIDListFromMultiTopics(log log.Logger, msgIDs []MessageID, findConsumer f
 		}
 	}
 
-	ackError := AckError{}
-	for consumer, ids := range consumerToMsgIDs {
-		if err := consumer.AckIDList(ids); err != nil {
-			if topicAckError := err.(AckError); topicAckError != nil {
+	subErrCh := make(chan error, len(consumerToMsgIDs))
+	errCh := make(chan error)
+	go func() {
```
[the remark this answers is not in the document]
BewareMyPower: Yes. I used to use the goroutine to support timeout mechanism, e.g.

```go
	select {
	case <-time.After(5 * time.Second):
		return errors.New("timeout waiting for ackIDList")
	case <-errCh:
		return <-errCh
	}
```

But it seems unnecessarily complicated because each `subErrCh` has its own timeout. So it's now unnecessary

nodece (approved):

RobertIndie (approved):

PR #1308, opened 2024-11-19
Fix close blocked

nodece:
### Motivation

When a consumer or producer reconnects to the broker, the event loop handling the reconnection becomes occupied, preventing the execution of the `close` command. This can result in the inability to properly close the consumer or producer, leading to potential resource management issues.

### Modifications

- Introduce passing a `context` to the retry function. This ensures that when the `context` is completed or canceled, the retry logic is interrupted and stops further attempts.
- Modify the `producer.Close` and `consumer.Close` methods to call the associated `context.CancelFunc` to cancel the reconnection attempts.


crossoverJie (approved):
LGTM

PR #1310, opened 2024-11-21
Fix possible deadlock when AckWithResponse is true due to queueCh is full

BewareMyPower:
### Motivation

When `AckWithResponse` is enabled, a deadlock could happen easily if `queueCh` and `messagesCh` of `partitionConsumer` are full. In this case, if the consumer receives new messages from broker, `MessageReceived` will be blocked at https://github.com/apache/pulsar-client-go/blob/9366a0eaff772b7097ab1bb2147e287460dc61ba/pulsar/consumer_partition.go#L1376

The stacks could be:

```
3  0x0000000100ea83c0 in github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).MessageReceived
   at ./consumer_partition.go:1376
4  0x0000000100b1c73c in github.com/apache/pulsar-client-go/pulsar/internal.(*connection).handleMessage
   at ./internal/connection.go:751
5  0x0000000100b1ab7c in github.com/apache/pulsar-client-go/pulsar/internal.(*connection).internalReceivedCommand
   at ./internal/connection.go:590
```

As shown in the stacks above, the `connection.run()` goroutine is blocked so that it could not handle any new command anymore, including the ACK response. Then the ACK related method will fail with "request timed out". The deadlock cannot be resolved unless the consumer peeks new messages again to make `messagesCh` not full, then `queueCh` will move messages to `messagesCh` so that `queueCh` will be not full.

The root cause is that `queueCh` is a buffered channel that has a fixed size of `ReceiverQueueSize`. However, the broker could dispatch more messages than the `ReceiverQueueSize` because the permits in Flow requests only limits the number of entries to read, not the number of messages. Hence this issue could be easily reproduced by reducing the `ReceiverQueueSize` and sending many messages with a great batch size. See how `TestAckResponseNotBlocked` reproduces this issue.

### Modifications

Add a `list` of messages (`pendingMessages`) to support queueing unlimited number of messages for `partitionedConsumer`. The flow control is actually controlled by the `availablePermits` and related Flow requests so that the queue size won't be too large (it depends on the batch size).

Add two channels (`queueInCh` and `queueOutCh`) for the following loops:
- Run a background goroutine to send messages from `queueInCh -> pendingMessages -> queueOutCh`
- In `MessageReceived`, send the received message to `queueInCh` so that the message will be queued to `pendingMessages`. **It could never be blocked.**
- In `dispatcher()`, poll messages from `queueOutCh`, which polls the 1st message from `pendingMessages`.

The background goroutine will exist by `close(pc.closeCh)` when the consumer is closed. It happens after completing the `CloseConsumer` RPC in `internalClose` so there should not be a `MessageReceived` call that sends messages to the closed `queueInCh`.

Add `TestAckResponseNotBlocked` to verify it works with a very small receiver queue size config.


pulsar/consumer_partition.go
```diff
@@ -1564,7 +1578,7 @@ func (pc *partitionConsumer) dispatcher() {
 				pc.log.Debug("skip dispatching messages when seeking")
 			}
 		} else {
-			queueCh = pc.queueCh
+			queueCh = pc.queueCh.outCh
```
fracasula, line 1572: I find this code a bit convoluted. Can you help me understand why you assign the channel to `queueCh` and then later range over `pc.queueCh` directly if you get a signal to clear the queue?

When is the variable needed and when is not is not very clear. Perhaps a comment might help demistify this?

pulsar/consumer_partition.go
```diff
@@ -1133,6 +1138,15 @@ func (pc *partitionConsumer) internalAckList(request *ackListRequest) {
 }
 
 func (pc *partitionConsumer) MessageReceived(response *pb.CommandMessage, headersAndPayload internal.Buffer) error {
+	defer func() {
+		if r := recover(); r != nil {
+			if err, ok := r.(error); ok && err.Error() == "send on closed channel" {
+				pc.log.WithField("panic", r).Debug("panic recovered in MessageReceived")
```
fracasula: If this were to happen it'd be quite bad in my opinion. Do you really want to raise is as a `DEBUG` only?
I would want to know if you're sending on a closed channel in production code. I suggest a `WARNING` here.

pulsar/consumer_partition.go
```diff
@@ -949,6 +949,11 @@ func (pc *partitionConsumer) Close() {
 
 	// wait for request to finish
 	<-req.doneCh
+
+	// It will close `queueCh.in`. If `MessageReceived` was called after that, it will panic because new messages
+	// will be sent to a closed channel. However, generally it's impossible because the broker will not be able to
+	// dispatch messages to this consumer after receiving the close request.
+	pc.queueCh.stop()
```
fracasula, line 956: I see that you're not closing `closeCh` here but you're just closing `inCh` and `outCh`.

Beware because `closeCh` is an unbuffered channel and once you call `stop` the first time the listener of that channel will return so any subsequent `c.closeCh <- struct{}{}` is going to block forever and it will be hard to detect.

Wouldn't it be better to just close `closeCh`? It will have the same side effect but at least if you `stop` twice, you'll catch it instead of silently being stuck. Given that you're not planning to call `stop` multiple times, wouldn't it be better to close the channel then? Wdyt?

fracasula (commented):
I reviewed this PR because it was not in a draft state when I started.

I wanted to comment also that perhaps using two channels `queueInCh` and `queueOutCh` might help make the code more understandable and easier to maintain.

What it's being proposed here is to use a more complex object with an embedded list, 3 channels and its own go routine running in the background.

pulsar/consumer_partition.go
BewareMyPower: > I wanted to comment also that perhaps using two channels queueInCh and queueOutCh might help make the code more understandable and easier to maintain.

I agree.

> When is the variable needed and when is not is not very clear.

It's only used here:

```go
		case msg, ok := <-queueCh:
			if !ok {
				return
			}

			queueMsg = msg
```

It polls a message from `queueCh`. If the case is hit, then in the next loop, `queueCh` will be nil and it won't try to poll a message from `queueCh`. Instead, `nextMessage` will be the wrapped message so that it will reach here:

```go
		case messageCh <- nextMessage:
			queueMsg = nil
```

I admit it's a bit complicated. The current design comes from https://github.com/apache/pulsar-client-go/pull/86. I'd like not to do some refactoring in this PR so I just changed the actual message channel from `queueCh` to `queueCh.outCh`.

And just as you mentioned, I agree that using two channels will be more clear so I will make some changes.

pulsar/consumer_partition.go
BewareMyPower: I removed this part because it should not reach here. If it reached here, it means `MessageReceived` is called on a closed `partitionConsumer`, which does not make sense because the consumer should unregister itself from the connection.

So I'd like to fail fast rather than recovering from the panic and print a warning message.

pulsar/consumer_partition.go
fracasula: Yeah I agree a refactoring here is beyond the scope of this PR. Perhaps consider adding a comment explaining what you did just here. Might help for future reverse engineering.

BewareMyPower:
Hi @fracasula, thanks for your comments and I've addressed all of them. You can check out the design from updated ** Modifications** part in the PR description.

@shibd @RobertIndie Please also take a look.

shibd:
hi, @BewareMyPower I put our discuss on here.

Before #1283, There won't be the issue mentioned in this PR because we can ensure that the pulled messages will never overflow `queueCh`. Previously, it was an **array**, and if an **entry** had batch messages, they were stored as an **array**.

Furthermore, the `availablePermits` mechanism will ensure that the number of pulled entries does not exceed the size set by `pendingQueueSize`.

shibd:
I have revert #1283 and copy unit test from this PR: https://github.com/apache/pulsar-client-go/pull/1311

This is an alternative solution to the issue.

The current PR offers a method similar to an `unbounded` channel.

The PR #1311 maintains the initial behavior, using `availablePermits` to ensure the channel is never filled.

We can compare which approach is more user-friendly. From my perspective, #1311 is simpler and the code is more readable.

PR #1323, opened 2024-12-27
[fix] [client] fix same producer/consumer use more than one connection per broker

shibd:
### Motivation

This is a catch up for https://github.com/apache/pulsar/pull/21144

When a producer or consumer reconnects, a random number will be generated as the key suffix in `ConnectionPool` to create or get the `Connection` object from the pool.

https://github.com/apache/pulsar-client-go/blob/ffba2a8fd2781f536bc03ec1408f82f6e4ea4d3a/pulsar/internal/connection_pool.go#L159-L160

### Modifications
- Change GetConnection() of connection_pool and add params: keySuffix
- Generate once keySuffix for consumer/producer when create it.
- When reconnecting, use this keySuffix to ensure obtaining the same connection

### Verifying this change
- Add TestSelectConnectionForSameConsumer and TestSelectConnectionForSameProducer to cover this.

### DIFF
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -5038,3 +5038,31 @@ func TestClientVersion(t *testing.T) {
 	assert.True(t, strings.HasSuffix(publisher.ClientVersion, "-test-client"))
 
 }
+
+func TestSelectConnectionForSameConsumer(t *testing.T) {
+	client, err := NewClient(ClientOptions{
+		URL:                     serviceURL,
+		MaxConnectionsPerBroker: 10,
+	})
+	assert.NoError(t, err)
+	defer client.Close()
+
+	topicName := newTopicName()
+
+	_consumer, err := client.Subscribe(ConsumerOptions{
+		Topic:            topicName,
+		SubscriptionName: "sub-1",
+		Type:             Shared,
+	})
+	assert.NoError(t, err)
+	defer _consumer.Close()
+
+	partitionConsumerImp := _consumer.(*consumer).consumers[0]
+	conn := partitionConsumerImp._getConn()
+
+	for i := 0; i < 5; i++ {
+		partitionConsumerImp.grabConn("")
+		assert.Equal(t, conn.ID(), partitionConsumerImp._getConn().ID(),
+			"The consumer uses a different connection when reconnecting")
+	}
+}
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -5058,11 +5058,11 @@ func TestSelectConnectionForSameConsumer(t *testing.T) {
 	defer _consumer.Close()
 
 	partitionConsumerImpl := _consumer.(*consumer).consumers[0]
-	conn := partitionConsumerImp._getConn()
+	conn := partitionConsumerImpl._getConn()
 
 	for i := 0; i < 5; i++ {
-		assert.NoError(t, partitionConsumerImp.grabConn(""))
-		assert.Equal(t, conn.ID(), partitionConsumerImp._getConn().ID(),
+		assert.NoError(t, partitionConsumerImpl.grabConn(""))
+		assert.Equal(t, conn.ID(), partitionConsumerImpl._getConn().ID(),
 			"The consumer uses a different connection when reconnecting")
 	}
 }
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -5057,11 +5057,11 @@ func TestSelectConnectionForSameConsumer(t *testing.T) {
 	assert.NoError(t, err)
 	defer _consumer.Close()
 
-	partitionConsumerImp := _consumer.(*consumer).consumers[0]
+	partitionConsumerImpl := _consumer.(*consumer).consumers[0]
 	conn := partitionConsumerImp._getConn()
 
 	for i := 0; i < 5; i++ {
-		partitionConsumerImp.grabConn("")
+		assert.NoError(t, partitionConsumerImp.grabConn(""))
 		assert.Equal(t, conn.ID(), partitionConsumerImp._getConn().ID(),
 			"The consumer uses a different connection when reconnecting")
 	}
--- a/pulsar/internal/connection_pool.go
+++ b/pulsar/internal/connection_pool.go
@@ -38,7 +38,7 @@ type ConnectionPool interface {
 	GetConnections() map[string]Connection
 
 	// GenerateRoundRobinIndex generates a round-robin index.
-	GenerateRoundRobinIndex() int32
+	GenerateRoundRobinIndex() uint32
 
 	// Close all the connections in the pool
 	Close()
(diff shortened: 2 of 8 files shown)


shibd:
> I see the ACK requests still use pc._getConn() and RequestOnCnx or RequestOnCnxNoWait in some places, should they use RequestWithCnxKeySuffix as well?

No need, in [grabConn](https://github.com/apache/pulsar-client-go/blob/a8713050d55c21140c48cb30d3e5ae6cc04e938f/pulsar/consumer_partition.go#L1990) will reset conx

pulsar/consumer_test.go
```diff
@@ -5038,3 +5038,31 @@ func TestClientVersion(t *testing.T) {
...
+
+	_consumer, err := client.Subscribe(ConsumerOptions{
+		Topic:            topicName,
+		SubscriptionName: "sub-1",
+		Type:             Shared,
+	})
+	assert.NoError(t, err)
+	defer _consumer.Close()
+
+	partitionConsumerImp := _consumer.(*consumer).consumers[0]
```
RobertIndie, line 5060:
```suggestion
	partitionConsumerImpl := _consumer.(*consumer).consumers[0]
```

pulsar/consumer_test.go
```diff
@@ -5038,3 +5038,31 @@ func TestClientVersion(t *testing.T) {
...
+	conn := partitionConsumerImp._getConn()
+
+	for i := 0; i < 5; i++ {
+		partitionConsumerImp.grabConn("")
```
RobertIndie, line 5064:
```suggestion
		assert.NoError(t, partitionConsumerImp.grabConn(""))
```

pulsar/internal/connection_pool.go
```diff
@@ -84,9 +87,11 @@ func NewConnectionPool(
 	return p
 }
 
-func (p *connectionPool) GetConnection(logicalAddr *url.URL, physicalAddr *url.URL) (Connection, error) {
-	p.log.WithField("logicalAddr", logicalAddr).WithField("physicalAddr", physicalAddr).Debug("Getting pooled connection")
-	key := p.getMapKey(logicalAddr, physicalAddr)
+func (p *connectionPool) GetConnection(logicalAddr *url.URL, physicalAddr *url.URL,
+	keySuffix int32) (Connection, error) {
+	p.log.WithField("logicalAddr", logicalAddr).
+		WithField("physicalAddr", physicalAddr).Debug("Getting pooled connection")
```
RobertIndie, line 93:
```suggestion
	p.log.WithField("logicalAddr", logicalAddr).
		WithField("physicalAddr", physicalAddr).
		WithField("keySuffix", keySuffix).Debug("Getting pooled connection")
```
Better to add a new field `keySuffix` to help debug.

pulsar/internal/connection_pool.go
```diff
@@ -141,6 +146,14 @@ func (p *connectionPool) GetConnections() map[string]Connection {
 	return conns
 }
 
+func (p *connectionPool) GenerateRoundRobinIndex() int32 {
+	cnt := atomic.AddInt32(&p.roundRobinCnt, 1)
+	if cnt < 0 {
+		cnt = -cnt
+	}
+	return cnt % p.maxConnectionsPerHost
+}
```
RobertIndie, line 155:
```suggestion
func (p *connectionPool) GenerateRoundRobinIndex() int32 {
	return atomic.AddUint32(&p.roundRobinCnt, 1) % p.maxConnectionsPerHost
}
```

`cnt = -cnt` is incorrect. I think you're trying to handle the situation where `uint32` overflow results in a negative number. However, if `cnt` overflows to `math.MinInt32`, negating it will still result in a negative number.

Could you make `roundRobinCnt` as uint32, so that we can simplify this method.

BewareMyPower:
Yes, the consumer's connection is called after the Subscribe RPC is done. 

When a partition consumer (`pc`) receives a `CLOSE_CONSUMER` command,
1. In `connection.handleCloseConsumer`, it calls `consumer.ConnectionClosed(closeConsumer)`, which sends a `connectionClosed` event to `pc.connectCloseCh`.
2. In `pc.runEventsLoop`, `reconnectToBroker` is called, **all** requests sent via `pc._getConn()` will wait until this method is done.

What I have concern is, before the `reconnectToBroker` is done, the `pc._getConn()` will return a non-nil outdated connection. If `_getConn()` is not called in `runEventsLoop()`, it might return an outdated connection.

There is another possible case that `internalXxx` could be called before `reconnectToBroker` is called, i.e. `pc.eventsCh` receives an event before `pc.connectClosedCh`.

Ideally, I think the best solution is to call `pc._setConn(nil)` before sending the `connectionClosed` event in `ConnectionClosed` and check if `_getConn` is nil in `RequestOnCnx` or `RequestOnCnxNoWait`.

pulsar/internal/connection_pool.go
BewareMyPower: I'm wondering why not just generate a random integer?

shibd: Here just keep original logic: https://github.com/apache/pulsar-client-go/blob/ffba2a8fd2781f536bc03ec1408f82f6e4ea4d3a/pulsar/internal/connection_pool.go#L154-L161

shibd:
hi, @BewareMyPower 

1. Even if we set the `connection` to null when receive cmd `ConsumerClosed`, we cannot completely guarantee that other API calls won't use the old connection to send to the old broker. Setting the connection to null and performing checks can help mitigate this issue to some extent.
2. If the `connection` is null when processing API requests (e.g., getLastMessageId), we should have an internal retry mechanism instead of throwing an exception to the client, as connection reconnection should be transparent to the caller.


In this PR, I will throw an error when the connection is null. In a future PR, I will implement the retry logic.

BewareMyPower:
I just think it again. Resetting `cnx` with `nil` does not bring much benefits for the current implementation because Go client does not have the internal retry mechanism. So it's better to improve it in another PR

pulsar/internal/rpc_client.go
```diff
@@ -64,6 +64,9 @@ type RPCClient interface {
 	RequestToHost(serviceNameResolver *ServiceNameResolver, requestID uint64,
 		cmdType pb.BaseCommand_Type, message proto.Message) (*RPCResult, error)
 
+	RequestWithCnxKeySuffix(logicalAddr *url.URL, physicalAddr *url.URL, cnxKeySuffix int32, requestID uint64,
```
nodece, line 67: @shibd You can change the `Request` method to avoid introducing an unnecessary method because the `PRCClient` is an internal interface. I checked the usage of the `Request` method and found one call now.

shibd: Lookup will use this method, And separate them to make the meaning of the two methods more direct.

nodece (approved):
LGTM

RobertIndie:
> In this PR, I will throw an error when the connection is null. In a future PR, I will implement the retry logic.

+1. I don't think we need to reimplement the retry logic. If there are connection errors when handling API requests, we shouldn't throw an error. We can just put it back to the event queue and wait for the reconnection to complete.

RobertIndie (approved):

PR #1336, opened 2025-02-24
Fix: Potential data/write conflicts

### LINKED ISSUE #1335: [Bug][Connection] Potential data/write conflicts
#### Expected behavior

write calls of one connection should be sequential

#### Actual behavior

write calls of one connection are invoked in 2 goroutine, which may lead to data/write conflict

#### Steps to reproduce

Code review:
```go
go func() {
		for {
			select {
			case <-c.closeCh:
				c.failLeftRequestsWhenClose()
				return

			case req := <-c.incomingRequestsCh:
				if req == nil {
					return // TODO: this never gonna be happen
				}
				c.internalSendRequest(req)
			}
		}
	}()

	for {
		select {
		case <-c.closeCh:
			return

		case cmd := <-c.incomingCmdCh:
			c.internalReceivedCommand(cmd.cmd, cmd.headersAndPayload)
		case data := <-c.writeRequestsCh:
			if data == nil {
				return
			}
			c.internalWriteData(data)

		case <-pingSendTicker.C:
			c.sendPing()
		}
	}
```

`c.internalSendRequest(req)` and `c.internalWriteData(data)/c.sendPing()` are invoked in 2 goroutines, and they will call `c.cnx.Write()` finally.
#### System configuration
**Pulsar version**: x.y

gunli:
Fixes #1335

Master Issue: #1335

### Motivation

Sync all the write calls of a connection into one goroutine to avoid potential data/write conflicts

### Modifications

Sync all the write calls of a connection into one goroutine.

### Verifying this change

No

@shibd @nodece @RobertIndie @BewareMyPower @crossoverJie 

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -44,14 +44,15 @@ import (
 	"github.com/apache/pulsar-client-go/pulsaradmin/pkg/admin/config"
 	"github.com/apache/pulsar-client-go/pulsaradmin/pkg/utils"
 
-	"github.com/apache/pulsar-client-go/pulsar/crypto"
-	"github.com/apache/pulsar-client-go/pulsar/internal"
-	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
-	plog "github.com/apache/pulsar-client-go/pulsar/log"
 	"github.com/google/uuid"
 	"github.com/pierrec/lz4/v4"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/protobuf/proto"
+
+	"github.com/apache/pulsar-client-go/pulsar/crypto"
+	"github.com/apache/pulsar-client-go/pulsar/internal"
+	pb "github.com/apache/pulsar-client-go/pulsar/internal/pulsar_proto"
+	plog "github.com/apache/pulsar-client-go/pulsar/log"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


gunli:
This bug has been discussed in #1333.
@shibd @nodece @RobertIndie @BewareMyPower @crossoverJie

nodece (approved):
LGTM

shibd (changes requested):
thanks for your contribution.

I don't understand what the issue is with different goroutines calling `ctx.write`. Didn't we add a lock before writing?

https://github.com/apache/pulsar-client-go/blob/024e230d32fc6a0d5586b8c8bc6b34a208b9e9d0/pulsar/internal/connection.go#L518-L519

If there is a issue, a single client may have multiple consumer and producer objects(many goroutines), which might use the same connection. How can we ensure they are all in the same goroutine?

BTW: Would it be possible to add unit tests to help reproduce the issue instead of guessing?

gunli:
> I don't understand what the issue is with different goroutines calling `ctx.write`

It is the same as writing a file in different threads, it will lead to data conflict.

nodece:
@shibd Writing data and listening for the close event in separate goroutines can lead to inconsistencies in the response. This way may cause incorrect responses due to the race condition between the broker's successful response and the connection close.

For example:
- **Time1**: The broker responds with success.
- **Time2**: The connection is closed.
- **Result**: Depending on timing, you may receive a 'success' response or an 'error: closed' response, which can lead to unexpected behavior."

BewareMyPower (approved):

BewareMyPower:
In addition to @nodece's explanation, currently the `internalWriteData` method could be called in two goroutines, which might have some underlying issues.

```go
	go func() {
		for {
			select {
			/* ... */
			case req := <-c.incomingRequestsCh:
				/* ... */
				c.internalSendRequest(req) // [1]
			}
		}
	}()

	for {
		select {
		/* ... */
		case data := <-c.writeRequestsCh:
			/* ... */
			c.internalWriteData(data) // [2]
```

gunli:
IT failed again, but it passed on my host, I think it is nothing to do with this PR.
image

nodece:
@gunli Please check CI.

nodece:
@gunli The failure of `TestAvailablePermitsLeak` occurs because when a consumer with `ConsumerCryptoFailureActionDiscard` receives a message, it sends an acknowledgment command to the broker. However, since both read and write operations are handled within the same goroutine, the write operation can block the read operation. This results in a deadlock-like situation where the read operation cannot proceed, ultimately leading to a `context.DeadlineExceeded` error.  

To resolve this issue, read and write operations should be handled in separate goroutines to avoid blocking and ensure smooth message processing.

gunli:
> @gunli The failure of `TestAvailablePermitsLeak` occurs because when a consumer with `ConsumerCryptoFailureActionDiscard` receives a message, it sends an acknowledgment command to the broker. However, since both read and write operations are handled within the same goroutine, the write operation can block the read operation. This results in a deadlock-like situation where the read operation cannot proceed, ultimately leading to a `context.DeadlineExceeded` error.
> 
> To resolve this issue, read and write operations should be handled in separate goroutines to avoid blocking and ensure smooth message processing.

OK, I will create a new PR to fix this, it seems `connection.incomingCmdCh` is redundant, because reading is run in a seperate goroutine: `go c.reader.readFromConnection()`.

gunli:
Please merge #1343 first, I will rebase and update this PR after that.

gunli:
mark it as draft until #1343 is merged, I will rebase and update this PR after that.

crossoverJie (approved):

PR #1340, opened 2025-02-26
Fix wrong result of reader.hasNext/Next after seeking by id or time

shibd:
There are **three attributes** to handle seek/filter/subscribe for Reader.

1. `startMessageID`: The start message ID, which can be specified by the user when creating the reader. It can be set to a specific position, or to the `earliest` or `latest`. When creating a consumer, this is [set in the request](https://github.com/apache/pulsar-client-go/blob/9aeeb4e7bbfcaa4039d66ab2857888be2d6fcaf3/pulsar/consumer_partition.go#L1979), and the broker uses it to set mark delete position.
2. `lastDequeuedMsg`: The most recent message dispatched from the consumer's internal queue to the user. **If it's nil, it means no messages have been delivered to the user. If there is a value, it indicates the user has consumed up to this point.**
3. `lastMessageInBroker`: The ID of the last message in the topic on the broker.

### Question-1: How to determine hasNext?

[Refer to this code](https://github.com/apache/pulsar-client-go/blob/9aeeb4e7bbfcaa4039d66ab2857888be2d6fcaf3/pulsar/consumer_partition.go#L2224)

In simple terms, to minimize interaction with the broker:

1. If `lastDequeuedMsg` has a value, it clearly indicates consumption up to this position. Compare this `lastDequeuedMsg with lastMessageInBroker`.
2. If there is no value, compare `startMessageID with lastMessageInBroker`.

The final comparison result, **if true**, indicates there really are messages. **If false**, it doesn’t necessarily mean there are no messages, because `lastMessageInBroker` is a cache, and the topic may have more messages already written. Therefore, **lastMessageInBroker is updated from the broker again for reevaluation**

### Question-2: What happens when the consumer reconnects?

When the consumer reconnects, it clears all messages in the `receive queue` and **updates startMessageId with the first message in the queue**. This indicates that the user has been consumed up to this position.

### Question-3: How does the seek operation work?

**For the broker:**
When the client sends a seek request to the broker, the broker will do the following:
1. Rewind the cursor to the seek position (ID or time).
2. Send a close request to the consumer.
3. Send a respond that the seek operation is complete.

**For the client:**
1. Send a seek request. Wait for the broker to respond that the seek is complete
2. Set `lastDequeuedMsg` to null because after seeking, this value will be invalid. **(Tip: Currently missing, this PR fixes it.)**
3. Clear the queue to avoid resetting `startMessageId` to the first position of the queue upon reconnection. **(Tip: Currently missing, this PR fixes it.)**
4. If seekByID:
    - If the passed ID is a specific position or earliest, set startMessageId to seek Id. **(Tip: Currently missing, this PR fixes it.)**
    - If the passed ID is latest, obtain the topic `lastMessageId` and set startMessageId to it. **(Tip: Currently missing, a separate PR will fix this later as it [currently doesn’t support directly passing latest id](https://github.com/apache/pulsar-client-go/blob/9aeeb4e7bbfcaa4039d66ab2857888be2d6fcaf3/pulsar/reader_impl.go#L210-L214).)**
5. If seekByTime, the client doesn’t know the messageId corresponding to the time during the seek, it needs to call getLastMessageId and set markdeleteposition to startMessageId. **(Tip: Currently missing, this PR fixes it.)**
6. Seek complete, then return to the user.
7. BTW: Actually, this is just the completion of the seek operation. The real completion requires waiting for the consumer to successfully reconnect to the broker, but this happens in the background and the user doesn't need to be concerned about it. Also, **note that we don't need to wait for the consumer to reconnect to the broker before returning the seek completion to the client**, because when the broker responds that the seek is complete, **it has already reset the cursor.** After the client receives the response, it has cleared the queue, so no previous messages will come in. In extremely rare cases (**where data from the old connection's buffer re-enters the queue after clearing**), [we will filter these messages using startMessageId to filter message ID](https://github.com/apache/pulsar-client-go/blob/9aeeb4e7bbfcaa4039d66ab2857888be2d6fcaf3/pulsar/consumer_partition.go#L1475-L1486).

### Modifications

You can run tests `TestReaderWithSeekByID` and `TestReaderWithSeekByTime` to understand the user-side issues fixed by this PR. Simply put, after seeking, calling `hasNext`, `Next` again should meet expectations.

### Verifying this change
1. Add TestReaderWithSeekByID and TestReaderWithSeekByTime to cover this change.
2. I did not change any other test, other tests passed mean not introducing any break change.

- No

### Documentation
- No

### DIFF
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -628,19 +628,19 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 	return res.msgID, err
 }
 
-func (pc *partitionConsumer) getLastMessageIDAndMarkDeletePosition() (*getLastMsgResult, error) {
+func (pc *partitionConsumer) getLastMessageIDAndMarkDeletePosition() (*getLastMsgIDResult, error) {
 	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
 		pc.log.WithField("state", state).Error("Failed to getLastMessageID for the closing or closed consumer")
 		return nil, errors.New("failed to getLastMessageID for the closing or closed consumer")
 	}
 	bo := pc.backoffPolicyFunc()
-	request := func() (*getLastMsgResult, error) {
+	request := func() (*getLastMsgIDResult, error) {
 		req := &getLastMsgIDRequest{doneCh: make(chan struct{})}
 		pc.eventsCh <- req
 
 		// wait for the request to complete
 		<-req.doneCh
-		res := &getLastMsgResult{req.msgID, req.markDeletePosition}
+		res := &getLastMsgIDResult{req.msgID, req.markDeletePosition}
 		return res, req.err
 	}
 
@@ -1740,7 +1740,7 @@ type getLastMsgIDRequest struct {
 	err                error
 }
 
-type getLastMsgResult struct {
+type getLastMsgIDResult struct {
 	msgID              *trackingMessageID
 	markDeletePosition *trackingMessageID
 }
@@ -2307,12 +2307,12 @@ func convertToMessageID(id *pb.MessageIdData) *trackingMessageID {
 		messageID: &messageID{
 			ledgerID:  int64(id.GetLedgerId()),
 			entryID:   int64(id.GetEntryId()),
-			batchIdx:  -1,
+			batchIdx:  id.GetBatchIndex(),
 			batchSize: id.GetBatchSize(),
 		},
 	}
-	if id.GetBatchSize() > 1 {
-		msgID.batchIdx = id.GetBatchIndex()
+	if msgID.batchIdx == -1 {
+		msgID.batchIdx = 0
 	}
 
 	return msgID
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -2230,7 +2230,7 @@ func (pc *partitionConsumer) hasNext() bool {
 	// If a seek by time has been performed, then the `startMessageId` becomes irrelevant.
 	// We need to compare `markDeletePosition` and `lastMessageId`,
 	// and then reset `startMessageID` to `markDeletePosition`.
-	if pc.hasSoughtByTime.CompareAndSwap(true, false) {
+	if pc.lastDequeuedMsg == nil && pc.hasSoughtByTime.CompareAndSwap(true, false) {
 		res, err := pc.getLastMessageIDAndMarkDeletePosition()
 		if err != nil {
 			pc.log.WithError(err).Error("Failed to get last message id")
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -2234,6 +2234,7 @@ func (pc *partitionConsumer) hasNext() bool {
 		res, err := pc.getLastMessageIDAndMarkDeletePosition()
 		if err != nil {
 			pc.log.WithError(err).Error("Failed to get last message id")
+			pc.hasSoughtByTime.CompareAndSwap(false, true)
 			return false
 		}
 		pc.lastMessageInBroker = res.msgID
--- a/pulsar/consumer_partition.go
+++ b/pulsar/consumer_partition.go
@@ -190,9 +190,13 @@ type partitionConsumer struct {
 	backoffPolicyFunc    func() backoff.Policy
 
 	dispatcherSeekingControlCh chan struct{}
-	isSeeking                  atomic.Bool
-	ctx                        context.Context
-	cancelFunc                 context.CancelFunc
+	// handle to the dispatcher goroutine
+	isSeeking atomic.Bool
+	// After executing seekByTime, the client is unaware of the startMessageId.
+	// Use this flag to compare markDeletePosition with BrokerLastMessageId when checking hasMoreMessages.
+	hasSoughtByTime atomic.Bool
+	ctx             context.Context
+	cancelFunc      context.CancelFunc
 }
 
 // pauseDispatchMessage used to discard the message in the dispatcher goroutine.
@@ -429,11 +433,12 @@ func newPartitionConsumer(parent Consumer, client *client, options *partitionCon
 
 	startingMessageID := pc.startMessageID.get()
 	if pc.options.startMessageIDInclusive && startingMessageID != nil && startingMessageID.equal(latestMessageID) {
-		msgID, err := pc.requestGetLastMessageID()
+		msgIDResp, err := pc.requestGetLastMessageID()
 		if err != nil {
 			pc.Close()
 			return nil, err
 		}
+		msgID := convertToMessageID(msgIDResp.GetLastMessageId())
 		if msgID.entryID != noMessageEntry {
 			pc.startMessageID.set(msgID)
 
@@ -616,18 +621,27 @@ func (pc *partitionConsumer) internalUnsubscribe(unsub *unsubscribeRequest) {
 }
 
 func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
+	res, err := pc.getLastMessageIDAndMarkDeletePosition()
+	if err != nil {
+		return nil, err
+	}
+	return res.msgID, err
+}
+
+func (pc *partitionConsumer) getLastMessageIDAndMarkDeletePosition() (*getLastMsgResult, error) {
 	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
 		pc.log.WithField("state", state).Error("Failed to getLastMessageID for the closing or closed consumer")
 		return nil, errors.New("failed to getLastMessageID for the closing or closed consumer")
 	}
 	bo := pc.backoffPolicyFunc()
-	request := func() (*trackingMessageID, error) {
+	request := func() (*getLastMsgResult, error) {
 		req := &getLastMsgIDRequest{doneCh: make(chan struct{})}
 		pc.eventsCh <- req
 
 		// wait for the request to complete
 		<-req.doneCh
-		return req.msgID, req.err
+		res := &getLastMsgResult{req.msgID, req.markDeletePosition}
+		return res, req.err
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), pc.client.operationTimeout)
@@ -647,10 +661,16 @@ func (pc *partitionConsumer) getLastMessageID() (*trackingMessageID, error) {
 
 func (pc *partitionConsumer) internalGetLastMessageID(req *getLastMsgIDRequest) {
 	defer close(req.doneCh)
-	req.msgID, req.err = pc.requestGetLastMessageID()
+	rsp, err := pc.requestGetLastMessageID()
+	if err != nil {
+		req.err = err
+		return
+	}
+	req.msgID = convertToMessageID(rsp.GetLastMessageId())
+	req.markDeletePosition = convertToMessageID(rsp.GetConsumerMarkDeletePosition())
 }
 
-func (pc *partitionConsumer) requestGetLastMessageID() (*trackingMessageID, error) {
+func (pc *partitionConsumer) requestGetLastMessageID() (*pb.CommandGetLastMessageIdResponse, error) {
 	if state := pc.getConsumerState(); state == consumerClosed || state == consumerClosing {
 		pc.log.WithField("state", state).Error("Failed to getLastMessageID closing or closed consumer")
 		return nil, errors.New("failed to getLastMessageID closing or closed consumer")
@@ -667,8 +687,7 @@ func (pc *partitionConsumer) requestGetLastMessageID() (*trackingMessageID, erro
 		pc.log.WithError(err).Error("Failed to get last message id")
 		return nil, err
 	}
-	id := res.Response.GetLastMessageIdResponse.GetLastMessageId()
-	return convertToMessageID(id), nil
+	return res.Response.GetLastMessageIdResponse, nil
 }
 
 func (pc *partitionConsumer) sendIndividualAck(msgID MessageID) *ackRequest {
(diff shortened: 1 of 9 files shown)


pulsar/reader_test.go
```diff
@@ -1070,3 +1070,159 @@ func TestReaderNextReturnsOnClosedConsumer(t *testing.T) {
...
+	assert.Nil(t, err)
+	defer producer.Close()
+
+	// send 100 messages
+	var lastMsgID MessageID
+	for i := 0; i < 10; i++ {
+		lastMsgID, err = producer.Send(ctx, &ProducerMessage{
+			Payload: []byte(fmt.Sprintf("hello-%d", i)),
+		})
+		fmt.Println(lastMsgID.String())
```
BewareMyPower, line 1099: Don't use `fmt.Println` in tests

pulsar/consumer_test.go
```diff
@@ -1262,8 +1262,9 @@ func TestConsumerSeek(t *testing.T) {
 	defer producer.Close()
 
 	consumer, err := client.Subscribe(ConsumerOptions{
-		Topic:            topicName,
-		SubscriptionName: "sub-1",
+		Topic:                   topicName,
+		SubscriptionName:        "sub-1",
+		StartMessageIDInclusive: true,
```
BewareMyPower, line 1267: This change seems irrelevant?

pulsar/consumer_zero_queue_test.go
```diff
@@ -474,6 +474,7 @@ func TestZeroQueueConsumer_Seek(t *testing.T) {
 		Topic:                   topicName,
 		EnableZeroQueueConsumer: true,
 		SubscriptionName:        "sub-1",
+		StartMessageIDInclusive: true,
```
BewareMyPower, line 477: irrelevant change again

pulsar/consumer_partition.go
```diff
@@ -997,7 +1016,15 @@ func (pc *partitionConsumer) requestSeek(msgID *messageID) error {
...
 		return err
 	}
-	pc.clearReceiverQueue()
+	// When the seek operation is successful, it indicates:
+	// 1. The broker has reset the cursor and sent a request to close the consumer on the client side.
+	//    Since this method is in the same goroutine as the reconnectToBroker,
+	//    we can safely clear the messages in the queue (at this point, it won't contain messages after the seek).
+	// 2. The startMessageID is reset to ensure accurate judgment when calling hasNext next time.
+	//    Since the messages in the queue are cleared here reconnection won't reset startMessageId.
+	pc.lastDequeuedMsg = nil
```
BewareMyPower, line 1025: I have a concern that if there is a race here. Before this change, `lastDequeuedMsg` is only updated by `setLastDequeuedMsg`, which is only called by the `Next` method called in user's goroutine. However, after that, `lastDequeuedMsg` could be reset with nil in the event loop goroutine (`runEventsLoop -> internalSeek -> requestSeek`)

BewareMyPower (changes requested):

pulsar/consumer_partition.go
```diff
@@ -2195,6 +2226,24 @@ func (pc *partitionConsumer) discardCorruptedMessage(msgID *pb.MessageIdData,
 }
 
 func (pc *partitionConsumer) hasNext() bool {
+
+	// If a seek by time has been performed, then the `startMessageId` becomes irrelevant.
+	// We need to compare `markDeletePosition` and `lastMessageId`,
+	// and then reset `startMessageID` to `markDeletePosition`.
+	if pc.hasSoughtByTime.CompareAndSwap(true, false) {
```
BewareMyPower, line 2233: This CAS operation seems confusing.

1. `hasSoughtByTime` is set with true in `internalSeekByTime`, so concurrent seek could happen. In Java client, it's prevented by `seekStatus.compareAndSet(SeekStatus.NOT_STARTED, SeekStatus.IN_PROGRESS)`.
2. Here it's changed from true to false, which means next time `hasNext` is called, it will skip the code block here.

pulsar/consumer_partition.go
shibd: When executing seek, will clear `messageChan` first and then `pauseDispatchMessage` message until the seek operator is complete(reconnect success)

Refer to this PR:

https://github.com/apache/pulsar-client-go/pull/1265

pulsar/consumer_partition.go
shibd: > Here it's changed from true to false, which means next time hasNext is called, it will skip the code block here.

Yes, After executing `seekByTime`, we only need to enter this method once, as it updates startMessageID to markDeletePosition. Now, it clearly knows what the messageId was from the last execution of seekByTime

shibd: However, we need to add a new condition: when `&& lastDequeuedMsg == nil`. This is because the user might have received the message but hasn't called this method.

pulsar/consumer_zero_queue_test.go
shibd: This is necessary, Because this test seeks to N−5 and expects to receive this message next time. 

In fact, StartMessageIDInclusive should also apply to the seek operation.

pulsar/consumer_test.go
shibd: This is necessary, Because this test seeks to N−5 and expects to receive this message next time.

In fact, StartMessageIDInclusive should also apply to the seek operation.

pulsar/consumer_partition.go
```diff
@@ -2195,6 +2226,24 @@ func (pc *partitionConsumer) discardCorruptedMessage(msgID *pb.MessageIdData,
...
 
 func (pc *partitionConsumer) hasNext() bool {
+
+	// If a seek by time has been performed, then the `startMessageId` becomes irrelevant.
+	// We need to compare `markDeletePosition` and `lastMessageId`,
+	// and then reset `startMessageID` to `markDeletePosition`.
+	if pc.lastDequeuedMsg == nil && pc.hasSoughtByTime.CompareAndSwap(true, false) {
+		res, err := pc.getLastMessageIDAndMarkDeletePosition()
+		if err != nil {
+			pc.log.WithError(err).Error("Failed to get last message id")
```
BewareMyPower, line 2236: In this case, `hasSoughtByTime` will still be `false`. Should we restore it with true here?

pulsar/consumer_test.go
BewareMyPower: Anyway, this is a breaking change. Though the existing behavior might not be expected, users might already depend on this behavior.

Unlike `hasNext()`, the existing behavior is that if a consumer seeks to the message id, it will still consume that message. While this PR's main purpose is to resolve the wrong result of `hasNext()`. I think you should use a different PR for changing the `seek` behavior.

BewareMyPower (changes requested):
If any user depends on the existing `seek` behavior (inclusive by default), he/she might treat it as a "message lost" issue.

This behavior change is beyond the scope of this PR. I also think it needs a discussion at the mail list.  We also need to document the change and tell existing users to set `startMessageInclusive` with true to keep the old behavior. But it's still a breaking change because if `startMessageInclusive` is true, the behavior of subscribing also changes.

Therefore, my opinion is to keep the existing seek behavior not changed for now. The "fix" could be delayed to next release.

pulsar/consumer_test.go
BewareMyPower: Unlike `hasNext()`, if it returns true and there is no more message, the application will be stuck at `Next`.

However, the existing behavior of `seek` won't cause such serious result. Indeed, some users might think it's a special behavior of Pulsar Go client and set the message id with the inclusive semantics.

shibd: I do not agree that this is a break change; it is more like a behavior fix or bug fix.

After this PR, it will keep the same behavior with the java client. Refer to(peek to id-5 and receive msg from id-6):
https://github.com/apache/pulsar/blob/a1f3322ed358ab6841f0d3e43f2afcc54788b887/pulsar-broker/src/test/java/org/apache/pulsar/broker/service/SubscriptionSeekTest.java#L142-L165

shibd:
> This behavior change is beyond the scope of this PR.

I changed the title. I think my intention was to clarify the issue with `hasNext` or `receive` after `seek`.

I believe email discussions are pointless because this is a behavior fix. If we need to inform users, we should clearly explain it in the release notes for the next release.

shibd:
Start discuss on thread: https://lists.apache.org/thread/knxsq637pyhltthn77859mpxb5wcxglj

geniusjoe:
> If any user depends on the existing `seek` behavior (inclusive by default), he/she might treat it as a "message lost" issue.
> 
> This behavior change is beyond the scope of this PR. I also think it needs a discussion at the mail list. We also need to document the change and tell existing users to set `startMessageInclusive` with true to keep the old behavior. But it's still a breaking change because if `startMessageInclusive` is true, the behavior of subscribing also changes.
> 
> Therefore, my opinion is to keep the existing seek behavior not changed for now. The "fix" could be delayed to next release.

I'm not very sure what do you mean "message lost"? Compared to current fixes, does previous behavior continue to consume messages regardless of `seek()` happened, or just continue to consume messages in client cache and then jump to consume the position which `seek()` defined?

pulsar/consumer_partition.go
```diff
@@ -2256,12 +2306,14 @@ func convertToMessageID(id *pb.MessageIdData) *trackingMessageID {
...
+			ledgerID:  int64(id.GetLedgerId()),
+			entryID:   int64(id.GetEntryId()),
+			batchIdx:  id.GetBatchIndex(),
+			batchSize: id.GetBatchSize(),
 		},
 	}
-	if id.BatchIndex != nil {
-		msgID.batchIdx = *id.BatchIndex
+	if msgID.batchIdx == -1 {
+		msgID.batchIdx = 0
```
nodece, line 2316: Why reset this value to 0? -1 is not the batch message, 0 is batch.

BewareMyPower: Sounds like a concern. Could you provide a test case to show this change might not work?

shibd: In the Go Client, it's not like this. You can run the following code to test:

Need change msg id String method first.
``` go
func (id *messageID) String() string {
	return fmt.Sprintf("%d:%d:%d:%d", id.ledgerID, id.entryID, id.batchIdx, id.partitionIdx)
}
```


``` go
func TestMessageID_BatchIdx(t *testing.T) {
		URL: lookupURL,
	})

	assert.Nil(t, err)
	defer client.Close()

	topic := newTopicName()
	ctx := context.Background()

	// create producer
		Topic:           topic,
		DisableBatching: true,
	})
	assert.Nil(t, err)
	defer producer.Close()

	// create consumer
		Topic:             topic,
		SubscriptionName:  "my-sub",
		Type:              Exclusive,
		ReceiverQueueSize: 10,
	})
	assert.Nil(t, err)
	defer consumer.Close()

	// send 10 messages with no batch
	var lastMsgID MessageID
	for i := 0; i < 10; i++ {
		lastMsgID, err = producer.Send(ctx, &ProducerMessage{
		})
		t.Logf("Sended message msgId: %#v \n", lastMsgID.String())
		assert.NoError(t, err)
		assert.NotNil(t, lastMsgID)
	}

	// Receive 10 message with no batch
	for i := 0; i < 10; i++ {
		msg, err := consumer.Receive(ctx)
		assert.Nil(t, err)
		t.Logf("Received message msgId: %#v \n", msg.ID().String())
		consumer.Ack(msg)
	}
}

output:
   consumer_test.go:572: Sended message msgId: "3141:0:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:1:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:2:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:3:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:4:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:5:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:6:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:7:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:8:0:0" 
    consumer_test.go:572: Sended message msgId: "3141:9:0:0" 
    consumer_test.go:581: Received message msgId: "3141:0:0:0" 
    consumer_test.go:581: Received message msgId: "3141:1:0:0" 
    consumer_test.go:581: Received message msgId: "3141:2:0:0" 
    consumer_test.go:581: Received message msgId: "3141:3:0:0" 
    consumer_test.go:581: Received message msgId: "3141:4:0:0" 
    consumer_test.go:581: Received message msgId: "3141:5:0:0" 
    consumer_test.go:581: Received message msgId: "3141:6:0:0" 
    consumer_test.go:581: Received message msgId: "3141:7:0:0" 
    consumer_test.go:581: Received message msgId: "3141:8:0:0" 
    consumer_test.go:581: Received message msgId: "3141:9:0:0" 
```

In Java, it will be a different implementation class, so there's no batchIndex field when disabled batch.

If we want to keep it consistent, we can change it later.

nodece: I see, because the batchIndex type is `int32`, not point.

I think you should remove this change. When the batch index is -1, do not change the value to 0.

nodece (changes requested):
Please address the batch index issue.

pulsar/consumer_partition.go
shibd: If remove this, when seek to a messageId(no batch), let say`(3141:1:0:0)`, When we compare, there will be unexpected behavior because the `lastMessageId` we get from the broker is: `3141:1:-1:0`.

What are your main concerns? Currently, this method is only used for comparison in `hasNext`.

I think we need to maintain consistency because now the non-batch message ID returned by the producer and consumer is `batchIndex=0`.

nodece (approved):

pulsar/consumer_partition.go
nodece: The PR didn't change the original behavior.

BTW, I think this is a worthy point of improvement.

PR #1345, opened 2025-03-12
Fix: SendAsync callback was not invoked when producer is in reconnecting

### LINKED ISSUE #1332: [Bug][Producer] The callback was not invoked during reconnecting.
#### Expected behavior

Callback should be called with a timeout error.

#### Actual behavior

Callback was not called

#### Steps to reproduce

1. Connect to the server;
2. Call SendAsync() to send a message, it will send successfully;
3. Stop the server;
4. Call SendAsync() to send some messages, the messages' callback won't be called anymore

#### System configuration
**Pulsar version**: x.y

nodece:
Fixes #1332

### Modifications

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -110,7 +110,7 @@ type partitionProducer struct {
 	compressionProvider      compression.Provider
 
 	// Channel where app is posting messages to be published
-	dataChan             chan *sendRequest
+	dataChan             chan *pendingItem
 	cmdChan              chan interface{}
 	connectClosedCh      chan *connectionClosed
 	publishSemaphore     internal.Semaphore
@@ -177,16 +177,15 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	p := &partitionProducer{
-		client:           client,
-		topic:            topic,
-		log:              logger,
-		cnxKeySuffix:     client.cnxPool.GenerateRoundRobinIndex(),
-		options:          options,
-		producerID:       client.rpcClient.NewProducerID(),
-		dataChan:         make(chan *sendRequest, maxPendingMessages),
-		cmdChan:          make(chan interface{}, 10),
-		connectClosedCh:  make(chan *connectionClosed, 1),
-		batchFlushTicker: time.NewTicker(batchingMaxPublishDelay),
+		client:          client,
+		topic:           topic,
+		log:             logger,
+		cnxKeySuffix:    client.cnxPool.GenerateRoundRobinIndex(),
+		options:         options,
+		producerID:      client.rpcClient.NewProducerID(),
+		dataChan:        make(chan *pendingItem, maxPendingMessages),
+		cmdChan:         make(chan interface{}, 10),
+		connectClosedCh: make(chan *connectionClosed, 1),
 		compressionProvider: internal.GetCompressionProvider(pb.CompressionType(options.CompressionType),
 			compression.Level(options.CompressionLevel)),
 		publishSemaphore:  internal.NewSemaphore(int32(maxPendingMessages)),
@@ -199,9 +198,6 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		ctx:               ctx,
 		cancelFunc:        cancelFunc,
 		backOffPolicyFunc: boFunc,
-	}
-	if p.options.DisableBatching {
-		p.batchFlushTicker.Stop()
 	}
 	p.setProducerState(producerInit)
 
@@ -219,7 +215,6 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 	}
 	err := p.grabCnx("")
 	if err != nil {
-		p.batchFlushTicker.Stop()
 		logger.WithError(err).Error("Failed to create producer at newPartitionProducer")
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


nodece:
@gunli Could you have a chance to review?

RobertIndie (changes requested):
This PR introduces two separate message writing paths:
1. Non-batch message: SendAsync -> dataChan
2. Batch messages: SendAsync -> batchChan -> dataChan

Both paths are asynchronous, which breaks the message order guaranteed by SendAsync.

For example, if we send two messages using SendAsync like this:

```golang
    Payload: []byte("A"),
}, func(_ MessageID, _ *ProducerMessage, err error) {
    errChan <- err
})

    Payload: []byte("B"),
}, func(_ MessageID, _ *ProducerMessage, err error) {
    errChan <- err
})
```

We need to ensure the message order is A, B.

pulsar/producer_test.go
```diff
@@ -2605,3 +2605,82 @@ func TestSelectConnectionForSameProducer(t *testing.T) {
...
+	testProducer.SendAsync(context.Background(), &ProducerMessage{
+		Payload: []byte("test"),
+	}, func(_ MessageID, _ *ProducerMessage, err error) {
+		errChan <- err
+	})
+	select {
+	case <-time.After(10 * time.Second):
+		t.Fatal("test timeout")
+	case <-errChan:
+		// fine
```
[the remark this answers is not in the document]
nodece: Fixed.

nodece:
ping @RobertIndie

pulsar/producer_test.go
```diff
@@ -2605,3 +2605,82 @@ func TestSelectConnectionForSameProducer(t *testing.T) {
...
+	require.NoError(t, err)
+	defer client.Close()
+
+	var testProducer Producer
+	require.Eventually(t, func() bool {
+		testProducer, err = client.CreateProducer(ProducerOptions{
+			Topic:           newTopicName(),
+			Schema:          NewBytesSchema(nil),
+			SendTimeout:     3 * time.Second,
+			DisableBatching: isEnabledBatching,
```
RobertIndie, line 2647:
```suggestion
			DisableBatching: isDisableBatching,
```

pulsar/producer_partition.go
```diff
@@ -898,16 +911,17 @@ func (p *partitionProducer) writeData(buffer internal.Buffer, sequenceID uint64,
...
 			createdAt:    now,
 			sentAt:       now,
 			buffer:       buffer,
 			sequenceID:   sequenceID,
 			sendRequests: callbacks,
-		})
-		p._getConn().WriteData(ctx, buffer)
+		}
+		p.pendingQueue.Put(&item)
+		p.writeChan <- &item
```
RobertIndie, line 924: The `writeChan` is a zero buffer channel. So if the client is reconnecting, the goroutine in line 564 will be stuck here. It couldn't timeout the message in the `dataChan`.

pulsar/producer_test.go
```diff
@@ -2605,3 +2605,82 @@ func TestSelectConnectionForSameProducer(t *testing.T) {
...
+	timeout := 10 * time.Second
+	err = c.Stop(context.Background(), &timeout)
+	require.NoError(t, err)
+
+	// send again
+	testProducer.SendAsync(context.Background(), &ProducerMessage{
+		Payload: []byte("test"),
+	}, func(_ MessageID, _ *ProducerMessage, err error) {
+		errChan <- err
+	})
```
RobertIndie, line 2678:
```suggestion
	for i := 0; i < 10; i++ {
		testProducer.SendAsync(context.Background(), &ProducerMessage{
			Payload: []byte("test"),
		}, func(id MessageID, producerMessage *ProducerMessage, err error) {
			fmt.Println("send async callback", id, producerMessage, err)
		})
	}
	// send again
	testProducer.SendAsync(context.Background(), &ProducerMessage{
		Payload: []byte("test"),
	}, func(_ MessageID, _ *ProducerMessage, err error) {
		errChan <- err
	})
```

You can eaisly reproduce the issue by sending some messages before this step. And also set the `BatchingMaxMessages` of the producer to a smaller value like `5`.

RobertIndie (changes requested):

pulsar/producer_test.go
nodece: Done.

pulsar/producer_partition.go
```diff
@@ -184,6 +185,7 @@ func newPartitionProducer(client *client, topic string, options *ProducerOptions
 		options:          options,
 		producerID:       client.rpcClient.NewProducerID(),
 		dataChan:         make(chan *sendRequest, maxPendingMessages),
+		writeChan:        make(chan *pendingItem, maxPendingMessages),
```
RobertIndie, line 188: This doesn't address the root cause. If we send too many messages, the `writeChan` can still get stuck, and the issue persists. I can reproduce the problem by producing at least `maxPendingMessages` after the cluster is shut down.

RobertIndie (changes requested):

pulsar/producer_partition.go
nodece: @RobertIndie Once the internal queue reaches its configured capacity (`maxPendingMessages`), the producer should either block or reject incoming messages based on the `DisableBlockIfQueueFull` setting:

- If `DisableBlockIfQueueFull` is **`false`**, the producer should **block** until space becomes available.
- If `DisableBlockIfQueueFull` is **`true`**, the producer should **fail fast** by rejecting new messages.

Proper backpressure handling is essential to avoid unbounded memory usage and ensure the system remains responsive under load or during broker unavailability.

By default, `DisableBlockIfQueueFull` is set to `false`, which means the producer will block when the number of pending messages exceeds `maxPendingMessages`. If it is explicitly set to true, the producer will fail fast with a producer send queue is full error.

I think this topic is different from the current PR and might be better addressed separately. Let’s keep this PR focused on the specific issue it aims to resolve.

PR #1372, opened 2025-05-20
[Issue 1371][producer] Put buffer back in pool only when safe

### LINKED ISSUE #1371: Buffer data race
#### Expected behavior

Buffer in pool should not be shared among go routines.

#### Actual behavior

We have observed a data race with a flush operation demonstrating that a buffer has been returned by the pool while already used by a different one:

```
WARNING: DATA RACE
Read at 0x00c0294414a0 by goroutine 25517:
  runtime.racereadrange()
      <autogenerated>:1 +0x24
  internal/poll.ignoringEINTRIO()
      internal/poll/fd_unix.go:745 +0x563
  internal/poll.(*FD).Write()
      internal/poll/fd_unix.go:381 +0x31e
  net.(*netFD).Write()
      net/fd_posix.go:96 +0x4a
  net.(*conn).Write()
      net/net.go:206 +0xac
  github.com/apache/pulsar-client-go/pulsar/internal.(*connection).internalWriteData()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/internal/connection.go:501 +0x1f4
  github.com/apache/pulsar-client-go/pulsar/internal.(*connection).run()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/internal/connection.go:434 +0x644
  github.com/apache/pulsar-client-go/pulsar/internal.(*connection).start.func1()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/internal/connection.go:240 +0x9b

Previous write at 0x00c0294414a5 by goroutine 25984:
  encoding/binary.bigEndian.PutUint32()
      encoding/binary/binary.go:185 +0xe5
  github.com/apache/pulsar-client-go/pulsar/internal.(*buffer).WriteUint32()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/internal/buffer.go:192 +0x8a
  github.com/apache/pulsar-client-go/pulsar/internal.serializeMessage()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/internal/commands.go:293 +0x4bc
  github.com/apache/pulsar-client-go/pulsar/internal.(*batchContainer).Flush()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/internal/batch_builder.go:279 +0x65a
  github.com/apache/pulsar-client-go/pulsar.(*partitionProducer).internalFlushCurrentBatch()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/producer_partition.go:864 +0xb0
  github.com/apache/pulsar-client-go/pulsar.(*partitionProducer).runEventsLoop()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/producer_partition.go:586 +0x356
  github.com/apache/pulsar-client-go/pulsar.newPartitionProducer.gowrap2()
      github.com/apache/pulsar-client-go@v0.15.1-candidate-1/pulsar/producer_partition.go:239 +0x33
```

Furthermore, the flush call takes a buffer from the pool but never releases it. 

#### Steps to reproduce

Very hard to reproduce since it involves a data race. We are still trying to identify the piece of code causing this sharing of buffer.

UPDATE: it seems that the buffer management is risky because a buffer is put back into the pool while the write request could still be in the write loop if the connection is not closed.

#### System configuration
**Pulsar version**: v4.0.4
**Pulsar client**: v0.15.1-candidate-1

Gilthoniel:
Fixes #1371 

### Motivation

Some buffer ends up in two go routines, one reading to write to the connection and another one that writes a new message. This happens because the buffer is released without any guarantee that the connection is done with it.

### Modifications

This adds a check to reuse the buffer only when the sending request is successful which means that a receipt has been received and thus that the buffer has been already read. Any error when finalizing the pending item is a risk of the buffer being not completely done.

### Verifying this change

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -1743,14 +1743,21 @@ func (i *pendingItem) done(err error) {
 	}
 
 	i.isDone = true
-	// return the buffer to the pool after all callbacks have been called.
-	defer buffersPool.Put(i.buffer)
+
 	if i.flushCallback != nil {
 		i.flushCallback(err)
 	}
 
 	if i.cancel != nil {
 		i.cancel()
+	}
+
+	if err == nil {
+		// Buffer is returned after executing all the callbacks and only if the
+		// pending item is successful because that is the only state that ensure
+		// a finality of the buffer at that time (an erroneuous item might still
+		// be in the connection sending queue).
+		buffersPool.Put(i.buffer)
 	}
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,8 @@
 *.out
 coverage.html
 
-perf/perf
-pulsar-perf
-bin
-
+perf/perf/
+pulsar-perf/
+bin/
+logs/
 vendor/


pulsar/producer_partition.go
```diff
@@ -1743,15 +1743,22 @@ func (i *pendingItem) done(err error) {
...
 	if i.cancel != nil {
 		i.cancel()
 	}
+
+	if err == nil {
+		// Buffer is returned after executing all the callbacks and only if the
+		// pending item is successful because that is the only state that ensure
+		// a finality of the buffer at that time (an erroneuous item might still
+		// be in the connection sending queue).
+		buffersPool.Put(i.buffer)
```
RobertIndie: Why is the buffer only returned when there is no error? It seems the buffer will not be returned if this pending item fails.

Gilthoniel: Yes that's correct because with current logic, it's impossible to be certain 100% that the buffer is not in a connection queue which is causing data race. So the idea is that in the rare cases when the item fails, the buffer is thrown away instead of putting it back in the pool and causing data races.

Gilthoniel: We observe that bug several times a day in our CI.

Gilthoniel: I didn't see your comment in the issue, let me have a look at it.

Gilthoniel: @RobertIndie I've tried a different approach that should go in the direction you were saying. Let me know if that makes sense.

pulsar/producer_partition.go
RobertIndie, line 1: At line 397, the pi.buffer is not borrowed.

Gilthoniel: That is the same pending item so it's already borrowed in that case.

pulsar/internal/connection.go
RobertIndie, line 1: Just found another buffer leak issue. We should cleanup the `writeRequestsCh` when closing the connection

RobertIndie: Oh. Just see that you push the new codes. Let me review it again.

Gilthoniel: Added the cleanup anyway and also fixed an issue with the backoff with the latest merge on master.

RobertIndie:
> @RobertIndie while implementing your proposal, I discovered another race because we can have the situation where the buffer is sent to multiple connections and read in concurrence. There is then a risk of partial reading even if the buffer is released properly. I've tried something different that should solve it once and for all.

I think this should be a separate issue. The approach in this PR causes a deep copy of data between the producer partition and the connection. I believe this will affect performance. It's unnecessary to sacrifice the perf to fix a very rare problem.

To solve this, I think we should clear all `writeRequestsCh` so the old connection does not hold any buffer before the producer reconnects and gets a new connection.

pulsar/internal/buffer.go
```diff
@@ -217,3 +225,62 @@ func (b *buffer) Clear() {
...
+	return nil
+}
+
+func (p synchronizedBufferPool) Put(buffer *SharedBuffer) {
+	if !buffer.isCurrentlyShared.CompareAndSwap(true, false) {
+		// When swapping fails, it means that a previous go routine has already
+		// tried to recycle the buffer, indicating that the sharing is not there
+		// anymore and the buffer can safely be reused.
+		p.pool.Put(buffer)
+	}
```
RobertIndie, line 262: If:
- Buffer A created from pool
- Call A.Retain, create a new buffer B (isCurrentlyShared=true)
- Put back the A(isCurrentlyShared=false), the buffer A will be returned to the pool. And the buffer get recycled.
- Panic! The buffer B still holds this buffer.

Why not just use the ref count here?

pulsar/internal/connection.go
```diff
@@ -456,10 +474,19 @@ func (c *connection) runPingCheck(pingCheckTicker *time.Ticker) {
 	}
 }
 
-func (c *connection) WriteData(ctx context.Context, data Buffer) {
+func (c *connection) WriteData(ctx context.Context, data *SharedBuffer) {
+	written := false
+	defer func() {
+		if !written {
+			// Buffer has failed to be written and can now be reused.
+			c.bufferPool.Put(data)
```
RobertIndie, line 482: Can we add a new method like `Recycle` into the `SharedBuffer` so we don't need to maintain the bufferPool in the `connection`. The SharedBuffer can hold the ref to the buffer pool.

PR #1380, opened 2025-06-06
fix: support json token file format authentication

mattisonchao:
### Motivation

Support JSON token file format authentication for Pulsar Admin.
E.g: 
```
{"file": "./mnt/token"}
```

### Modifications

- Add token file format for json.

### Verifying this change

### DIFF
--- a/pulsaradmin/pkg/admin/auth/token.go
+++ b/pulsaradmin/pkg/admin/auth/token.go
@@ -36,6 +36,7 @@ const (
 
 type Token struct {
 	Token string `json:"token"`
+	File string `json:"file"`
 }
 
 type TokenAuthProvider struct {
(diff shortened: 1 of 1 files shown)


BewareMyPower (commented):
Is it necessary? This could be done from the application side.

It looks unnecessarily complicated because now `File` has two supported format `<token>` or `{"file": "<token>"}`.

The existing `NewAuthenticationTokenFromFile` has poor design that should be deprecated.

BewareMyPower:
It makes sense to use the same way to configure JWT authentication for admin and client.

But from the perspective of designing the API from the very beginning:
1. One format is enough. Pure token string is enough, JSON is scalable but it's over-designed. At least the server side only processes the token string.
2. Reading from file (or any other external data source like DB or even a topic) is unnecessary. It should be done from the application side.

PR #1382, opened 2025-06-09
Skip very flaky TestMessageSingleRouter

BewareMyPower:
It should be a server side issue that we might need 4.0.6 to fix the flakiness.


BewareMyPower:
I pushed a upstream fix: https://github.com/apache/pulsar/pull/24401

crossoverJie (approved):

nodece (approved):

PR #1389, opened 2025-07-09
fix: move pulsarttracing from the internal to parent directory

xuthus5:
### What’s Changed

- **Tracing implementation fully migrated from opentracing to OpenTelemetry (otel):**
  - All tracing logic now uses OpenTelemetry APIs for span creation, propagation, and context injection/extraction.
  - Producer and consumer interceptors, context propagation utilities, and span enrichment are all otel-based.
  - All tests and documentation have been updated to reflect the new OpenTelemetry usage.

- **Tracing package moved out of internal:**
  - The `tracing` package was previously located in the internal directory and not accessible to external users.
  - It has now been moved to a top-level package (`pulsar/tracing`), making it publicly available.
  - This change is intended to allow external users and integrators to leverage Pulsar Go Client’s tracing capabilities directly in their own applications.

- **Documentation updated:**
  - The README has been rewritten in English to demonstrate the latest usage patterns with OpenTelemetry.
  - Usage examples now show how to set up OpenTelemetry, use interceptors, propagate context, and create child spans from messages.

### Motivation

Moving the tracing package out of internal and fully adopting OpenTelemetry makes distributed tracing more accessible and standard for all users of the Pulsar Go Client. External consumers can now easily integrate and extend tracing in their own systems, following modern observability best practices.

**Note:** The tracing package is now fully based on OpenTelemetry and is no longer compatible with opentracing.


xuthus5:
@RobertIndie PTAL.

PR #1394, opened 2025-07-17
Fix sending buffer race by using proper reference counting

RobertIndie:
Fixes https://github.com/apache/pulsar-client-go/issues/1371

### Motivation

The issue is that the buffer is sent to the connection when the data is flushed at this line: https://github.com/apache/pulsar-client-go/blob/0ab28c229e4dab2adb505a135325b6ede6e0e4f4/pulsar/producer_partition.go#L910

Because of this, both the data request in the connection and the pending item hold onto this buffer.

However, if the pending item fails due to a timeout or the producer closing, the buffer is returned to the buffer pool at this line: https://github.com/apache/pulsar-client-go/blob/0ab28c229e4dab2adb505a135325b6ede6e0e4f4/pulsar/producer_partition.go#L1747

Yet, the connection still has this buffer. When the connection tries to handle the data request, it will try to use a buffer that has already been returned. This causes a data race.

This fix implements proper reference counting to ensure buffers are only returned to the pool when all references are released, preventing the data race.

### Modifications

  - Added `Retain()` and `Release()` methods to the `Buffer` interface
  - Implemented atomic reference counting using `atomic.Int64`
  - Buffers are only returned to the pool when reference count reaches zero
  - Add new metric `pulsar_client_sending_buffers_count` to show the sending buffers count.
  - This PR also fixes serval minor issues like data/request cleanup when closing the connection 

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/internal/metrics.go
+++ b/pulsar/internal/metrics.go
@@ -26,7 +26,7 @@ var (
 		"client": "go",
 	}
 
-	SendingBuffersCount = prometheus.NewGauge(prometheus.GaugeOpts{
+	sendingBuffersCount = prometheus.NewGauge(prometheus.GaugeOpts{
 		Name:        "pulsar_client_sending_buffers_count",
 		Help:        "Number of sending buffers",
 		ConstLabels: defaultConstLabels,
@@ -548,7 +548,7 @@ func NewMetricsProvider(metricsCardinality int, userDefinedLabels map[string]str
 			metrics.RPCRequestCount = are.ExistingCollector.(prometheus.Counter)
 		}
 	}
-	_ = registerer.Register(SendingBuffersCount)
+	_ = registerer.Register(sendingBuffersCount)
 	return metrics
 }
 
--- a/pulsar/internal/metrics.go
+++ b/pulsar/internal/metrics.go
@@ -21,6 +21,18 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+var (
+	defaultConstLabels = map[string]string{
+		"client": "go",
+	}
+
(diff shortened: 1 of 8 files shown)


pulsar/producer_partition.go
```diff
@@ -1744,7 +1737,9 @@ func (i *pendingItem) done(err error) {
 
 	i.isDone = true
 	// return the buffer to the pool after all callbacks have been called.
-	defer buffersPool.Put(i.buffer)
+	if i.buffer != nil {
```
BewareMyPower: When is `i.buffer` nil?

pulsar/internal/metrics.go
```diff
@@ -21,6 +21,18 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+var (
+	defaultConstLabels = map[string]string{
+		"client": "go",
+	}
+
+	SendingBuffersCount = prometheus.NewGauge(prometheus.GaugeOpts{
```
crossoverJie:
```suggestion
	sendingBuffersCount = prometheus.NewGauge(prometheus.GaugeOpts{
```
It is recommended to change to a non-exported variable.

pulsar/producer_partition.go
RobertIndie: It shouldn't be possible. I removed this check.

pulsar/internal/metrics.go
RobertIndie: Yes. Thanks.

crossoverJie (approved):
LGTM

BewareMyPower (approved):

PR #1408, opened 2025-08-19
Change the pulsar_client_sending_buffers_count metric to client level

BewareMyPower:
#1394 introduces the `pulsar_client_sending_buffers_count` metric to track how many buffers are allocated for send purpose and not put back to the pool. However, unlike other metrics, this metric is not client level, so it cannot be attached with `CustomMetricsLabels` in client options.

When a send buffer was not put back to the pool, it means the `Release` method is not called due to some reason. Changing this metric to client level could help locate which client has triggered this bug in an application that has many client instances from different businesses.

Here is an example metric when I configured `CustomMetricsLabels: map[string]string{"key": "value"}` after this change

```
pulsar_client_sending_buffers_count{client="go",key="value"} 1
```

### DIFF
--- a/pulsar/producer_partition.go
+++ b/pulsar/producer_partition.go
@@ -798,7 +798,10 @@ func (p *partitionProducer) internalSingleSend(
 	payloadBuf := internal.NewBuffer(len(compressedPayload))
 	payloadBuf.Write(compressedPayload)
 
-	buffer := buffersPool.GetBuffer(int(payloadBuf.ReadableBytes()*3/2), p.client.metrics.SendingBuffersCount)
+	buffer := buffersPool.GetBuffer(int(payloadBuf.ReadableBytes() * 3 / 2))
+	bufferCount := p.client.metrics.SendingBuffersCount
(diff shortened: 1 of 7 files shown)


pulsar/producer_partition.go
```diff
@@ -126,6 +124,8 @@ type partitionProducer struct {
 	ctx                  context.Context
 	cancelFunc           context.CancelFunc
 	backOffPolicyFunc    func() backoff.Policy
+
+	buffersPool internal.BuffersPool
```
RobertIndie: This should be at the process level. It doesn't make sense to maintain the pool for each pulsar client.

I think maybe we don't need to add the custom labels for this process level metric.

crossoverJie:
Can you explain the motivation for this change?

BewareMyPower:
@crossoverJie I have updated the PR description, PTAL again

RobertIndie (approved):

crossoverJie (approved):
LGTM

PR #1409, opened 2025-08-25
[issue #1357][producer] fix: allow multiples callbacks with concurrent producer flushes (async publish)

### LINKED ISSUE #1357: Calling producer.Flush concurrently would hang indefinitely
#### Expected behavior

Calling `producer.Flush `concurrently should not hang forever

#### Actual behavior

Calling `producer.Flush` with multiple go-routines simutaneously, only one would succeed, others would hang forever.

#### Steps to reproduce

Assuming having a function that would send async messages and flush immediately.
Disclamer:
1. I know this is not a common usage but just to illustrate the situation
2. Normally we would add context with timeout for function calling, again, just to illustrate the situation, I've used the `context.Background()`

```
func test3(disableBatching bool) {
	fmt.Println("running test 3 with disableBatching", disableBatching)
	client, err := pulsarsdk.NewClient(pulsarsdk.ClientOptions{
		URL:            pulsarEndpoint,
		Authentication: pulsarsdk.NewAuthenticationToken(pulsarToken),
	})
	if err != nil {
		panic(err)
	}
	defer client.Close()
	producer, err := client.CreateProducer(pulsarsdk.ProducerOptions{
		Topic:           pulsarTopic,
		Name:            uuid.NewString(),
		DisableBatching: disableBatching,
	})
	if err != nil {
		panic(err)
	}
	defer producer.Close()

	ctx, cancel := registerExit()
	defer cancel()

	const n = 10
	var (
		wg sync.WaitGroup
	)
	wg.Add(n)
	for i := 0; i < n; i++ {
		go func(i int) {
			producer.SendAsync(ctx, &pulsarsdk.ProducerMessage{
				Payload: []byte(uuid.NewString()),
			}, func(id pulsarsdk.MessageID, message *pulsarsdk.ProducerMessage, err error) {
				wg.Done()
				if err != nil {
					panic(err)
				}
			})

			if err := producer.Flush(); err != nil {
				panic(err)
			}
		}(i)
	}

	wg.Wait()

```

With additional logs, you can see that one call of `producer.Flush` succeeds, while others hang indefinitely.

Check the `flush` related codes, first we can observe that the expected `case <-flushReq.doneCh` is never triggered.

```
func (p *partitionProducer) FlushWithCtx(ctx context.Context) error {
		return ErrProducerClosed
	}

	flushReq := &flushRequest{
		doneCh: make(chan struct{}),
		err:    nil,
	}
	select {
	case <-ctx.Done():
		return ctx.Err()
	case p.cmdChan <- flushReq:
	}

	// wait for the flush request to complete
	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-flushReq.doneCh:
		return flushReq.err
	}
}
```

The root cause might be here:

```
	p.clearPendingSendRequests()

	if !p.options.DisableBatching {
		p.internalFlushCurrentBatch()
	}

	pi, ok := p.pendingQueue.PeekLast().(*pendingItem)
	if !ok {
		close(fr.doneCh)
		return
	}

	// lock the pending request while adding requests
	// since the ReceivedSendReceipt func iterates over this list
	pi.Lock()
	defer pi.Unlock()

	if pi.isDone {
		// The last item in the queue has been completed while we were
		// looking at it. It's safe at this point to assume that every
		// message enqueued before Flush() was called are now persisted
		close(fr.doneCh)
		return
	}

	pi.flushCallback = func(err error) {
		fr.err = err
		close(fr.doneCh)
	}
}

```

`pi, ok := p.pendingQueue.PeekLast().(*pendingItem)` returns the same pi for all concurrent flush calls, causing `pi.flushCallback` to be overwritten. As a result, only the most recent `fr.doneCh` will be closed, causing all other flush requests to hang indefinitely.

#### Possible fix

I've applied a monkey patch of `pendingItem` by altering:
```
type pendingItem struct {
	sync.Mutex
	buffer        internal.Buffer
	sequenceID    uint64
	createdAt     time.Time
	sentAt        time.Time
	sendRequests  []interface{}
	isDone        bool
	flushCallback func(err error)
}
```

to:
```
type pendingItem struct {
	sync.Mutex
	buffer        internal.Buffer
	sequenceID    uint64
	createdAt     time.Time
	sentAt        time.Time
	sendRequests  []interface{}
	isDone        bool
	flushCallbacks []func(err error)
}
```

And:
```
func (i *pendingItem) done(err error) {
	if i.isDone {
		return
	}
	i.isDone = true
	buffersPool.Put(i.buffer)
	if i.flushCallback != nil {
		i.flushCallback(err)
	}
}

```

to
```
func (i *pendingItem) done(err error) {
	if i.isDone {
		return
	}
	i.isDone = true
	buffersPool.Put(i.buffer)
	for _, cb := range i.flushCallback {
		cb(err)
	}
}

```

Same applies to `internalFlush`:
```
	pi.flushCallbacks = append(pi.flushCallbacks, func(err error) {
		fr.err = err
		close(fr.doneCh)
	})

```

And the `test3` would pass.

#### System configuration
**Pulsar version**:  3.2.4
**Pulsar-client-go version**: 0.14.0

thomas-bousquet:
Fixes #1357

### Motivation

When flushing concurrently with the same producer, the producer potentially creates a deadlock when the callback is only called once but the client waits for the flush request to be done in all `FlushWithCtx()`.

### Modifications

Change the callback function for to a list of callbacks.

### Verifying this change

Testcase `TestConcurrentFlushInProducer` added in the producer test suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Check for Docker images
         id: check_images
-        run: echo "::set-output name=images::$(docker images -q | wc -l)"
+        run: echo "{name}={images::$(docker images -q | wc -l)}" >> $GITHUB_OUTPUT
       - name: Clean Docker cache if images exist
         if: ${{ steps.check_images.outputs.images > 0 }}
         run: docker rmi $(docker images -q) -f && df -h        
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Check for Docker images
         id: check_images
-        run: echo "{name}={images::$(docker images -q | wc -l)}" >> $GITHUB_OUTPUT
+        run: echo "images=$(docker images -q | wc -l)" >> $GITHUB_OUTPUT
       - name: Clean Docker cache if images exist
         if: ${{ steps.check_images.outputs.images > 0 }}
         run: docker rmi $(docker images -q) -f && df -h        
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,13 +50,7 @@ jobs:
       matrix:
         go-version: [ '1.23', '1.24' ]
     steps:
-      - uses: actions/checkout@v3
-      - name: Check for Docker images
-        id: check_images
-        run: echo "images=$(docker images -q | wc -l)" >> $GITHUB_OUTPUT
-      - name: Clean Docker cache if images exist
-        if: ${{ steps.check_images.outputs.images > 0 }}
-        run: docker rmi $(docker images -q) -f && df -h        
+      - uses: actions/checkout@v3      
       - uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go-version }}
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ pulsar-perf
 bin
 
 vendor/
+logs/
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -5230,7 +5230,7 @@ func sendMessages(t *testing.T, client Client, topic string, startIndex int, num
 			}
 		}
 	}
-	assert.Nil(t, producer.Flush())
+	assert.Nil(t, producer.FlushWithCtx(ctx))
 }
 
 func receiveMessages(t *testing.T, consumer Consumer, numMessages int) []Message {
@@ -5276,10 +5276,10 @@ func TestAckResponseNotBlocked(t *testing.T) {
 			}
 		})
 		if i%100 == 99 {
-			assert.Nil(t, producer.Flush())
+			assert.Nil(t, producer.FlushWithCtx(ctx))
 		}
 	}
-	producer.Flush()
+	producer.FlushWithCtx(ctx)
 	producer.Close()
 
 	// Set a small receiver queue size to trigger ack response blocking if the internal `queueCh`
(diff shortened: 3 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Check for Docker images
         id: check_images
-        run: echo "::set-output name=images::$(docker images -q | wc -l)"
+        run: echo "{name}={images::$(docker images -q | wc -l)}" >> $GITHUB_OUTPUT
```
lhotari: The extra curly braces don't look right here. I'd assume that this could simply be:
```suggestion
        run: echo "images=$(docker images -q | wc -l)" >> $GITHUB_OUTPUT
```

lhotari: Side note: This step in GitHub Actions CI is not very useful in the first place since I assume that there would always be some pre-existing images. In any case, the initial state in GitHub Actions CI is the same each time, so it doesn't make much sense to have condition logic like this.

thomas-bousquet: My bad, I took it from the templating I saw in the docs, thinking it was the actual pattern required. I'll remove them.

crossoverJie: > Side note: This step in GitHub Actions CI is not very useful in the first place since I assume that there would always be some pre-existing images. In any case, the initial state in GitHub Actions CI is the same each time, so it doesn't make much sense to have condition logic like this.

I also agree that the "`Clean Docker cache if images exist`" job has never been executed. maybe this job can be deleted.

thomas-bousquet: 👍 I removed the steps (image count + clean docker cache)

PR #1410, opened 2025-08-26
[improve] modify the negativeACK structure to reduce memory overhead

syslucas:
### Motivation
The main content of this PR is to implement the Pulsar Java community's [PIP-393](https://github.com/apache/pulsar/pull/23601) proposal in the Go client.

Negative Acknowledgement handling in now Pulsar Go client exhibits inefficiencies related to memory usage, timing precision, and batch-level contention.
This effort aims to bring parity with the refinements implemented in the Java client (and [CPP client](https://github.com/apache/pulsar-client-cpp/pull/497)), notably around memory usage and redelivery coordination.

### Modifications

Refactor the `negativeAcksTracker` to solve the above problems.

### Verifying this change

  - *Functional testing of negative acks is covered by existing tests (e.g., TestConsumerNack)*
  - *Extended integration test for evaluating redelivery precision across varying negative acknowledgement bit-widths (i.e., TestNegativeAckPrecisionBitCnt*

  - The public API: (no)
  - The schema: (no)
  - The wire protocol: (no)

### Documentation

### DIFF
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -1263,8 +1263,8 @@ func TestNegativeAckPrecisionBitCnt(t *testing.T) {
 		_, err = producer.Send(ctx, &ProducerMessage{Payload: []byte(content1)})
 		assert.Nil(t, err)
 
-		// Send second message around 1/2 into the window (still in same window)
-		time.Sleep(time.Duration(windowMs/2) * time.Millisecond)
+		// Send second message around 3/4 into the window (still in same window)
+		time.Sleep(time.Duration(windowMs*3/4) * time.Millisecond)
 		content2 := fmt.Sprintf("msg2-p%d", boundaryBits)
 		_, err = producer.Send(ctx, &ProducerMessage{Payload: []byte(content2)})
 		assert.Nil(t, err)
@@ -1291,10 +1291,10 @@ func TestNegativeAckPrecisionBitCnt(t *testing.T) {
 		assert.Nil(t, err)
 		redeliveryTime2 := time.Now()
 
-		// KEY TEST: Both messages should be redelivered simultaneously (within 1 ms)
-		// because precision alignment groups both messages in the same time bucket
-		// filter out precision bits < 5 because they are flaky, their time granularity is too small (i.e., 1 ms to 16 ms)
-		if boundaryBits >= 5 {
+		// For both the default precision (nil) and precisionBit=8, boundaryBits is 8.
+		// This checks that the default precisionBit is correctly set to 8,
+		// and that its redelivery behavior matches a consumer explicitly configured with precisionBit=8.
+		if boundaryBits == 8 {
 			assert.InDelta(t, redeliveryTime1.UnixMilli(), redeliveryTime2.UnixMilli(), 1)
 		}
 
--- a/pulsar/negative_acks_tracker.go
+++ b/pulsar/negative_acks_tracker.go
@@ -30,7 +30,7 @@ type redeliveryConsumer interface {
 	Redeliver(msgIDs []messageID)
 }
 
-type LedgerID = int64
+type ledgerID = int64
 
 type negativeAcksTracker struct {
 	sync.Mutex
@@ -104,11 +104,11 @@ func putNackEntry(t *negativeAcksTracker, batchMsgID *messageID, delay time.Dura
 	// try get trimmedTime
 	value, exists := t.negativeAcks.Get(trimmedTime)
 	if !exists {
-		newMap := make(map[LedgerID]*roaring64.Bitmap)
+		newMap := make(map[ledgerID]*roaring64.Bitmap)
 		t.negativeAcks.Put(trimmedTime, newMap)
 		value = newMap
 	}
-	bitmapMap, ok := value.(map[LedgerID]*roaring64.Bitmap)
+	bitmapMap, ok := value.(map[ledgerID]*roaring64.Bitmap)
 	if !ok {
 		panic("negativeAcksTracker: value is not of expected type map[LedgerID]*roaring64.Bitmap")
 	}
@@ -168,7 +168,7 @@ func (t *negativeAcksTracker) track() {
 						break
 					}
 
-					ledgerMap := iterator.Value().(map[LedgerID]*roaring64.Bitmap)
+					ledgerMap := iterator.Value().(map[ledgerID]*roaring64.Bitmap)
 					for ledgerID, entrySet := range ledgerMap {
 						for _, entryID := range entrySet.ToArray() {
 							msgID := messageID{


pulsar/negative_acks_tracker.go
```diff
@@ -103,14 +134,20 @@ func (t *negativeAcksTracker) AddMessage(msg Message) {
...
+	if !exists {
+		newMap := make(map[LedgerID]*roaring64.Bitmap)
+		t.negativeAcks.Put(trimmedTime, newMap)
+		value = newMap
+	}
+	bitmapMap := value.(map[LedgerID]*roaring64.Bitmap)
+	if _, exists := bitmapMap[batchMsgID.ledgerID]; !exists {
+		bitmapMap[batchMsgID.ledgerID] = roaring64.NewBitmap()
+	}
+	bitmapMap[batchMsgID.ledgerID].Add(uint64(batchMsgID.entryID))
```
RobertIndie: Please reduce the duplicate codes since it's getting more complex.

pulsar/consumer.go
```diff
@@ -183,6 +183,11 @@ type ConsumerOptions struct {
 	// processed. Default is 1 min. (See `Consumer.Nack()`)
 	NackRedeliveryDelay time.Duration
 
+	// NackPrecisionBit specifies the precision bit for nack redelivery delay.
+	// This is used to trim the lower bits of the nack redelivery delay to reduce memory usage.
+	// Default is 8 bits.
+	NackPrecisionBit int64
```
RobertIndie: How do you set the default value `8`?

pulsar/consumer_test.go
```diff
@@ -1216,13 +1217,86 @@ func TestConsumerNack(t *testing.T) {
...
+			assert.True(t, id%2 == 1) // Optional check, included for clarity
+			receivedOdd++
+		}
+
+		// Acknowledge message to mark it as processed
 		consumer.Ack(msg)
 	}
+
+	// Verify that the correct number of odd messages were received
+	assert.Equal(t, expectedOdd, receivedOdd)
```
RobertIndie: What's the reason for these changes? It seems after this change, the consumer would be expected to receive the even messages again which seems incorrect?
You can see the test would still be passed even though the consumer receive even messages which is not expected:
```
Receive redelivered message: msg-content-0
Receive redelivered message: msg-content-1
Receive redelivered message: msg-content-2
Receive redelivered message: msg-content-3
Receive redelivered message: msg-content-4
Receive redelivered message: msg-content-5
Receive redelivered message: msg-content-6
Receive redelivered message: msg-content-7
Receive redelivered message: msg-content-8
Receive redelivered message: msg-content-9
Receive redelivered message: msg-content-10
Receive redelivered message: msg-content-11
Receive redelivered message: msg-content-12
Receive redelivered message: msg-content-13
Receive redelivered message: msg-content-14
Receive redelivered message: msg-content-15
Receive redelivered message: msg-content-16
```

image

pulsar/consumer_test.go
```diff
@@ -1216,13 +1217,86 @@ func TestConsumerNack(t *testing.T) {
...
+		deviation := time.Duration(int64(1)<<precision) * time.Millisecond
+
+		// Wait for redelivery
+		redelivered, err := consumer.Receive(ctx)
+		assert.Nil(t, err)
+		assert.Equal(t, content, string(redelivered.Payload()))
+
+		now := time.Now()
+		// Assert that redelivery happens >= expected - deviation
+		assert.GreaterOrEqual(t, now.UnixMilli(), expectedRedelivery.UnixMilli()-deviation.Milliseconds())
```
RobertIndie: We should add the `LessOrEqual` assertion to ensure the redelivery time is not too long.

pulsar/consumer_test.go
```diff
@@ -1185,6 +1185,7 @@ func TestConsumerNack(t *testing.T) {
 		SubscriptionName:    "sub-1",
 		Type:                Shared,
 		NackRedeliveryDelay: 1 * time.Second,
+		NackPrecisionBit:    8,
```
RobertIndie: We should add a test to verfy the default value of `NackPrecisionBit`.

pulsar/negative_acks_tracker.go
syslucas: Added `putNackEntry` function to reduce code redundancy.

pulsar/consumer.go
syslucas: I added the default value of `NackPrecisionBit` with reference to `NackBackoffPolicy` and `ExpireTimeOfIncompletChunk`.
image

pulsar/consumer_test.go
syslucas: Because after adding `NackPrecisionBit`, the messages in the same time bucket are out of order. The original test strictly depends on the ordered redelivery of messages, so I modified it.

pulsar/consumer_impl.go
```diff
@@ -121,6 +121,10 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		options.NackBackoffPolicy = new(defaultNackBackoffPolicy)
 	}
 
+	if options.NackPrecisionBit <= 0 {
+		options.NackPrecisionBit = 8
+	}
```
RobertIndie: According to the java doc, the `0` should be a valid value:

> If the value is 0, the redelivery time will be accurate to ms.

If this config is a optional config, we could use a pointer type.

pulsar/consumer_test.go
RobertIndie: The consumer consumption should be strictly ordering. Otherwise, it would break many use cases.

syslucas: I misunderstood this test before and have reverted the relevant changes

pulsar/consumer_test.go
RobertIndie: This doesn't seem to be resolved.

pulsar/consumer_test.go
RobertIndie: This doesn't seem to be resolved.

pulsar/consumer_impl.go
```diff
@@ -121,6 +121,10 @@ func newConsumer(client *client, options ConsumerOptions) (Consumer, error) {
 		options.NackBackoffPolicy = new(defaultNackBackoffPolicy)
 	}
 
+	if options.NackPrecisionBit == nil || *options.NackPrecisionBit < 0 {
```
RobertIndie: We should return the error if it's set to be <0.

pulsar/negative_acks_tracker.go
```diff
@@ -95,15 +100,19 @@ func putNackEntry(t *negativeAcksTracker, batchMsgID *messageID, delay time.Dura
...
 		newMap := make(map[LedgerID]*roaring64.Bitmap)
 		t.negativeAcks.Put(trimmedTime, newMap)
 		value = newMap
 	}
-	bitmapMap := value.(map[LedgerID]*roaring64.Bitmap)
+	bitmapMap, ok := value.(map[LedgerID]*roaring64.Bitmap)
+	if !ok {
+		t.log.Errorf("negativeAcksTracker: value for time %v is not of expected type map[LedgerID]*roaring64.Bitmap", trimmedTime)
+		return
+	}
```
RobertIndie, line 114: This will hide the error. Why don't we panic it here?

pulsar/negative_acks_tracker.go
```diff
@@ -153,3 +203,5 @@ func (t *negativeAcksTracker) Close() {
 		t.doneCh <- nil
 	})
 }
+
+var defaultNackPrecisionBit = int64(8)
```
RobertIndie: It's better to make it as a const value.

pulsar/consumer_test.go
```diff
@@ -1225,6 +1225,201 @@ func TestConsumerNack(t *testing.T) {
...
+
+		// Acknowledge both messages
+		consumer.Ack(redelivered1)
+		consumer.Ack(redelivered2)
+	}
+
+	// Test both consumers
+	testPrecisionBitBehavior(consumerDefault, producerDefault, "default")
+	testPrecisionBitBehavior(consumerExplicit, producerExplicit, "explicit")
+}
```
RobertIndie: Is it possible to reuse the testing logic in `TestNegativeAckPrecisionBitCnt`?

pulsar/negative_acks_tracker.go
```diff
@@ -153,3 +201,7 @@ func (t *negativeAcksTracker) Close() {
 		t.doneCh <- nil
 	})
 }
+
+func Ptr[T any](v T) *T { return &v }
```
RobertIndie: Please don't expose this method.

pulsar/negative_acks_tracker.go
```diff
@@ -21,35 +21,57 @@ import (
...
+	"github.com/RoaringBitmap/roaring/v2/roaring64"
 	log "github.com/apache/pulsar-client-go/pulsar/log"
+	"github.com/emirpasic/gods/trees/avltree"
 )
 
 type redeliveryConsumer interface {
 	Redeliver(msgIDs []messageID)
 }
 
+type LedgerID = int64
```
RobertIndie: Please don't expose this type.

PR #1417, opened 2025-09-03
fix: add DLQPolicy.DeadLetterTopicProducerName

thomas-bousquet:
### Motivation

When the retry mechanism is enabled, the RLQ producer first sends messages to the RLQ topic using the producer name if passed in the producer options inside the `DLQPolicy`. 
However, **with `DLQPolicy.ProducerOptions.Name` set**, when the delivery count is reached and the consumer `Nack()` the messages, **the DLQ producer sending messages to the DLQ uses RLQ producer name.**

This is an issue, especially for message deduplication process where each producers must have a globally unique name for a given topic.

### Modifications

This PR adds the `DeadLetterTopicProducerName` field to the `DLQPolicy` in order to either allow using a custom producer name or let the client create the producer name with the existing naming pattern.

### Verifying this change

  - Added integration test `TestWithoutDeadLetterTopicDeadLetterTopicProducerName`
  - Added integration test `TestWithDeadLetterTopicDeadLetterTopicProducerName`
 
  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

  - Does this pull request introduce a new feature?  no
  - The `DeadLetterTopicProducerName` field added is documented where it's defined in the `DLQPolicy` struct

### DIFF
--- a/pulsar/consumer.go
+++ b/pulsar/consumer.go
@@ -75,7 +75,7 @@ type DLQPolicy struct {
 	// DeadLetterTopic specifies the name of the topic where the failing messages will be sent.
 	DeadLetterTopic string
 
-	// DeadLetterTopicProducerName specifies a name for the producer specifically for the DQL topic.
+	// DeadLetterTopicProducerName specifies a name for the producer specifically for the DLQ topic.
 	// If not assigned, the system will generate a globally unique name which can be access with
 	// Producer.ProducerName().
 	// When specifying a name, it is up to the user to ensure that, for a given topic, the producer name is unique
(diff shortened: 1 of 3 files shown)


pulsar/consumer.go
```diff
@@ -75,6 +75,14 @@ type DLQPolicy struct {
 	// DeadLetterTopic specifies the name of the topic where the failing messages will be sent.
 	DeadLetterTopic string
 
+	// DeadLetterTopicProducerName specifies a name for the producer specifically for the DQL topic.
```
lhotari:
```suggestion
	// DeadLetterTopicProducerName specifies a name for the producer specifically for the DLQ topic.
```

thomas-bousquet: Thank you, fixed.

thomas-bousquet:
Looks like CI failed with a timeout in the `1.23` run. Would it be possible to rerun it since it passed with `1.24`?

PR #1425, opened 2025-09-25
feat: add WithContext() admin client methods

thomas-bousquet:
### Motivation

The admin client does not carry / accept the context `context.Context`. Accepting a context as a first parameter is Go idiom, and can be useful when a consumer wants to set a request timeout instead of tracking manually a timeout with separate goroutine.

### Modifications

In order to accept that context and make this a non-breaking change, a `WithContext()` method equivalent  is added to each of the methods in the admin client. Existing methods are now calling internally the `WithContext` method, with a `context.Background()` for the required context.

### Verifying this change

This change can be verified with existing E2E tests on the admin client for regression.

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsaradmin/pkg/rest/client.go
+++ b/pulsaradmin/pkg/rest/client.go
@@ -19,6 +19,7 @@ package rest
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"io"
 	"net/http"
@@ -67,8 +68,8 @@ func (c *Client) newRequest(method, path string) (*request, error) {
 	return req, nil
 }
 
-func (c *Client) doRequest(r *request) (*http.Response, error) {
-	req, err := r.toHTTP()
+func (c *Client) doRequest(ctx context.Context, r *request) (*http.Response, error) {
+	req, err := r.toHTTP(ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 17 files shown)


pulsaradmin/pkg/rest/client.go
```diff
@@ -90,45 +91,45 @@ func (c *Client) doRequest(r *request) (*http.Response, error) {
 }
 
 // MakeRequest can make a simple request and handle the response by yourself
-func (c *Client) MakeRequest(method, endpoint string) (*http.Response, error) {
+func (c *Client) MakeRequest(ctx context.Context, method, endpoint string) (*http.Response, error) {
```
nodece: For the public API, maintain backward compatibility by adding new methods that accept a context parameter instead of changing existing method signatures.

Example:
```
 func (c *Client) MakeRequestWithContext(ctx context.Context, method, endpoint string) (*http.Response, error)
```

thomas-bousquet: Sounds good I will make the similar change I did for the rest. I did not realize this was exported, thanks for the feedback

thomas-bousquet: I went through the changes and added `...WithContext()` methods to the client as well

PR #1437, opened 2025-11-13
Fix partition update failure causing existing producers to close

RobertIndie:
### Motivation

When a partition update is triggered and the creation of one of the new producers fails, it currently fails all existing producers here:https://github.com/apache/pulsar-client-go/blob/148c10c41f37ee1e15bb66d7ba09d2a8984b87b0/pulsar/producer_impl.go#L267-L271 This causes the entire producer to become unavailable. All subsequent produce operations will fail with a "producer already closed" error.

### Modifications

Currently, two variables, `producers` and `producersPtr`, are used to manage the producer list inside producer_impl:

https://github.com/apache/pulsar-client-go/blob/148c10c41f37ee1e15bb66d7ba09d2a8984b87b0/pulsar/producer_impl.go#L55-L56

Both are accessed across multiple methods inside `producer_impl`, which is misleading and makes `producersPtr` ineffective.

This PR addresses the issue by:

* Using only `producers` as a atmoic value to access the producer list inside `producer_impl`. Only `internalCreatePartitionsProducers` can modify `producers`, which protects the producer list.
* Ensuring that during a partition update, new producers are not added to `producers` until all new producers are successfully created.

### Verifying this change

This change added tests

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/producer_impl.go
+++ b/pulsar/producer_impl.go
@@ -47,7 +47,6 @@ const (
 )
 
 type producer struct {
-	sync.RWMutex
 	client        *client
 	options       *ProducerOptions
 	topic         string
@@ -192,9 +191,6 @@ func (p *producer) internalCreatePartitionsProducers() error {
 	oldNumPartitions := 0
 	newNumPartitions := len(partitions)
 
-	p.Lock()
-	defer p.Unlock()
-
 	oldProducers := p.getProducers()
 	oldNumPartitions = len(oldProducers)
 
@@ -286,9 +282,6 @@ func (p *producer) Topic() string {
 }
 
 func (p *producer) Name() string {
-	p.RLock()
-	defer p.RUnlock()
-
 	return p.getProducer(0).Name()
 }
 
@@ -307,6 +300,9 @@ func (p *producer) SendAsync(ctx context.Context, msg *ProducerMessage,
 
 func (p *producer) getProducer(partition int) Producer {
 	producers := p.getProducers()
+	if len(producers) == 0 {
+		panic("producer has not been initialized properly")
+	}
 	if partition >= len(producers) {
 		// We read the old producers list while the count was already
 		// updated
@@ -330,9 +326,6 @@ func (p *producer) getPartition(msg *ProducerMessage) Producer {
 }
 
 func (p *producer) LastSequenceID() int64 {
-	p.RLock()
-	defer p.RUnlock()
-
 	var maxSeq int64 = -1
 	for _, pp := range p.getProducers() {
 		s := pp.LastSequenceID()
@@ -348,9 +341,6 @@ func (p *producer) Flush() error {
 }
 
 func (p *producer) FlushWithCtx(ctx context.Context) error {
-	p.RLock()
-	defer p.RUnlock()
-
 	for _, pp := range p.getProducers() {
 		if err := pp.FlushWithCtx(ctx); err != nil {
 			return err
@@ -363,9 +353,6 @@ func (p *producer) FlushWithCtx(ctx context.Context) error {
 func (p *producer) Close() {
 	p.closeOnce.Do(func() {
 		p.stopDiscovery()
-
-		p.Lock()
-		defer p.Unlock()
 
 		producers := p.getProducers()
 		for _, pp := range producers {


BewareMyPower (commented):
I suggest not refactor the `producers` related code. Please just address the issue that existing producers are closed for such a failure.

Actually I think the existing design of https://github.com/apache/pulsar-client-go/pull/286 is not safe.

```go
	producers := *(*[]Producer)(atomic.LoadPointer(&p.producersPtr))
	if partition >= len(producers) {
```

`producers` refers the memory of the slice that `p.producersPtr` points to, which is the pair of the pointer and the length.

However, when `p.producers = make([]Producer, newNumPartitions)` is called, the slice content could be modified, so the following race condition could happen:

| time | goroutine | action |
| :- | :- | :- |
| t1 | A | `atomic.LoadPointer(&p.producersPtr)` loads the pointer `p0` |
| t2 | B | modify the length from 5 to 10 |
| t3 | A | dereference the slice and read pointer `p0` and new length 10 |
| t4 | B | modify the pointer from `p0` to `p1` |

The key point is that the modification of a slice (pointer and length) is not atomic.

pulsar/producer_impl.go
```diff
@@ -277,8 +280,8 @@ func (p *producer) internalCreatePartitionsProducers() error {
 	} else {
 		p.metrics.ProducersPartitions.Add(float64(partitionsToAdd))
 	}
-	atomic.StorePointer(&p.producersPtr, unsafe.Pointer(&p.producers))
-	atomic.StoreUint32(&p.numPartitions, uint32(len(p.producers)))
+	atomic.StorePointer(&p.producersPtr, unsafe.Pointer(&producers))
```
BewareMyPower: This modification make it worse than the existing code. Now `producers` is just a local field, so ~~it could be garbaged collected even if it's tracked by an **UNSAFE POINTER**.~~ (determined by the implementation details)

> Dangling Pointers: If you create an unsafe pointer to this local slice and then return that pointer (or otherwise allow it to "escape" the function's scope), you'll end up with a dangling pointer. This pointer points to memory that is no longer valid for your use.

BewareMyPower: https://pkg.go.dev/unsafe#Pointer

> Pointer therefore allows a program to defeat the type system and read and write arbitrary memory. It should be used with extreme care.
>
> The following patterns involving Pointer are valid. Code not using these patterns is likely to be invalid today or to become invalid in the future. Even the valid patterns below come with important caveats.

It's dangerous to use `unsafe.Pointer` in such cases no matter if it's safe with the current Golang implementation.

pulsar/producer_impl.go
```diff
@@ -319,12 +315,26 @@ func (p *producer) getPartition(msg *ProducerMessage) Producer {
 	return producers[partition]
 }
 
+func (p *producer) getProducers() []Producer {
+	if producers := p.producers.Load(); producers != nil {
+		return producers.([]Producer)
+	}
+	return []Producer{}
```
BewareMyPower, line 318: It's better to panic here. Otherwise, method call like `getProducer(partition)` could also panic but it's hard to see if it's caused by the out of range of `partition`.

pulsar/producer_impl.go
RobertIndie: I agree. I refactored the code to simply use `atomic.Value` to hold the producers slice. Each producer can now obtain the current number of partitions just by checking the length of that slice.

BewareMyPower (commented):
After this change, it seems that `p.Lock()` in `internalCreatePartitionsProducers` becomes unnecessary because it's added to guarantee the safety to access the `producers` slice. With an atomic value, the slice pointed by `producers` is always immutable.

However, I suggest just keeping `producers` a slice and acquire read lock in `getPartition`. Removing the lock in this case looks like a pre-mature optimization. The modification on `producers` is a rare case, most time it quickly returns in `internalCreatePartitionsProducers` because `oldNumPartitions == newNumPartitions`.

BewareMyPower:
I wrote an example locally:

```golang
package main

import (
	"fmt"
	"math/rand"
	"sync"
	"time"
)

func main() {
	slice := make([]int, 10)
	for i := 0; i < len(slice); i++ {
		slice[i] = rand.Intn(100)
	}
	wg := sync.WaitGroup{}
	wg.Add(2)
	lock := sync.Mutex{}
	go func() {
		start := time.Now()
		sum := int64(0)
		const n = 10000000
		for i := 0; i < n; i++ {
			partition := rand.Int()
			lock.Lock()
			sum = sum + int64(slice[partition%len(slice)])
			lock.Unlock()
		}
		finished := time.Since(start)
		fmt.Println(sum, " elapsed time:", finished, " avg: ", finished.Nanoseconds()/n, " ns")
		wg.Done()
	}()
	go func() {
		for i := 0; i < 100; i++ {
			lock.Lock()
			for j := 0; j < 10; j++ {
				slice = append(slice, rand.Intn(100))
			}
			lock.Unlock()
			time.Sleep(1 * time.Millisecond)
		}
		wg.Done()
	}()
	wg.Wait()
}
```

- goroutine 1: choose a random element from the slice and calculate the sum
- goroutine 2: append 10 random values to the slice

Example outputs:

```
502615270  elapsed time: 166.572958ms  avg:  16  ns
```

The lock contention is not as scary as thought by many people. I'm wondering how much can be improved by using such an atomic approach?

pulsar/producer_impl.go
RobertIndie: Yes. Add a panic check in the `getProducer(partition)`.

This method is also used at https://github.com/apache/pulsar-client-go/blob/7940ff88dd91b2af3b721b0e4c0cf4910bafdc53/pulsar/producer_impl.go#L194 so it's expected to return the empty producers slice during the init.

BewareMyPower (approved):

PR #1443, opened 2025-12-01
fix: enhance zero queue consumer reconnection handling and message permit management

crossoverJie:
### Motivation

https://github.com/apache/pulsar-client-go/pull/1434#issuecomment-3542398661

When the broker restarts or a topic is unloaded, the zero consumer sends an extra permit, causing the number of UnAckedMessages to increase.

### Modifications

https://github.com/apache/pulsar/blob/cc5e479d63103f81e3af833e8b06227d1a6563e1/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java#L107-L152

Following the Java client implementation, add a `waitingOnReceive` variable to determine whether a permit needs to be sent.

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsar/consumer_zero_queue_test.go
+++ b/pulsar/consumer_zero_queue_test.go
@@ -244,22 +244,52 @@ func TestReconnectConsumer(t *testing.T) {
 }
 
 func TestReconnectedBrokerSendPermits(t *testing.T) {
+	req := testcontainers.ContainerRequest{
+		Name:         "pulsar-test",
+		Image:        getPulsarTestImage(),
+		ExposedPorts: []string{"6650/tcp", "8080/tcp"},
+		WaitingFor:   wait.ForExposedPort(),
+		HostConfigModifier: func(config *container.HostConfig) {
+			config.PortBindings = map[nat.Port][]nat.PortBinding{
+				"6650/tcp": {{HostIP: "0.0.0.0", HostPort: "6659"}},
+				"8080/tcp": {{HostIP: "0.0.0.0", HostPort: "8089"}},
+			}
+		},
+		Cmd: []string{"bin/pulsar", "standalone", "-nfw"},
+	}
+	c, err := testcontainers.GenericContainer(context.Background(), testcontainers.GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+		Reuse:            true,
+	})
+	require.NoError(t, err, "Failed to start the pulsar container")
+	endpoint, err := c.PortEndpoint(context.Background(), "6650", "pulsar")
+	require.NoError(t, err, "Failed to get the pulsar endpoint")
+
 	sLogger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
 	client, err := NewClient(ClientOptions{
-		URL:    lookupURL,
+		URL:    endpoint,
 		Logger: plog.NewLoggerWithSlog(sLogger),
 	})
 	assert.Nil(t, err)
-	topic := newTopicName()
-	consumer, err := client.Subscribe(ConsumerOptions{
-		Topic:                   topic,
-		SubscriptionName:        "my-sub",
-		EnableZeroQueueConsumer: true,
-		Type:                    Shared, // using Shared subscription type to support unack subscription stats
+	adminEndpoint, err := c.PortEndpoint(context.Background(), "8080", "http")
+	assert.Nil(t, err)
+	admin, err := pulsaradmin.NewClient(&config.Config{
+		WebServiceURL: adminEndpoint,
 	})
 	assert.Nil(t, err)
-	admin, err := pulsaradmin.NewClient(&config.Config{})
-	assert.Nil(t, err)
+
+	topic := newTopicName()
+	var consumer Consumer
+	require.Eventually(t, func() bool {
+		consumer, err = client.Subscribe(ConsumerOptions{
+			Topic:                   topic,
+			SubscriptionName:        "my-sub",
+			EnableZeroQueueConsumer: true,
+			Type:                    Shared, // using Shared subscription type to support unack subscription stats
+		})
+		return err == nil
+	}, 30*time.Second, 1*time.Second)
 	ctx := context.Background()
 
 	// create producer
@@ -335,7 +365,7 @@ func TestReconnectedBrokerSendPermits(t *testing.T) {
 			require.Equal(c, subscriptionStats.MsgBacklog, int64(1))
 			require.Equal(c, subscriptionStats.Consumers[0].UnAckedMessages, 0)
 		}
-	}, 60*time.Second, 1*time.Second)
+	}, 30*time.Second, 1*time.Second)
 
 	// ack
 	msg, err := consumer.Receive(context.Background())
@@ -351,7 +381,7 @@ func TestReconnectedBrokerSendPermits(t *testing.T) {
 			require.Equal(c, subscriptionStats.MsgBacklog, int64(0))
 			require.Equal(c, subscriptionStats.Consumers[0].UnAckedMessages, 0)
 		}
-	}, 60*time.Second, 1*time.Second)
+	}, 30*time.Second, 1*time.Second)
 
 }
 


pulsar/consumer_zero_queue.go
```diff
@@ -71,11 +73,18 @@ func newZeroConsumer(client *client, options ConsumerOptions, topic string,
 		return nil, err
 	}
 	opts := newPartitionConsumerOpts(zc.topic, zc.consumerName, tn.Partition, zc.options)
-	conn, err := newPartitionConsumer(zc, zc.client, opts, zc.messageCh, zc.dlq, zc.metrics)
+	opts.zeroQueueConsumer = zc
+	opts.zeroQueueReconnectedPolicy = func(pc *partitionConsumer, z *zeroQueueConsumer) {
```
RobertIndie: Is it better to capture the `*zeroQueueConsumer` in the lambda so that we can simplfy the `zeroQueueReconnectedPolicy`?

geniusjoe: +1

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,103 @@ func TestReconnectConsumer(t *testing.T) {
...
+
+	admin, err := pulsaradmin.NewClient(&config.Config{})
+	assert.Nil(t, err)
+	log.Println("unloading topic")
+	topicName, err := utils.GetTopicName(topic)
+	assert.Nil(t, err)
+	err = admin.Topics().Unload(*topicName)
+	assert.Nil(t, err)
+	log.Println("unloaded topic")
+	time.Sleep(5 * time.Second) // wait for topic unload finish
```
geniusjoe: Do we need to increase this sleep period? I'm not sure whether 5s is enough in CI environment.

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,103 @@ func TestReconnectConsumer(t *testing.T) {
...
+	//	send one more message and we do not manually receive it
+	_, err = producer.Send(ctx, &ProducerMessage{
+		Payload: []byte(fmt.Sprintf("hello-%d", 10)),
+		Key:     "pulsar",
+		Properties: map[string]string{
+			"key-1": "pulsar-1",
+		},
+	})
+	//	wait for broker send messages to consumer and topic stats update finish
+	time.Sleep(5 * time.Second)
```
geniusjoe: Maybe use assert.Eventually() for speeding up test case?

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,103 @@ func TestReconnectConsumer(t *testing.T) {
...
+	}
+
+	// ack
+	msg, err := consumer.Receive(context.Background())
+	assert.Nil(t, err)
+	err = consumer.Ack(msg)
+	assert.Nil(t, err)
+
+	// check topic stats
+	time.Sleep(5 * time.Second)
```
geniusjoe: Consider using assert.eventually() as well?

pulsar/consumer_zero_queue.go
crossoverJie: Good point, I’ll change `zeroQueueReconnectedPolicy` to capture `*zeroQueueConsumer` and drop the extra parameter/field.

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,135 @@ func TestReconnectConsumer(t *testing.T) {
...
+		log.Printf("send message: %s", msg.String())
+	}
+
+	log.Println("unloading topic")
+	topicName, err := utils.GetTopicName(topic)
+	assert.Nil(t, err)
+	err = admin.Topics().Unload(*topicName)
+	assert.Nil(t, err)
+	log.Println("unloaded topic")
+	time.Sleep(5 * time.Second) // wait for topic unload finish
```
nodece: Unnecessary

crossoverJie: Done

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,135 @@ func TestReconnectConsumer(t *testing.T) {
...
+	})
+	assert.Nil(t, err)
+	//	wait for broker send messages to consumer and topic stats update finish
+	time.Sleep(10 * time.Second)
+	topicStats, err := admin.Topics().GetStats(*topicName)
+	assert.Nil(t, err)
+	for _, subscriptionStats := range topicStats.Subscriptions {
+		assert.Equal(t, subscriptionStats.MsgBacklog, int64(1))
+		assert.Equal(t, subscriptionStats.Consumers[0].UnAckedMessages, 0)
+	}
```
nodece:
```suggestion
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		topicStats, err := admin.Topics().GetStats(*topicName)
		require.Nil(c, err)
		for _, subscriptionStats := range topicStats.Subscriptions {
			require.Equal(c, subscriptionStats.MsgBacklog, int64(1))
			require.Equal(c, subscriptionStats.Consumers[0].UnAckedMessages, 0)
		}
	}, 30*time.Second, 1*time.Second)	
```
Use `require.EventuallyWithT` instead of `time.Sleep`. The status may already be ready, and forcing a 10s wait is unnecessary and unreliable.

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,135 @@ func TestReconnectConsumer(t *testing.T) {
...
+	assert.Nil(t, err)
+
+	// check topic stats
+	time.Sleep(10 * time.Second)
+	topicStats, err = admin.Topics().GetStats(*topicName)
+	assert.Nil(t, err)
+	for _, subscriptionStats := range topicStats.Subscriptions {
+		assert.Equal(t, subscriptionStats.MsgBacklog, int64(0))
+		assert.Equal(t, subscriptionStats.Consumers[0].UnAckedMessages, 0)
+	}
```
nodece:
```suggestion
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		topicStats, err := admin.Topics().GetStats(*topicName)
		require.Nil(c, err)
		for _, subscriptionStats := range topicStats.Subscriptions {
			require.Equal(c, subscriptionStats.MsgBacklog, int64(0))
			require.Equal(c, subscriptionStats.Consumers[0].UnAckedMessages, 0)
		}
	}, 30*time.Second, 1*time.Second)	
```
Use `require.EventuallyWithT` instead of `time.Sleep`. The status may already be ready, and forcing a 10s wait is unnecessary and unreliable.

pulsar/consumer_zero_queue_test.go
crossoverJie: Good point👍

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,136 @@ func TestReconnectConsumer(t *testing.T) {
...
+		ExposedPorts: []string{"6650/tcp", "8080/tcp"},
+		WaitingFor:   wait.ForExposedPort(),
+		HostConfigModifier: func(config *container.HostConfig) {
+			config.PortBindings = map[nat.Port][]nat.PortBinding{
+				"6650/tcp": {{HostIP: "0.0.0.0", HostPort: "6659"}},
+				"8080/tcp": {{HostIP: "0.0.0.0", HostPort: "8089"}},
+			}
+		},
+		Cmd: []string{"bin/pulsar", "standalone", "-nfw"},
+	}
```
RobertIndie, line 259: Why do we need a dedicated container for the testing?

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,136 @@ func TestReconnectConsumer(t *testing.T) {
...
+	//	send one more message and we do not manually receive it
+	_, err = producer.Send(ctx, &ProducerMessage{
+		Payload: []byte(fmt.Sprintf("hello-%d", 10)),
+		Key:     "pulsar",
+		Properties: map[string]string{
+			"key-1": "pulsar-1",
+		},
+	})
+	assert.Nil(t, err)
+	//	wait for broker send messages to consumer and topic stats update finish
```
RobertIndie, line 357: We should unload the topic here to simulate the issue. I couldn't reproduce the original issue by reverting other changes from this PR.

RobertIndie (changes requested):
It seesm the test doesn't cover the issue. I couldn't reproduce the original issue by reverting other changes from this PR.

pulsar/consumer_zero_queue_test.go
crossoverJie:
```go
func TestUnloadTopicBeforeConsume2(t *testing.T) {

	sLogger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
		URL:    lookupURL,
		Logger: plog.NewLoggerWithSlog(sLogger),
	})
	assert.Nil(t, err)
	admin, err := pulsaradmin.NewClient(&config.Config{})
	assert.Nil(t, err)

	defer client.Close()

	topic := newTopicName()
	ctx := context.Background()
		Topic:                   topic,
		SubscriptionName:        "my-sub",
		EnableZeroQueueConsumer: true,
		Type:                    Shared,
	})

	assert.Nil(t, err)
	_, ok := consumer.(*zeroQueueConsumer)
	assert.True(t, ok)
	defer consumer.Close()

	// create producer
		Topic:           topic,
		DisableBatching: false,
	})
	assert.Nil(t, err)

	// send 10 messages
	for i := 0; i < 10; i++ {
		msg, err := producer.Send(ctx, &ProducerMessage{
			Key:     "pulsar",
			Properties: map[string]string{
				"key-1": "pulsar-1",
			},
		})
		assert.Nil(t, err)
		log.Printf("send message: %s", msg.String())
	}

	log.Println("unloading topic")
	topicName, err := utils.GetTopicName(topic)
	assert.Nil(t, err)
	err = admin.Topics().Unload(*topicName)
	assert.Nil(t, err)
	log.Println("unloaded topic")
	time.Sleep(1 * time.Minute) // wait for topic unload finish

	// receive 10 messages
	for i := 0; i < 10; i++ {
		if err != nil {
			log.Fatal(err)
		}

		expectMsg := fmt.Sprintf("hello-%d", i)
		expectProperties := map[string]string{
			"key-1": "pulsar-1",
		}
		assert.Equal(t, []byte(expectMsg), msg.Payload())
		assert.Equal(t, "pulsar", msg.Key())
		assert.Equal(t, expectProperties, msg.Properties())
		// ack message
		err = consumer.Ack(msg)
		assert.Nil(t, err)
		log.Printf("receive message: %s", msg.ID().String())
	}
	//	send one more message and we do not manually receive it
	_, err = producer.Send(ctx, &ProducerMessage{
		Payload: []byte(fmt.Sprintf("hello-%d", 11)),
		Key:     "pulsar",
		Properties: map[string]string{
			"key-1": "pulsar-1",
		},
	})
	//	wait for broker send messages to consumer and topic stats update finish
	time.Sleep(1 * time.Minute)
	topicStats, err := admin.Topics().GetStats(*topicName)
	for _, subscriptionStats := range topicStats.Subscriptions {
		assert.Equal(t, subscriptionStats.Consumers[0].UnAckedMessages, 0)
	}
}
```
You can switch to the master branch's code and then run this unit test to reproduce the issue.

image

It sends the permit 11 times (sending one extra), causing `UnackedMessages` to increase.

----
image

After using the fixed code, only 10 permits are sent, and the unit test also passes.

pulsar/consumer_zero_queue_test.go
crossoverJie: It's possible to run things locally without containers, but in CI, the `GetStats` data is inaccurate (perhaps the broker running in CI is resource-constrained), causing unit tests to fail.

Everything worked fine after I switched to containers.

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,136 @@ func TestReconnectConsumer(t *testing.T) {
...
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		topicStats, err := admin.Topics().GetStats(*topicName)
+		require.Nil(c, err)
+		for _, subscriptionStats := range topicStats.Subscriptions {
+			require.Equal(c, subscriptionStats.MsgBacklog, int64(1))
+			require.Equal(c, subscriptionStats.Consumers[0].UnAckedMessages, 0)
+		}
+	}, 30*time.Second, 1*time.Second)
```
geniusjoe: Maybe tick time be a 5 seconds, so that broker can have enough time to send messages to consumer

pulsar/consumer_zero_queue_test.go
```diff
@@ -243,6 +243,136 @@ func TestReconnectConsumer(t *testing.T) {
...
+
+	// check topic stats
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		topicStats, err := admin.Topics().GetStats(*topicName)
+		require.Nil(c, err)
+		for _, subscriptionStats := range topicStats.Subscriptions {
+			require.Equal(c, subscriptionStats.MsgBacklog, int64(0))
+			require.Equal(c, subscriptionStats.Consumers[0].UnAckedMessages, 0)
+		}
+	}, 30*time.Second, 1*time.Second)
```
geniusjoe: Maybe adjust tick time as well

pulsar/consumer_zero_queue_test.go
geniusjoe: @nodece I think maybe we need a  `time.Sleep(1 * time.Minute) // wait for topic unload finish` after topic unload, since unload() it's an async operation, and we need to make sure we call `consumer.Receive()` after topic unloaded finish.

pulsar/consumer_zero_queue_test.go
RobertIndie: Yes. I can reproduce the issue using the test `TestUnloadTopicBeforeConsume2 `. But I couldn't reproduce using `TestReconnectedBrokerSendPermits`. We should unload or restart the broker during the test to siumulate the issue.

pulsar/consumer_zero_queue_test.go
RobertIndie: I checked this failure: https://github.com/apache/pulsar-client-go/actions/runs/19845548673/job/56862120586#step:4:9441. 
Related codes:
https://github.com/crossoverJie/pulsar-client-go/blob/e6e9c045dba77b581835d0e76f3af89cd65979b6/pulsar/consumer_zero_queue_test.go#L324

It seems this is due to the inaccuracy of the returned message backlog. 

Maybe you can use the flag `-gpb, --get-precise-backlog`

crossoverJie: https://github.com/apache/pulsar-client-go/actions/runs/19920933910/job/57109331683?pr=1443
Tried using this flag, but it still didn't work.

pulsar/consumer_zero_queue_test.go
crossoverJie: This is because the `sleep` code after `unload` was previously commented out, causing the consumer not to trigger the reconnection mechanism.

I've added a `reconnectCount` field to record the number of reconnection attempts, ensuring the reconnection mechanism is definitely triggered.

> This field can also be used as a monitoring metric for the consumer in the future.


You can now reproduce this issue on the master branch using the latest code of `TestReconnectedBrokerSendPermits` (replacing `reconnectCount` with `sleep`).

pulsar/consumer_zero_queue_test.go
crossoverJie: @geniusjoe I added a `reconnectCount` field to ensure the consumer always triggers the reconnection mechanism, so there's no need for `sleep` here.

RobertIndie (approved):

pulsar/consumer_zero_queue_test.go
crossoverJie: All current tests have passed, so changing the tick time isn't very meaningful.

PR #1444, opened 2025-12-01
fix: return error when the client transaction coordinator is nil to p…

thomas-bousquet:
### Motivation

When the Pulsar client is created, if `EnableTransaction` is `true` then the transaction coordinator client is created.

If an application tries to create a transaction with  `NewTransaction()` when `EnableTransaction` is `false`, the application will panic; returning an error when the transaction coordinator is `nil` will prevent this.

### Modifications

In `NewTransaction()`, a nil check is added . If the transaction coordination client is `nil`, the method now returns an error.

### Verifying this change

  - The public API: no
  - The schema: no
  - The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/client_impl.go
+++ b/pulsar/client_impl.go
@@ -41,7 +41,7 @@ const (
 	minConnMaxIdleTime                 = 60 * time.Second
 )
 
-var errClientTransactionsNotEnabled = errors.New("transactions are not enabled with the client")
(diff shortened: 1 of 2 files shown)


crossoverJie:
Please add a unit test.

crossoverJie (approved):

PR #1490, opened 2026-05-02
[fix][consumer] Add reconnect failure listener and auto-close on max retry exhaustion

### LINKED ISSUE #1481: partitionConsumer reconnectToBroker reaches max retry but no notification/closure to external user
# Issue: partitionConsumer reconnectToBroker reaches max retry but no notification/closure to external user
## Environment
- pulsar-go sdk version: 0.16
- Component: partition consumer reconnection logic

## Problem Description
In the `reconnectToBroker` method of `partitionConsumer`, when the internal reconnection to broker reaches the **max retry count** set by `maxReconnectToBroker` option, there is **no mechanism to notify the external consumer user**, and the consumer is **not closed/stopped** either.

This causes critical usability issues:
1. We cannot listen/monitor when the internal reconnection logic exhausts all retries
2. The consumer remains in a broken state (silent failure)
3. We cannot safely recreate consumer/producer from outside because we have no event/callback for max reconnect failure
4. Our service frequently hits internal reconnection logic failures but has no way to react

## Root Cause Analysis
In the retry logic:
1. When `maxRetry == 0` or `bo.IsMaxBackoffReached()`, it only increments a metric (`ConsumersReconnectMaxRetry.Inc()`)
2. **No error is returned to the caller**
3. **No callback/event is triggered**
4. **The consumer is not closed or transitioned to a failed state**
5. The retry loop just exits silently, leaving the consumer in an invalid state

The core logic that causes the issue:
```go
if maxRetry == 0 || bo.IsMaxBackoffReached() {
    pc.metrics.ConsumersReconnectMaxRetry.Inc()
}
return struct{}{}, err
```
Even after max retry reached, it still returns `err` but the retry loop swallows it, and **no failure notification is sent to the consumer user**.

## Expected Behavior
When internal broker reconnection reaches max retry limit:
1. **Trigger a failure notification/callback** to the external consumer (e.g. error channel, connection listener, or state change)
2. **Close/stop the consumer** automatically (or provide an option to do so)
3. Expose consumer state change so external code can detect the failure
4. Allow users to react (recreate consumer/producer, alert, etc.) when max reconnect fails

## Actual Behavior
- Max reconnect retry reached → only metric incremented
- No notification to external user
- Consumer not closed
- No way to monitor this failure from outside
- Consumer stuck in broken state

## Code Snippet (Problematic Logic)
```go
if maxRetry == 0 || bo.IsMaxBackoffReached() {
    pc.metrics.ConsumersReconnectMaxRetry.Inc()
}

return struct{}{}, err
```
This is the key area that lacks failure handling and user notification.

## Impact
- Production services using pulsar-go 0.16 cannot reliably handle broker connection failures
- Silent consumer failures lead to message consumption stop without alerting
- Cannot implement automatic recovery (recreate consumer/producer) because no failure event is exposed
- Poor observability into internal reconnection failures

## Suggested Fix
1. Add a **reconnection failure callback/listener** when max retry reached
2. Close the consumer and mark it as failed after max retries
3. Propagate the error to the consumer's error channel
4. Update consumer state to a terminal failed state

PavelZeger:
Fixes #1481

### Motivation

When a `partitionConsumer` exhausts all broker reconnection attempts (controlled by
`MaxReconnectToBroker`), the client silently increments a metric and exits the retry loop,
leaving the consumer alive but unable to receive messages. There is no way for application
code to detect this failure or react to it (e.g. recreate the consumer or alert on-call).

### Modifications

Added two opt-in fields to `ConsumerOptions`:

- **`MaxReconnectToBrokerListener func(consumer Consumer, err error)`** — a callback invoked
  exactly once, on the same internal goroutine, immediately after the last reconnect attempt
  fails. The `consumer` argument is the parent `Consumer` the application holds, and `err`
  is the last connection error. The listener fires whenever `MaxReconnectToBroker` retries
  are exhausted **or** when the configured backoff policy signals `IsMaxBackoffReached`.

- **`CloseConsumerOnMaxReconnectToBroker bool`** — when `true`, automatically closes the
  consumer after exhausting reconnect attempts. The close runs asynchronously after
  `MaxReconnectToBrokerListener` (if set) returns. Internally `parentConsumer.Close()` is
  launched in a goroutine; this cancels the consumer's context, which unblocks the
  `internal.Retry` loop, allowing `runEventsLoop` to process the close request without
  deadlocking.

Both fields default to their zero values (`nil` / `false`), so there is no behaviour change
for existing consumers.

### Why points 3 and 4 from the issue are not implemented in this PR

The original issue suggested two additional fixes:

**3. Propagate the error to the consumer's error channel**

The `Consumer` interface does not expose an error channel. Adding one would be a breaking
API change: every implementation (`consumer`, `consumer_multitopic`, `consumer_regex`,
`consumer_zero_queue`) would need a new method, and all existing callers that perform a
type-assertion or embed the interface would break. This is a larger design decision that
warrants its own issue and a deprecation / migration path. The `MaxReconnectToBrokerListener`
callback achieves the same observable outcome (application code is notified of the failure)
without modifying the public interface.

**4. Update consumer state to a terminal "failed" state**

There is currently no `consumerFailed` state in the internal state machine
(`consumerInit → consumerReady → consumerClosing → consumerClosed`). Introducing a new
terminal state would require updating every state guard in `consumer_partition.go`
(there are more than a dozen) as well as the multi-topic and regex consumer wrappers.
In practice, enabling `CloseConsumerOnMaxReconnectToBroker` already transitions the
consumer through `consumerClosing → consumerClosed`, which is the correct terminal state
and prevents any further operations on a dead consumer. A separate "failed" state that
carries an error cause can be considered as a follow-up if observability tooling needs to
distinguish a failed-closed consumer from a normally-closed one.

### Verifying this change

This change added behaviour that requires a running broker to test end-to-end. Unit-level
verification can be done by constructing a `partitionConsumerOpts` directly with a
`maxReconnectToBroker` of 1 and asserting the listener fires and the consumer closes.
Integration test coverage is tracked as a follow-up.

- The public API: yes — two new opt-in fields added to `ConsumerOptions`
- The schema: no
- The wire protocol: no

### Documentation

- If yes, how is the feature documented? GoDocs on the new `ConsumerOptions` fields


pulsar/consumer_partition.go
```diff
@@ -2100,6 +2102,12 @@ func (pc *partitionConsumer) reconnectToBroker(connectionClosed *connectionClose
 		pc.metrics.ConsumersReconnectFailure.Inc()
 		if maxRetry == 0 || bo.IsMaxBackoffReached() {
 			pc.metrics.ConsumersReconnectMaxRetry.Inc()
+			if pc.options.maxReconnectToBrokerListener != nil {
+				pc.options.maxReconnectToBrokerListener(pc.parentConsumer, err)
+			}
```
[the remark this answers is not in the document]
PavelZeger: I confirmed the bug. With `MaxReconnectToBroker` unset, `bo.IsMaxBackoffReached()` stays true and the branch returns err, so the retry loop  re-enters indefinitely — listener fires every iteration, close goroutines pile up.                                                  
                                                                                                                                      
Didn't take the "make it terminal" option: it would silently change the default contract of retrying forever at the backoff ceiling.

Fix: closure-local `maxRetryNotified bool` gating only the listener and close goroutine; retry flow and metric untouched. Plain bool  is fine since `opFn` runs sequentially.                                                                                             
                                                                                                                                      
Added `TestConsumerMaxReconnectToBrokerListenerFiresOnceWhenBackoffMaxed: MaxReconnectToBroker` unset + always-maxed backoff, kills  broker, waits 3 s past first invocation, asserts counter == 1. Passes.

pulsar/consumer_partition.go
```diff
@@ -2100,6 +2102,12 @@ func (pc *partitionConsumer) reconnectToBroker(connectionClosed *connectionClose
...
+			if pc.options.closeConsumerOnMaxReconnectToBroker {
+				go pc.parentConsumer.Close()
+			}
```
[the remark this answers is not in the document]
PavelZeger: Addressed by the same patch as the previous thread - `go pc.parentConsumer.Close()` and `maxReconnectToBrokerListener` are both inside one `if !maxRetryNotified { … }` block, so they fire together exactly once per reconnect cycle. Used a  plain bool rather than `sync.Once/atomic` because `opFn` is invoked sequentially by `internal.Retry` - no concurrent entry into the closure, no race to protect against. Happy to swap to `sync.Once` if you'd prefer the explicit signal.

pulsar/consumer.go
```diff
@@ -222,6 +222,18 @@ type ConsumerOptions struct {
...
+	// MaxReconnectToBrokerListener is called when the consumer exhausts all reconnect attempts
+	// set by MaxReconnectToBroker. The consumer argument is the parent consumer, and err is the
+	// last connection error. Use this callback to detect silent failure and take recovery action
+	// (e.g. recreate the consumer). Only fires when MaxReconnectToBroker is set to a finite value
+	// or when the backoff policy signals IsMaxBackoffReached.
+	MaxReconnectToBrokerListener func(consumer Consumer, err error)
+
+	// CloseConsumerOnMaxReconnectToBroker, when true, automatically closes the consumer after
+	// exhausting all reconnect attempts. The close happens asynchronously after
+	// MaxReconnectToBrokerListener (if set) returns. Default: false.
```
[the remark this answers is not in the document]
PavelZeger: Addressed in the commit instead of committing suggestion because it was outdated due to the previous commit.

pulsar/consumer_test.go
```diff
@@ -5710,3 +5710,112 @@ func TestSelectConnectionForSameConsumer(t *testing.T) {
...
+		WaitingFor:   wait.ForExposedPort(),
+		Cmd:          []string{"bin/pulsar", "standalone", "-nfw", "--advertised-address", "localhost"},
+	}
+	c, err := testcontainers.GenericContainer(context.Background(), testcontainers.GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+	})
+	require.NoError(t, err)
+	endpoint, err := c.PortEndpoint(context.Background(), "6650", "pulsar")
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
PavelZeger: Addressed in latest commit: registered `t.Cleanup` right after container creation in all three tests as a best-effort safety net (logs on error), and changed the in-body `_ = c.Terminate(...)` to `require.NoError(t, c.Terminate(...))` since that call is part of the  scenario.

pulsar/consumer_test.go
```diff
@@ -5710,3 +5710,112 @@ func TestSelectConnectionForSameConsumer(t *testing.T) {
...
+	topic := newTopicName()
+	var testConsumer Consumer
+	require.Eventually(t, func() bool {
+		testConsumer, err = pulsarClient.Subscribe(ConsumerOptions{
+			Topic:                topic,
+			SubscriptionName:     "test-max-reconnect-listener",
+			MaxReconnectToBroker: &maxRetry,
+			BackOffPolicyFunc: func() backoff.Policy {
+				return newTestBackoffPolicy(100*time.Millisecond, 1*time.Second)
+			},
```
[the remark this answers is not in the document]
PavelZeger: Already addressed in a follow-up commit: added `TestConsumerMaxReconnectToBrokerListenerFiresOnceWhenBackoffMaxed` with a `maxBackoffReachedPolicy` whose `IsMaxBackoffReached()` always returns true. It asserts the listener fires exactly once even after many additional retry iterations, covering the previously untested exhaustion path.

pulsar/consumer_test.go
```diff
@@ -5710,3 +5710,112 @@ func TestSelectConnectionForSameConsumer(t *testing.T) {
...
+		Image:        getPulsarTestImage(),
+		ExposedPorts: []string{"6650/tcp", "8080/tcp"},
+		WaitingFor:   wait.ForExposedPort(),
+		Cmd:          []string{"bin/pulsar", "standalone", "-nfw", "--advertised-address", "localhost"},
+	}
+	c, err := testcontainers.GenericContainer(context.Background(), testcontainers.GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+	})
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
PavelZeger: Already addressed in the latest commit: `t.Cleanup` registered immediately after container creation in `TestConsumerMaxReconnectToBrokerAutoClose` (and the other two reconnect tests) - best-effort terminate that logs on error, with the in-body terminate now checked via `require.NoError`.

nodece:
Hi @PavelZeger, could you follow up on the Java client logic?

PavelZeger:
> Hi @PavelZeger, could you follow up on the Java client logic?

 Sure, @nodece. 
I pushed an update that takes the parts of your proposal we can apply without breaking the existing API:                                                                                                  
                                                                                                                                      
 **What's done now**                                                                                                                 
                                                                                                                                      
  1. **Backoff cap no longer ends retries or fires the listener.** Dropped the `IsMaxBackoffReached` check from `reconnectToBroker`.  The consumer now keeps retrying forever at the capped delay, just like Java's `ConnectionHandler.reconnectLater()`.                 
  2. **Non-retriable errors are detected and reported.** Added a check for the consumer-relevant errors that Java treats as non-retriable: `AuthorizationError`, `TopicNotFound`, `TopicTerminatedError`, `SubscriptionNotFound`, `IncompatibleSchema`, `ConsumerBusy`, `InvalidTopicName`, `ConsumerAssignError`, `NotAllowedError`. When the broker returns one of these, the listener fires with the error and the retry loop stops - same idea as Java's `connectionFailed()`.                                           
                                                                                                                                      
**What I left for later**                   
                                                                                                                                      
  3. I kept `MaxReconnectToBroker` as an opt-in cap on runtime reconnects. Issue #1481 specifically asks for a way to stop the consumer after a bounded number of failed reconnects, so removing it would break users who already set it. Can revisit if maintainers want it removed.
  4. Applying `MaxReconnectToBroker` (or a new `LookupTimeout`) at initial subscribe is a different code path. I'd rather file it as a follow-up issue so it gets its own focused review.                                                                               
                                                                                                                                      
Does this work for you? If you want me to go further on 3 or 4 in this PR, let me know — otherwise I'll open a follow-up for 4 after this PR.                   

  Sources:                                                                                                                            
  - https://raw.githubusercontent.com/apache/pulsar/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConnectionHandler.java                                                                                                                               
  - https://raw.githubusercontent.com/apache/pulsar/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java
  - https://raw.githubusercontent.com/apache/pulsar/master/pulsar-common/src/main/java/org/apache/pulsar/common/util/Backoff.java     
  - https://raw.githubusercontent.com/apache/pulsar/master/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/PulsarClientException.java

pulsar/consumer_partition.go
```diff
@@ -2087,19 +2146,19 @@ func (pc *partitionConsumer) reconnectToBroker(connectionClosed *connectionClose
...
 		}
 
 		if maxRetry > 0 {
 			maxRetry--
 		}
 		pc.metrics.ConsumersReconnectFailure.Inc()
-		if maxRetry == 0 || bo.IsMaxBackoffReached() {
+		if maxRetry == 0 {
 			pc.metrics.ConsumersReconnectMaxRetry.Inc()
+			notifyReconnectGiveUp(err)
```
nodece:
```suggestion
			notifyReconnectGiveUp(errors.New("max retry attempts reached for reconnecting to broker"))
			return struct{}{}, nil
```

pulsar/consumer_partition.go
```diff
@@ -2061,6 +2106,20 @@ func (pc *partitionConsumer) reconnectToBroker(connectionClosed *connectionClose
...
+			return
+		}
+		giveUpNotified = true
+		if pc.options.maxReconnectToBrokerListener != nil {
+			pc.options.maxReconnectToBrokerListener(pc.parentConsumer, cause)
+		}
+		if pc.options.closeConsumerOnMaxReconnectToBroker {
+			go pc.parentConsumer.Close()
+		}
+	}
```
nodece, line 2126:
```suggestion
	var giveUpNotified bool
	notifyReconnectGiveUp := func(cause error) {
		if giveUpNotified {
			return
		}
		giveUpNotified = true
        go pc.parentConsumer.Close()
	}
```

pulsar/consumer.go
```diff
@@ -222,6 +222,28 @@ type ConsumerOptions struct {
...
+	//   1. The retry budget set by MaxReconnectToBroker is exhausted.
+	//   2. The broker reports a non-retriable error (e.g. AuthorizationError, TopicNotFound,
+	//      TopicTerminated, IncompatibleSchema). In this case the listener fires regardless of
+	//      whether MaxReconnectToBroker was set, since retrying cannot recover.
+	// This callback is invoked from the partition consumer event loop, so applications must not
+	// call consumer.Close() synchronously from within the callback — doing so can deadlock. If
+	// closing is required, do it asynchronously (for example, in another goroutine), or enable
+	// CloseConsumerOnMaxReconnectToBroker to let the client close the consumer safely after the
+	// callback returns.
+	MaxReconnectToBrokerListener func(consumer Consumer, err error)
```
nodece: I suggest we add this feature to `ConsumerInterceptor`, so like `OnConsumerClose`.

pulsar/consumer.go
```diff
@@ -222,6 +222,28 @@ type ConsumerOptions struct {
...
+
+	// CloseConsumerOnMaxReconnectToBroker, when true, automatically closes the consumer after
+	// the reconnect loop gives up (either MaxReconnectToBroker exhausted or a non-retriable
+	// broker error was received). The close happens asynchronously after
+	// MaxReconnectToBrokerListener (if set) returns, and is the recommended option when the
+	// consumer should be closed after reconnect failure. Default: false.
+	CloseConsumerOnMaxReconnectToBroker bool
```
nodece: This config is unnecessary, and this value should be the `true` default.

nodece (changes requested):

pulsar/consumer_partition.go
nodece: Please fix code style

pulsar/consumer.go
PavelZeger: Before I push the changes, I want to confirm a couple of details  so I implement it the way you have in mind.                                                                                         
                                                                                                                                      
1. Signature - should it be `OnConsumerClose(consumer Consumer)` or `OnConsumerClose(consumer Consumer, cause error)`? The original issue #1481 asked for a way to know why the consumer gave up reconnecting, so passing the last error as cause would keep that  information. For a normal user-initiated Close(), cause would be nil. Do you prefer this, or the simpler version without the error?
2. When to call it - I plan to fire `OnConsumerClose` once per consumer, at the start of `consumer.Close()` (before teardown), guarded by a `sync.Once`. That way it fires exactly once whether the close comes from the user, from max-reconnect exhaustion, or from a non-retriable error. Does that match what you had in mind?
3. Breaking change - adding a method to ConsumerInterceptor will break existing implementations. I'll update the in-repo ones and  add a note to the changelog. Are you OK with this, or would you prefer a separate optional interface (e.g., `ConsumerCloseInterceptor`) to keep backward compatibility?

pulsar/consumer.go
PavelZeger: Agreed, I'll remove the option and make auto-close the only behavior (the runtime already does this after the last commit). I can update the `MaxReconnectToBroker` GoDoc to say "on exhaustion the consumer is closed; observe via ConsumerInterceptor.OnConsumerClose."

pulsar/consumer.go
nodece: @PavelZeger 

1. `OnConsumerClose(consumer Consumer, err error)` looks good. When `err` is nil, this means the consumer has been closed by the user.
2. This hook should be called once when the consumer has been closed.
3. Good catch, please add `ConsumerCloseInterceptor` to keep backward compatibility.

pulsar/consumer.go
PavelZeger: The comment has been addressed in the latest push.

pulsar/consumer.go
PavelZeger: The comment has been addressed in the latest push.

nodece (approved):
@PavelZeger Good work

nodece:
@PavelZeger please rebase the master branch, `TestPriorityConsumer ` has been fixed by https://github.com/apache/pulsar-client-go/pull/1495

PavelZeger:
> @PavelZeger please rebase the master branch, `TestPriorityConsumer ` has been fixed by #1495

Done. Waiting for workflow's approve.

PR #1492, opened 2026-05-10
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `pulsar/negative_backoff_policy.go:L30`

In pulsar/negative_backoff_policy.go, the backoff calculation uses bit shift (minNackTime << redeliveryCount) which can overflow when redeliveryCount is large. The overflow causes backoff to become 0, resulting in immediate retry without any delay, potentially causing a denial of service.

## Solution

Add overflow check before the shift operation: if redeliveryCount >= 63 { return maxNackTime }. Also consider using math.Min with a pre-computed maximum to avoid the overflow entirely.

## Changes

- `pulsar/negative_backoff_policy.go` (modified)
- `oauth2/oidc_endpoint_provider.go` (modified)

### DIFF
--- a/oauth2/oidc_endpoint_provider.go
+++ b/oauth2/oidc_endpoint_provider.go
@@ -40,6 +40,17 @@ func GetOIDCWellKnownEndpointsFromIssuerURL(issuerURL string) (*OIDCWellKnownEnd
 	if err != nil {
 		return nil, errors.Wrap(err, "could not parse issuer url to build well known endpoints")
 	}
+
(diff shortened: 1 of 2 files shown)


RobertIndie (commented):
Could you add a test to verify the change? And please refine the PR to reflect the change. It's better to split two issue fixes into 2 PRs.

PR #1493, opened 2026-05-11
fix: avoid nil partition consumer panic during ack

mattisonchao:
## Motivation
Partition auto-discovery can grow `consumer.consumers` before the newly-created partition consumers are assigned into the slice. During that window, `AckIDList` can receive a real message ID from the new partition and read a nil `*partitionConsumer` from `consumer.consumers[msgID.PartitionIdx()]`.

Without this fix, the deterministic regression test panics with the same nil partition consumer path:

```text
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x20 pc=0x1027d7cd8]

goroutine 544 [running]:
github.com/apache/pulsar-client-go/pulsar.(*partitionConsumer).AckIDList(0x0, {0x44909cd9e410?, 0x1?, 0x1?})
	/Users/mattison/projects/apache/pulsar-client-go/pulsar/consumer_partition.go:742 +0x28
github.com/apache/pulsar-client-go/pulsar.ackIDListFromMultiTopics.func1()
	/Users/mattison/projects/apache/pulsar-client-go/pulsar/consumer_multitopic.go:228 +0x30
created by github.com/apache/pulsar-client-go/pulsar.ackIDListFromMultiTopics in goroutine 395
	/Users/mattison/projects/apache/pulsar-client-go/pulsar/consumer_multitopic.go:227 +0x27c
```

## Modifications
- Guard partition consumer lookup for `AckWithTxn`, `AckID`, `AckIDList`, and `AckIDCumulative` with the parent consumer `RWMutex`.
- Add a shared lookup helper that validates the partition index and rejects nil partition consumers instead of dispatching to them.
- Add a test-only `FailureInjectHook` that pauses partition consumer assignment at the deterministic race window.
- Add a root-level `tests` package with a generic Pulsar testcontainer and an `AckIDList` regression test using a real message from the newly discovered partition.

This is the smallest fix that avoids the panic without refactoring partition consumer publication. A higher-performance approach would require a broader redesign of how partition consumers are installed and read.

## Testing
- `PULSAR_IMAGE=apachepulsar/pulsar:4.2.0 env -u GOROOT /Users/mattison/.local/share/mise/installs/go/1.26.1/bin/go test ./tests -run '^TestConsumerAckIDListDuringPartitionUpdate$' -count=1 -timeout 5m -v`
- `env -u GOROOT /Users/mattison/.local/share/mise/installs/go/1.26.1/bin/go test ./pulsar -run '^$' -count=1`
- Verified the regression test panics without the production fix using the stack trace above.


BewareMyPower:
It's simply a thread safety issue, I don't think it's worth wasting token to let LLM write complicated tests to reproduce.

I pushed a PR instead: https://github.com/apache/pulsar-client-go/pull/1494

PR #1494, opened 2026-05-12
fix: the access to consumer.consumers might not be thread safe

BewareMyPower:
`c.consumers` can be updated by `internalTopicSubscribeToPartitions`, so we must access `c.consumers` when holding a lock. However, many code places don't follow the implicit requirement, so nil pointer dereference could happen.

This PR replaces the combination of c.checkMsgIDPartition(msgID) and c.consumers[msgID] with a wrapped method findPartitionConsumer, which retrieves the consumer with lock held. Specifically, for Seek, add an unsafe version (i.e. no lock acquired) because the lock has been acquired to check if len(c.consumers) > 1 before.

For other places that access `c.consumers` (all are range for loops), ensure lock is acquired. The exceptional case is the `hasNext` method, we have to iterate over a snapshot, otherwise, `c.Unlock()` must be called after `wg.Wait()`, which is impossible

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -549,11 +549,11 @@ func (c *consumer) Receive(ctx context.Context) (message Message, err error) {
 
 func (c *consumer) AckWithTxn(msg Message, txn Transaction) error {
 	msgID := msg.ID()
-	consumer, err := c.findPartitionConsumer(msgID)
-	if err != nil {
+	if err := c.checkMsgIDPartition(msgID); err != nil {
 		return err
 	}
-	return consumer.AckIDWithTxn(msgID, txn)
+
+	return c.consumers[msgID.PartitionIdx()].AckIDWithTxn(msgID, txn)
 }
 
 // Chan return the message chan to users
@@ -568,19 +568,23 @@ func (c *consumer) Ack(msg Message) error {
 
 // AckID the consumption of a single message, identified by its MessageID
 func (c *consumer) AckID(msgID MessageID) error {
-	consumer, err := c.findPartitionConsumer(msgID)
-	if err != nil {
+	if err := c.checkMsgIDPartition(msgID); err != nil {
 		return err
 	}
+
 	if c.options.AckWithResponse {
-		return consumer.AckIDWithResponse(msgID)
+		return c.consumers[msgID.PartitionIdx()].AckIDWithResponse(msgID)
 	}
-	return consumer.AckID(msgID)
+
+	return c.consumers[msgID.PartitionIdx()].AckID(msgID)
 }
 
 func (c *consumer) AckIDList(msgIDs []MessageID) error {
 	return ackIDListFromMultiTopics(c.log, msgIDs, func(msgID MessageID) (acker, error) {
-		return c.findPartitionConsumer(msgID)
+		if err := c.checkMsgIDPartition(msgID); err != nil {
+			return nil, err
+		}
+		return c.consumers[msgID.PartitionIdx()], nil
 	})
 }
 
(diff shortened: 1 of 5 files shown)


pulsar/consumer_impl.go
```diff
@@ -781,15 +773,22 @@ func (c *consumer) SeekByTime(time time.Time) error {
...
 	partition := msgID.PartitionIdx()
 	if partition < 0 || int(partition) >= len(c.consumers) {
 		c.log.Errorf("invalid partition index %d expected a partition between [0-%d]",
 			partition, len(c.consumers))
-		return fmt.Errorf("invalid partition index %d expected a partition between [0-%d]",
+		return nil, fmt.Errorf("invalid partition index %d expected a partition between [0-%d]",
 			partition, len(c.consumers))
 	}
-	return nil
+	return c.consumers[msgID.PartitionIdx()], nil
```
nodece: I'm concerned that `c.consumers[msgID.PartitionIdx()]` could be `nil`.

BewareMyPower: Why could it be nil? There is already an index range check before. It should be safe with `c.Lock` acquired. If it's `nil`, there must be something wrong with the modification, which is also protected by `c.Lock`.

nodece: Ok, I just noticed this value is a pointer type.

crossoverJie:
I still noticed a few other direct accesses to `c.consumers`, for example:

- https://github.com/apache/pulsar-client-go/blob/master/pulsar/consumer_impl.go#L698
- https://github.com/apache/pulsar-client-go/blob/master/pulsar/consumer_impl.go#L710

Was the intention here to only fix the msgID-based paths, or is there a reason not to update these remaining access sites in the same PR as well?

BewareMyPower:
@crossoverJie Good finding! Let me try a better approach, maybe we need to wrap a thread-safe list

BewareMyPower:
@crossoverJie Now I replaced the trivial slice with an `atomic.Value`, PTAL again

pulsar/consumer_impl.go
```diff
@@ -53,7 +54,7 @@ type consumer struct {
 	topic                     string
 	client                    *client
 	options                   ConsumerOptions
-	consumers                 []*partitionConsumer
+	consumers                 atomic.Value
```
mattisonchao: This change will make the `internalTopicSubscribeToPartitions` method not atomic. That means you can receive messages during the partition changes, but you can't ack them.

BewareMyPower: > you can receive messages during the partition changes, but you can't ack them.

Yes, it's correct. Let me think of a better approach

BewareMyPower:
I've reverted to the previous lock-based design and added comments for the issue [here](https://github.com/apache/pulsar-client-go/pull/1494#discussion_r3233525333). PTAL again @mattisonchao

BewareMyPower:
The failed blue-green migration is not caused by this PR.

----

Based on my analysis, **the test failure is NOT caused by the commits since master**. Here's why:

**Changes since master:**
All 8 commits only modify `pulsar/consumer_impl.go` — a pure refactoring of the `consumer` struct's `consumers` slice access for thread safety (replacing direct indexing with `findPartitionConsumer`/`unsafeFindPartitionConsumer`, adding a snapshot in `hasNext()`). The `sync.Mutex` was already embedded before these changes.

**Test failure root cause:**
The failure occurs in `consumer_partition.go` (unchanged by these commits). The `proxyConnection` sub-test times out because during topic migration reconnection:

1. `grabConn` sends a Subscribe request after getting a connection from the pool
2. The request times out (line 81: `request timed out` at `consumer_partition.go:2244`)
3. The timeout handler tries to close the consumer via `RequestOnCnx` on the same connection (`consumer_partition.go:2251-2256`), but this also likely fails silently
4. On subsequent retries, `GetConnection` returns the same pooled connection where the broker still has consumerID=1 registered from the timed-out subscribe, resulting in: `UnknownError: Consumer that failed is already present on the connection`

This is a pre-existing race condition in the reconnection logic — the consumer doesn't properly handle the case where a Subscribe request reaches the broker but the response doesn't reach the client.

nodece (approved):

crossoverJie (approved):

PR #1495, opened 2026-05-14
test: fix TestPriorityConsumer

nodece:
### Motivation

`TestPriorityConsumer` was failing with `context.deadlineExceededError` because the original test only sent 10 messages to 2 priority-1 consumers and passively asserted exact delivery counts (consumer1 gets 5, consumer2 gets 5). This is fragile — the broker distributes messages among consumers with available permits, so the exact count per consumer is not deterministic when messages are still being dispatched.

The Java reference test validates priority dispatch differently: it actively drains messages from specific consumers to replenish their flow-control permits, then sends additional messages and asserts the low-priority consumer receives nothing.

### Modifications

Rewrite `TestPriorityConsumer` to align with the Java reference implementation (`PriorityConsumerTest` in `DispatcherTest.java`):

1. Send 15 messages (distributed among all priority-1 consumers)
2. Drain 20 messages each from consumer1 and consumer2 to replenish their individual permits — now they have more available permits than consumer3 at the same priority level
3. Send 5 more messages — the broker should only dispatch to consumer1/consumer2 because they have more permits
4. Assert consumer3 (priority 2) receives nothing

### DIFF
--- a/pulsar/consumer_test.go
+++ b/pulsar/consumer_test.go
@@ -296,27 +296,33 @@ func TestPriorityConsumer(t *testing.T) {
 	assert.Nil(t, err)
 	defer producer.Close()
 
-	for i := 0; i < 10; i++ {
+	// Phase 1: Send 15 messages — distributed among the three priority-1 consumers
+	for i := 0; i < 15; i++ {
 		_, err := producer.Send(context.Background(), &ProducerMessage{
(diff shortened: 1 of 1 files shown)


geniusjoe (approved):
LGTM

PavelZeger (approved):

crossoverJie (approved):

PR #1499, opened 2026-05-19
fix: add transaction id to message metadata

### LINKED ISSUE #1498: Transactional publishes do not set txn id in MessageMetadata
﻿## Bug description

Transactional publishes currently put the transaction id on `CommandSend`, but the serialized `MessageMetadata` does not carry `txnid_most_bits` and `txnid_least_bits` for producer-created messages.

Because the broker-side transaction buffer uses the transaction id in message metadata to identify transactional entries, messages published inside a transaction can be treated as normal visible messages. In particular, after aborting the transaction, those messages may still be delivered to consumers.

## Expected behavior

Messages published with `ProducerMessage.Transaction` should include the transaction id in `MessageMetadata` as well as in `CommandSend`, so aborted transaction messages remain invisible and are removed by the transaction buffer.

## Actual behavior

The producer resolves the transaction id for send commands, but does not copy it into `MessageMetadata` for single-message sends or batch metadata.

## Suggested fix

When a send request has a transaction:

- set `MessageMetadata.TxnidMostBits` and `MessageMetadata.TxnidLeastBits` for single-message sends
- set the same fields on batch-level `MessageMetadata` when the first message in a batch is transactional
- clear the batch transaction fields after flushing so a later non-transactional batch cannot inherit them

jiangmocc:
﻿### Motivation

Fixes #1498.

Transactional sends already set the transaction id on `CommandSend`, but the serialized `MessageMetadata` did not carry `txnid_most_bits` and `txnid_least_bits`. Brokers rely on the message metadata transaction id when handling transaction buffer entries, so aborted transactional publishes can leak as visible messages.

### Modifications

- Set transaction id fields on `MessageMetadata` for single-message transactional sends.
- Set transaction id fields on batch-level `MessageMetadata` for transactional batches.
- Clear transaction id fields from reusable batch metadata and send command state after flushing.
- Add focused unit tests for single-message metadata, batch metadata, and batch reset behavior.

### Verifying this change

- [x] `GOWORK=off go test -v ./pulsar/internal -run "TestBatchBuilder(AddsTxnIDToMessageMetadata|ClearsTxnIDAfterFlush)" -count=1`
- [x] `GOWORK=off go test -v ./pulsar -run TestUpdateMetaDataAddsTxnID -count=1`

- Dependencies: no
- The public API: no
- The schema: no
- The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/internal/batch_builder.go
+++ b/pulsar/internal/batch_builder.go
@@ -233,6 +233,8 @@ func (bc *batchContainer) Add(
 		if useTxn {
 			bc.cmdSend.Send.TxnidMostBits = proto.Uint64(mostSigBits)
 			bc.cmdSend.Send.TxnidLeastBits = proto.Uint64(leastSigBits)
+			bc.msgMetadata.TxnidMostBits = proto.Uint64(mostSigBits)
+			bc.msgMetadata.TxnidLeastBits = proto.Uint64(leastSigBits)
 		}
 	}
 	addSingleMessageToBatch(bc.buffer, metadata, payload)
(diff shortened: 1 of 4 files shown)


RobertIndie (approved):

Ebispongebob:
Hi @jiangmocc, thanks for the fix! Just wanted to gently check if there are any plans to address the Copilot suggestion above regarding mixed-txn batch validation, or if it's safe to merge as-is.

@RobertIndie Since this was already approved and CI passed, could we help move this forward when convenient? Thanks!

PR #1500, opened 2026-05-19
fix: deadlock when increasing partitioned consumers

BewareMyPower:
### Motivation

#1494 introduces a possible deadlock after ensuring the thread safety.

To fix the thread safety issue, every APIs that need to find a specific sub-consumer, like `Ack` and `Seek`, will now require locking the parent consumer. However, in `internalTopicSubscribeToPartitions`, it will wait all sub-consumers' creations are done:

https://github.com/apache/pulsar-client-go/blob/87ce8f90d0e2df9402133ebc52560638bb09778d/pulsar/consumer_impl.go#L414-L416

https://github.com/apache/pulsar-client-go/blob/87ce8f90d0e2df9402133ebc52560638bb09778d/pulsar/consumer_impl.go#L428

`range ch` will wait the `err` returned by `newPartitionConsumer` is sent to the `ch`.

However, it could be blocked by `grabConn`: https://github.com/apache/pulsar-client-go/blob/87ce8f90d0e2df9402133ebc52560638bb09778d/pulsar/consumer_partition.go#L434

this could start a Subscribe RPC in the connection that might receive user's Ack or Seek requests.

### Modifications

- Add `TestInternalTopicSubscribeToPartitionsDoesNotBlockExistingPartitionLookup` to reproduce the deadlock issue.
- Adopt the lock-free implementation to manage consumers, the test above will pass after test
- Delay the dispatching logic after adding new sub-consumers to `c.consumers`, otherwise, a message could be received from a new sub-consumer and acknowledged before that consumer is added to `c.consumers`.
- Add `TestInternalTopicSubscribeToPartitionsPublishesConsumersBeforeDispatchingMessages` to verify it works

### DIFF
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -784,7 +784,7 @@ func findPartitionConsumer(consumers []*partitionConsumer, msgID MessageID) (*pa
 	partition := int(msgID.PartitionIdx())
 	if partition < 0 || partition >= len(consumers) {
 		return nil, fmt.Errorf("invalid partition index %d expected a partition between [0-%d]",
-			partition, len(consumers))
+			partition, len(consumers)-1)
 	}
 	return consumers[partition], nil
 }
@@ -794,8 +794,9 @@ func (c *consumer) partitionConsumers() []*partitionConsumer {
 	if v == nil {
 		return nil
 	}
-	consumers := v.([]*partitionConsumer)
-	return append([]*partitionConsumer(nil), consumers...)
+	// The slice stored in c.consumers is published via copy-on-write.
+	// Callers must treat the returned slice as immutable.
+	return v.([]*partitionConsumer)
 }
 
 func (c *consumer) hasNext() bool {
--- a/pulsar/consumer_impl.go
+++ b/pulsar/consumer_impl.go
@@ -399,16 +399,16 @@ func (c *consumer) internalTopicSubscribeToPartitions() error {
 	for partitionIdx := startPartition; partitionIdx < newNumPartitions; partitionIdx++ {
 		partitionTopic := partitions[partitionIdx]
 
-		go func() {
+		go func(partitionIdx int, partitionTopic string) {
 			defer wg.Done()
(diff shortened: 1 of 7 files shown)


RobertIndie (commented):
Please take a look at the copilot's comments. Other part looks good to me.

BewareMyPower:
@RobertIndie addressed, PTAL again

nodece:
Great fix on the deadlock path from #1494 — the lock-free `partitionConsumers()` snapshot + delayed `startDispatcher()` is a solid direction for the original blocking issue.

I think there is still a new concurrency window introduced by removing the old mutex guarding `consumers`.

In the new code:

- `internalTopicSubscribeToPartitions()` builds `newConsumers`, then publishes via `c.consumers.Store(...)`, then starts dispatchers for new partitions.
- `closeWithCause()` closes only one snapshot (`consumers := c.partitionConsumers()`) and no longer serializes with partition expansion.

So this interleaving seems possible:

1. Goroutine A enters `internalTopicSubscribeToPartitions()` and starts creating new partition consumers.
2. Goroutine B calls `closeWithCause()`, takes an old snapshot, closes those consumers, and continues close flow.
3. Goroutine A continues and still executes `c.consumers.Store(...)` + `startDispatcher()` for newly created consumers.

Result: newly added partition consumers may be published/started after close has already closed only the old snapshot.

Before this PR, the parent lock serialized expansion and close-related operations over `c.consumers`; removing that lock fixes deadlock, but also removes that lifecycle serialization.

Could we gate expansion with context cancellation or a closing flag (checked before create and before Store/startDispatcher), so in-flight expansion stops/cleans up once close starts?

BewareMyPower:
@nodece `closeWithCause` calls `c.stopDiscovery()`, which cancels the timer and waits for the previous `internalTopicSubscribeToPartitions` is done by `wg.Wait()`.

https://github.com/apache/pulsar-client-go/blob/a1765ef9c22a4cbd7b7a4704b6136e96860b3815/pulsar/consumer_impl.go#L346-L349

nodece (approved):

PR #1502, opened 2026-05-20
fix: avoid infinite retry in blue-green migration test

nodece:
### Motivation

The extensible-load-manager CI job timed out in TestBlueGreenMigrationTestSuite/TestTopicMigration/proxyConnection after 5 minutes.

From the stack and logs, the test was stuck waiting on WaitGroup while producer/consumer goroutines were still looping in retry paths. During migration, producer can enter terminal states (for example TopicTerminated or ProducerClosed), but the test retry loops had no terminal-exit logic, causing effectively unbounded retries and suite timeout.

### Modifications

- Add terminal error handling in producer send retry loop:
  - if error is ErrTopicTerminated or ErrProducerClosed, fail fast instead of retrying forever.
- Add bounded retry windows for both producer and consumer loops (30 seconds per message stage).
- Add an error channel and stage-level wait helper around WaitGroup waits to fail early on goroutine errors.
- Add stage timeout protection while waiting for:
  - pre-unload send/receive synchronization
  - producer/consumer goroutine completion
- Keep per-iteration context cancellation immediate (cancel after each send/receive attempt).

These changes make the test deterministic under migration failures and prevent hanging until global test timeout.

### DIFF
--- a/pulsar/blue_green_migration_test.go
+++ b/pulsar/blue_green_migration_test.go
@@ -21,6 +21,7 @@ package pulsar
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"runtime"
@@ -148,6 +149,7 @@ func testTopicMigrate(
 	// Signals both producer and consumer have processed `messageCountBeforeUnload` messages
 	wgSendAndReceiveMessages := sync.WaitGroup{}
 	wgSendAndReceiveMessages.Add(2)
+	errCh := make(chan error, 1)
 
 	// Producer
 	go func() {
@@ -159,6 +161,7 @@ func testTopicMigrate(
 			}
 
 			pm := ProducerMessage{Payload: []byte(fmt.Sprintf("hello-%d", i))}
+			retryStarted := time.Now()
 
 			for true {
 				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
@@ -166,6 +169,20 @@ func testTopicMigrate(
 				cancel()
 				if err == nil {
 					break
+				}
+				if errors.Is(err, ErrTopicTerminated) || errors.Is(err, ErrProducerClosed) {
+					select {
+					case errCh <- fmt.Errorf("producer became terminal during migration at message %d: %w", i, err):
+					default:
+					}
+					return
+				}
+				if time.Since(retryStarted) > 30*time.Second {
+					select {
+					case errCh <- fmt.Errorf("producer send retry exceeded 30s at message %d: %w", i, err):
+					default:
+					}
+					return
 				}
 				time.Sleep(1000 * time.Millisecond)
 			}
@@ -185,6 +202,7 @@ func testTopicMigrate(
 				wgUnload.Wait()
 			}
 
+			retryStarted := time.Now()
 			for true {
 				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 				m, err := consumer.Receive(ctx)
@@ -194,6 +212,13 @@ func testTopicMigrate(
 					if err == nil {
 						break
 					}
+				}
+				if time.Since(retryStarted) > 30*time.Second {
+					select {
+					case errCh <- fmt.Errorf("consumer receive/ack retry exceeded 30s at message %d: %w", i, err):
+					default:
+					}
+					return
 				}
 				time.Sleep(100 * time.Millisecond)
 			}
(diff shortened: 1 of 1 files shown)


pulsar/blue_green_migration_test.go
```diff
@@ -159,14 +161,29 @@ func testTopicMigrate(
...
 				_, err := producer.Send(ctx, &pm)
+				cancel()
 				if err == nil {
 					break
 				}
+				if errors.Is(err, ErrTopicTerminated) || errors.Is(err, ErrProducerClosed) {
+					select {
+					case errCh <- fmt.Errorf("producer became terminal during migration at message %d: %w", i, err):
+					default:
+					}
```
BewareMyPower, line 180: Why did you use a select block rather than a simple `errCh <- fmt.Errorf(...)`?

pulsar/blue_green_migration_test.go
```diff
@@ -159,14 +161,29 @@ func testTopicMigrate(
...
+					return
+				}
+				if time.Since(retryStarted) > 30*time.Second {
+					select {
+					case errCh <- fmt.Errorf("producer send retry exceeded 30s at message %d: %w", i, err):
+					default:
+					}
```
BewareMyPower, line 187: The same question with my previous comment

pulsar/blue_green_migration_test.go
```diff
@@ -185,16 +202,24 @@ func testTopicMigrate(
 				wgUnload.Wait()
 			}
 
+			retryStarted := time.Now()
 			for true {
 				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
-				defer cancel()
 				m, err := consumer.Receive(ctx)
+				cancel()
```
BewareMyPower, line 212: The original `defer cancel()` call hits a typical error in golang: the deferred function won't be called until returned from a function, so if the loop has been iterated for many times, there could be many context objects not cancelled.

A better approach should be declaring an anonymous function and keep `defer cancel()` not changed:

```go
err := func() error {
    ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
    defer cancel()
    /* ... */
}
```

pulsar/blue_green_migration_test.go
```diff
@@ -185,16 +202,24 @@ func testTopicMigrate(
...
 					err = consumer.Ack(m)
 					if err == nil {
 						break
 					}
 				}
+				if time.Since(retryStarted) > 30*time.Second {
+					select {
+					case errCh <- fmt.Errorf("consumer receive/ack retry exceeded 30s at message %d: %w", i, err):
+					default:
+					}
```
BewareMyPower, line 223: The same question

pulsar/blue_green_migration_test.go
nodece: I intentionally used a non-blocking send here to avoid potential deadlocks. If `errCh` is no longer being consumed (or the buffer is full), a blocking send could prevent the migration goroutine from exiting cleanly.

pulsar/blue_green_migration_test.go
nodece: @BewareMyPower Done.

pulsar/blue_green_migration_test.go
```diff
@@ -148,6 +149,7 @@ func testTopicMigrate(
 	// Signals both producer and consumer have processed `messageCountBeforeUnload` messages
 	wgSendAndReceiveMessages := sync.WaitGroup{}
 	wgSendAndReceiveMessages.Add(2)
+	errCh := make(chan error, 1)
```
BewareMyPower, line 152: Continue the previous [discussion](https://github.com/apache/pulsar-client-go/pull/1502#discussion_r3332506349) here.

Since this channel could only be sent once because every `select { case errCh<- xxx, default }` block is followed by a return.

```go
		select {
		case <-doneCh:
			return true
		case err := <-errCh:
			req.NoError(err, stage)
			return false
```

However, it's possible that both `doneCh` and `errCh` have a message, in this case, any of them can be chosen randomly.

Therefore, I suggest just storing errors into a slice and check the errors in the `case <-doneCh` case.

BewareMyPower (dismissed):

pulsar/blue_green_migration_test.go
```diff
@@ -159,14 +161,32 @@ func testTopicMigrate(
...
 			for true {
-				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
-				defer cancel()
-				_, err := producer.Send(ctx, &pm)
+				err := func() error {
+					ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+					defer cancel()
+					_, err := producer.Send(ctx, &pm)
+					return err
+				}
```
BewareMyPower: You missed a call to this function, with current code base, `err` is a function but not an `error`.

pulsar/blue_green_migration_test.go
BewareMyPower: There are two goroutines, the 1st could send the error to `errCh`, the `doneCh` receives a message after two goroutines are done. i.e. if the 1st goroutine completes before the 2nd one, there could be a race.

But if `doneCh` receives a message, the error is guaranteed to be sent to `errCh` if there is an error. Therefore, changing the error into an `atomic.Value` should make sense

nodece: Race issue has been fixed.

crossoverJie (approved):

PR #1524, opened 2026-07-15
[Issue #1523][crypto] Fix encryption key metadata not refreshing on rotation

### LINKED ISSUE #1523: Producer encryption key metadata is never refreshed after first message
## Expected behavior

When `KeyReader.PublicKey()` returns updated metadata for a key name (e.g., after key rotation), subsequent `Encrypt` calls should use the new metadata.

## Actual behavior

`DefaultMessageCrypto.Encrypt` caches the `EncryptionKeyInfo` (including metadata) on the first call for each `keyName` in `encryptedDataKeyMap`. The guard at line 153 (`if _, ok := d.encryptedDataKeyMap.Load(keyName); !ok`) prevents `addPublicKeyCipher` from being called again, so metadata is frozen for the producer's lifetime.

Long-lived producers stamp stale key-id metadata onto every message. After key rotation, consumers request private keys that no longer exist.

## Steps to reproduce

1. Create a `KeyReader` returning `Metadata{"key-id": "uuid-A"}` for key name `"latest"`
2. Call `Encrypt(["latest"], ...)` — metadata has `uuid-A` ✓
3. Update `KeyReader` to return `Metadata{"key-id": "uuid-B"}`
4. Call `Encrypt(["latest"], ...)` again
5. **Expected:** `uuid-B` / **Actual:** still `uuid-A`

<details>
<summary>Failing test</summary>

```go
func TestKeyRotationMetadataNotRefreshed(t *testing.T) {
	keyName := "latest"
	keyIDs := []string{"uuid-A", "uuid-B"}
	keyReader := newRotatingKeyReader(t, keyIDs) // returns keyIDs[callCount % len] on each PublicKey call

	msgCrypto, _ := NewDefaultMessageCrypto("test", true, log.DefaultNopLogger())

	// First encrypt: captures uuid-A
	msg1 := &pb.MessageMetadata{}
	msgCrypto.Encrypt([]string{keyName}, keyReader, NewMessageMetadataSupplier(msg1), []byte("m1"))
	keyID1 := msg1.EncryptionKeys[0].Metadata // uuid-A ✓

	// KeyReader now returns uuid-B, but Encrypt won't call it again
	msg2 := &pb.MessageMetadata{}
	msgCrypto.Encrypt([]string{keyName}, keyReader, NewMessageMetadataSupplier(msg2), []byte("m2"))
	keyID2 := msg2.EncryptionKeys[0].Metadata

	// FAILS: keyID2 == "uuid-A" (stale), expected "uuid-B"
	assert.Equal(t, "uuid-B", keyID2)
}
```

</details>

## System configuration

- pulsar-client-go: v0.19.0, v0.20.0, master
- Go/OS: any

---

I plan to open a PR with a fix. Thanks!

jfretheim:
Fixes #1523

### Motivation

`DefaultMessageCrypto` caches `EncryptionKeyInfo` after the first `Encrypt` and never refreshes it. When using symbolic key names that resolve to different keys over time (rotation), consumers receive stale metadata pointing at keys that may no longer exist.

### Modifications

- Compare current metadata with cached on each `Encrypt`; re-encrypt data key if they differ
- Add `metadataEqual` helper and `addPublicKeyCipherWithKeyInfo` to avoid double `PublicKey()` calls

### Verifying this change

- `TestKeyRotationMetadataNotRefreshed` — proves the bug (fails without fix)
- `TestKeyRotationWorksWithRemoveKeyCipher` — proves the `RemoveKeyCipher` workaround
- `TestKeyRotationMultipleKeys` — proves independent rotation detection per key

- The public API: no
- The schema: no
- The wire protocol: no

### Documentation

### DIFF
--- a/pulsar/crypto/default_message_crypto.go
+++ b/pulsar/crypto/default_message_crypto.go
@@ -92,16 +92,24 @@ func (d *DefaultMessageCrypto) AddPublicKeyCipher(keyNames []string, keyReader K
 }
 
 func (d *DefaultMessageCrypto) addPublicKeyCipher(keyName string, keyReader KeyReader) error {
-	d.cipherLock.Lock()
-	defer d.cipherLock.Unlock()
 	if keyName == "" || keyReader == nil {
 		return fmt.Errorf("keyname or keyreader is nil")
 	}
 
-	// read the public key and its info using keyReader
 	keyInfo, err := keyReader.PublicKey(keyName, nil)
 	if err != nil {
 		return err
+	}
+
+	return d.addPublicKeyCipherWithKeyInfo(keyName, keyInfo)
+}
+
+func (d *DefaultMessageCrypto) addPublicKeyCipherWithKeyInfo(keyName string, keyInfo *EncryptionKeyInfo) error {
+	d.cipherLock.Lock()
+	defer d.cipherLock.Unlock()
+
+	if keyName == "" || keyInfo == nil {
+		return fmt.Errorf("keyname or keyinfo is nil")
 	}
 
 	parsedKey, err := d.loadPublicKey(keyInfo.Key())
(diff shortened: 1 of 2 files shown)


jfretheim:
Converting this to a draft while I rework the approach.

On a fresh review, calling `KeyReader.PublicKey()` on every `Encrypt` regresses the bundled `FileKeyReader`, which does an `os.ReadFile` per call — that turns a once-per-keyname cost into a per-message disk read on the hot path. The Java client avoids this by refreshing keys on a periodic timer (`keyGeneratorTask`) rather than per-encrypt, and I'd like this PR to follow that model instead.

I'll push an updated design and undraft. Thanks for your patience.

jfretheim:
Closing this in favor of a cleaner approach.

On deeper review, the per-`Encrypt` `PublicKey()` call regresses the bundled `FileKeyReader` (an `os.ReadFile` per message on the hot path), and the current cache-until-removed behavior actually matches the documented `MessageCrypto` contract — `RemoveKeyCipher` is the sanctioned way for an application to handle rotation. So #1523 isn't strictly a bug against the existing contract.

I'll open a new PR that instead implements the periodic-refresh idea noted in the crypto layer's own TODO, as an **opt-in** background refresh (default off, zero behavior change, no `FileKeyReader` regression). Will link it here.

Issue #1528, opened 2026-08-20
pulsaradmin: ProducerConfig cannot express PIP-401 batchingConfig


david-streamlio:
**Is your feature request related to a problem? Please describe.**

`pulsaradmin`'s `utils.ProducerConfig` cannot express PIP-401's producer batching configuration, so no Go-based tool can set it on a Pulsar Function, Source, or Sink.

PIP-401 (apache/pulsar#23860, merged 2025-03-12) added a `BatchingConfig` to the Java admin API's `ProducerConfig` and wired it through to the `ProducerBuilder`. The Java class today is:

```java
// pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/functions/ProducerConfig.java
public class ProducerConfig {
    private Integer maxPendingMessages;
    private Integer maxPendingMessagesAcrossPartitions;
    private Boolean useThreadLocalProducers;
    private CryptoConfig cryptoConfig;
    private String batchBuilder;
    private CompressionType compressionType;
    private BatchingConfig batchingConfig;   // <-- added by PIP-401
}
```

`pulsaradmin/pkg/utils/producer_config.go` on `master` has every field except the last one:

```go
type ProducerConfig struct {
	MaxPendingMessages int `json:"maxPendingMessages" yaml:"maxPendingMessages"`
	//nolint
	MaxPendingMessagesAcrossPartitions int `json:"maxPendingMessagesAcrossPartitions" yaml:"maxPendingMessagesAcrossPartitions"`

	UseThreadLocalProducers bool          `json:"useThreadLocalProducers" yaml:"useThreadLocalProducers"`
	CryptoConfig            *CryptoConfig `json:"cryptoConfig" yaml:"cryptoConfig"`
	BatchBuilder            string        `json:"batchBuilder" yaml:"batchBuilder"`
	CompressionType         string        `json:"compressionType" yaml:"compressionType"`
}
```

Since `ProducerConfig` is a typed struct serialized to JSON with no passthrough field, there is no workaround: a Go caller simply cannot put `batchingConfig` on the wire, even though the broker accepts it and the Java `pulsar-admin` CLI sends it happily via `functions create --producer-config '<json>'`.

The concrete impact is on infrastructure-as-code. The Terraform provider (`streamnative/terraform-provider-pulsar`) imports this package, so producer batching is unreachable from Terraform for every function runtime — including Java, where the broker and runtime support have been complete since PIP-401. `pulsarctl` has the same ceiling.

[...]

Issue #219, opened 2020-04-13
Selective replication


yarthur1:
I want  restrict replication selectively by specifying a replication list for a message and used ProducerMessage ReplicationClusters field.I have three pulsar clusters in a instance,A send messages to B and C,I want that A doesn't receive A send messages.The ReplicationClusters field value is pulsar cluster name only B and C,but it seems not work.A pulsar cluster still receive messages.Am I something wrong?

PR #1529, opened 2026-08-20
fix(pulsaradmin): preserve explicit zero receiver queue size

freeznet:

Fixes #<xyz>

Master Issue: #<xyz>

### Motivation

### Modifications

### Verifying this change

*(or)*

*(or)*

*(example:)*

  - The public API: (yes / no)

### Documentation

### DIFF
--- a/pulsaradmin/pkg/utils/consumer_config.go
+++ b/pulsaradmin/pkg/utils/consumer_config.go
@@ -17,13 +17,66 @@
 
 package utils
 
+import "encoding/json"
+
 type ConsumerConfig struct {
-	SchemaType         string            `json:"schemaType,omitempty" yaml:"schemaType"`
-	SerdeClassName     string            `json:"serdeClassName,omitempty" yaml:"serdeClassName"`
-	RegexPattern       bool              `json:"regexPattern,omitempty" yaml:"regexPattern"`
-	ReceiverQueueSize  int               `json:"receiverQueueSize,omitempty" yaml:"receiverQueueSize"`
-	SchemaProperties   map[string]string `json:"schemaProperties,omitempty" yaml:"schemaProperties"`
-	ConsumerProperties map[string]string `json:"consumerProperties,omitempty" yaml:"consumerProperties"`
-	CryptoConfig       *CryptoConfig     `json:"cryptoConfig,omitempty" yaml:"cryptoConfig"`
-	PoolMessages       bool              `json:"poolMessages,omitempty" yaml:"poolMessages"`
+	SchemaType           string `json:"schemaType,omitempty" yaml:"schemaType"`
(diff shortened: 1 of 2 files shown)


david-streamlio:
Thanks for putting this together — I verified the change end to end against Pulsar 4.0.3 while reviewing the downstream Terraform PR (streamnative/terraform-provider-pulsar#219), and it does exactly what is needed: an explicitly configured `receiverQueueSize: 0` now round-trips with `HasReceiverQueueSize() == true` and value `0`, where `omitempty` previously dropped it and the broker silently applied the 1000 default.

I had filed #1528 for this before seeing your PR. Since the template body here is still unfilled, it may be worth adding `Fixes #1528` and some motivation so a reviewer has the context without reconstructing it — offered as a starting point, edit or ignore as you see fit:

> ### Motivation
>
> `ConsumerConfig.ReceiverQueueSize` is an `int` tagged `json:"receiverQueueSize,omitempty"`, so an explicitly configured zero is indistinguishable from an unset field and is omitted from the request. Pulsar treats a missing value as "use the default" and applies 1000.
>
> Zero is a meaningful value: Pulsar validates `receiverQueueSize >= 0` (`FunctionConfigUtils.doCommonChecks`), and zero disables consumer prefetch, which is what latency-sensitive callers want. The Java admin API models the field as a nullable `Integer` precisely so unset and zero stay distinct; the Go model cannot currently express that.
>
> This blocks infrastructure-as-code callers in particular — the Terraform provider (streamnative/terraform-provider-pulsar#219) has to pin a fork to expose the setting at all.
>
> ### Modifications
>
> Tag `ReceiverQueueSize` as `json:"-"` and track explicitness in an unexported `receiverQueueSizeSet` flag, with `MarshalJSON`/`UnmarshalJSON` mapping it to a `*int` on the wire, plus `SetReceiverQueueSize()` and `HasReceiverQueueSize()`.
>
> The exported field stays an `int`, so existing callers that assign a non-zero value compile and behave exactly as before; `HasReceiverQueueSize()` falls back to `ReceiverQueueSize != 0` for values assigned directly rather than through the setter.

One question on the round trip, which is the only thing I could not settle by reading: `UnmarshalJSON` resets `receiverQueueSizeSet` to false and re-derives it from the payload, so a server response omitting `receiverQueueSize` yields `HasReceiverQueueSize() == false`. That is right for a fresh decode. Is a decode ever performed onto a struct a caller has already populated — where clearing a previously-set flag would lose the distinction? I could not find such a path in `pulsaradmin`, so I suspect the answer is no and this is purely defensive; flagging it only because it is the kind of thing that is easier to confirm now than to debug later.

Happy to help with the downstream bump in the provider once this lands and there is a tagged `apache/pulsar-client-go` version to depend on, so the `replace` directive in #219 can come out.

zymap (approved):

PR #1539, opened 2026-09-22
[fix][compression] Add backward compatibility for lz4 v2 uncompressed block format in Decompress

zjxxzjwang:
### Motivation

After upgrading `pierrec/lz4` from v2 to v4 (introduced in #1341), consumers may encounter `failed to decompress` errors when consuming messages that were produced by older client versions (e.g., 0.12.1) using lz4 v2.

The root cause is in the `Compress` method's `size == 0` branch. When `lz4.CompressBlock` returns 0 (meaning the data is incompressible — e.g., too small or already compressed/encrypted), the code writes a custom format: a `writeSize`-encoded header followed by the raw uncompressed data. This custom format is **not** a standard lz4 block, so `lz4.UncompressBlock` (both v2 and v4) cannot decode it. When a consumer running lz4 v4 encounters such a message produced by an older client, decompression fails.

### Modifications

- **`pulsar/internal/compression/lz4.go`**:
  - Modified `Decompress` to add a fallback path: when `lz4.UncompressBlock` fails, attempt to parse the custom `writeSize header + raw data` format and extract the original uncompressed payload.
  - Added `tryReadUncompressedBlock` helper that validates the `writeSize` header against the expected `originalSize` and total `src` length (double validation for safety).
  - Added `readSize` function as the inverse of the existing `writeSize`, parsing the encoded data size from the header bytes.

- **`pulsar/internal/compression/compression_test.go`**:
  - Added `TestLz4DecompressUncompressedBlock` to verify backward compatibility with the custom uncompressed block format (both small data `< 0xF` with 1-byte header and large data `>= 0xF` with multi-byte header).
  - Added `TestLz4ReadWriteSizeRoundTrip` to verify that `readSize` is the correct inverse of `writeSize` across various boundary values (0, 1, 14, 15, 16, 255, 256, 1000, 65535).

### Verifying this change

- Added `TestLz4DecompressUncompressedBlock` with two sub-cases: `small_uncompressed_block` (data length < 0xF, 1-byte header) and `large_uncompressed_block` (data length >= 0xF, multi-byte header) to verify the fallback decompression path.
- Added `TestLz4ReadWriteSizeRoundTrip` to verify the `readSize`/`writeSize` round-trip correctness across 12 boundary values.
- All existing tests (`TestCompression`, `TestCompressionNoBuffers`, `TestJavaCompatibility`, `TestDecompressionError`) continue to pass, confirming no regression.

- The public API: no
- The schema: no
- The wire protocol: no

### Documentation

- If a feature is not applicable for documentation, explain why? This is a bug fix for backward compatibility, no new feature is introduced.


wolfstudy (approved):
Great changes,  LGTM +1

crossoverJie (approved):
Good catch.

